PR #119, opened 2025-08-27
W-19443936: Simplify some factory architecture

kylewalke:
### What does this PR do?
 - Unified platform-specific factories into single implementations
  - Replaced separate browser/node/worker storage factories with unified StorageFactory
  - Consolidated environment detection into single EnvironmentUtils
  - Added UnifiedConnectionFactory to replace platform-specific variants
  - Removed redundant test files and simplified test architecture
  
@W-19443936@

### Functionality Before
Duplicated code and some redundancies

### Functionality After
Works like a charm with unified approach!


packages/apex-ls/src/communication/MessageTransports.ts
```diff
@@ -67,13 +67,19 @@ export class SelfMessageTransport implements MessageTransport {
...
-      this.selfContext = self as any;
     } else {
-      throw new Error(
-        'SelfMessageTransport can only be used in a worker context',
-      );
+      // Use type-safe access to worker self context
+      const {
+        getWorkerSelf,
+        isWorkerPostMessageAvailable,
+      } = require('../utils/EnvironmentUtils');
```
peternhale, line 75: I thought this was allowed only if module is pure esm?

packages/apex-ls/src/communication/MessageTransports.ts
```diff
@@ -96,7 +102,7 @@ export class SelfMessageTransport implements MessageTransport {
   }
 
   onError(handler: (error: Error) => void): { dispose(): void } {
-    const errorHandler = (event: ErrorEvent) => {
+    const errorHandler = (event: any) => {
```
peternhale, line 105: Why the type loosening?

packages/apex-ls/src/communication/PlatformBridges.browser.ts
```diff
@@ -46,7 +47,7 @@ export class BrowserMessageBridge extends BaseMessageBridge {
   /**
    * Creates a browser-to-worker message bridge for client-side communication
    */
-  static forWorkerClient(worker: Worker, logger?: Logger): MessageConnection {
+  static forWorkerClient(worker: any, logger?: Logger): MessageConnection {
```
peternhale, line 50: We are becoming less type safe?

packages/apex-ls/src/storage/StorageImplementations.ts
```diff
@@ -83,12 +83,24 @@ class MemoryStorage extends BaseStorage {
...
   private storeName = 'documents';
-  private db?: IDBDatabase;
+  private db?: any; // IDBDatabase - type only available in browser environment
 
   protected async performInitialization(): Promise<void> {
+    if (this.config?.storagePrefix) {
+      this.dbName = this.config.storagePrefix;
+    }
+
+    const { getIndexedDB, isIndexedDBAvailable } = await import('../utils/EnvironmentUtils');
```
peternhale, line 96: why not use require(), like in other places?

packages/apex-ls/src/communication/MessageTransports.ts
kylewalke: I've tested adding these back in after importing webworker and the ErrorEvent works as well!

PR #389, opened 2026-05-19
build(deps): combine all dependabot dependency updates - W-22567873

kylewalke:
## Summary
- Combines 11 open Dependabot PRs into a single update
- All tests pass (4232 passed, 0 failed)
- No breaking changes detected

## Updates
| Package | From | To |
|---------|------|-----|
| @types/node | 20.19.39 | 20.19.40 |
| knip | 6.11.0 | 6.12.1 |
| fast-uri | 3.1.0 | 3.1.2 |
| fast-xml-builder | 1.1.5 | 1.2.0 |
| ws | 8.19.0 | 8.20.1 |
| protobufjs | 7.5.5 | 7.5.8 |
| @protobufjs/utf8 | 1.1.0 | 1.1.1 |
| jest | 30.3.0 | 30.4.1 |
| @jest/globals | 30.3.0 | 30.4.1 |
| jest-environment-jsdom | 30.3.0 | 30.4.1 |
| @salesforce/vscode-services | 66.9.0 | 66.10.1 |

@W-22567873@

## Test plan
- [x] All unit tests pass locally (4232 passed)
- [x] Lint passes
- [ ] CI passes on PR


package.json
```diff
@@ -123,6 +123,9 @@
     "vscode-languageserver-textdocument": "^1.0.12",
     "vscode-uri": "^3.1.0"
   },
+  "overrides": {
+    "vscode-jsonrpc": "8.2.1"
```
kylewalke, line 127: We force a single vscode-jsonrpc copy across the tree here. Our workspaces declare ^8.2.1 but vscode-languageserver-protocol@3.17.5 hard-pins exact "8.2.0" with no caret. Without this override, npm installs both 8.2.1 (top-level) and 8.2.0 (nested under v-l-protocol), and TypeScript treats the two physical copies as distinct type identities, causing TS2306 "not a module" errors when types cross the boundary. 8.2.0 -> 8.2.1 is a patch release, API-compatible at runtime, so deduping to 8.2.1 is safe.

peternhale (approved):

PR #412, opened 2026-06-01
fix(apex-lsp-compliant-services): use globalThis.setTimeout for Node 24 compatibility - W-22769237

peternhale:
### What does this PR do?
Uses `globalThis.setTimeout` instead of bare `setTimeout` in tests that call `jest.useRealTimers()`. Node 24 does not expose `setTimeout` as a global in Jest's node test environment after switching to real timers.

@W-22769237@

### Before/After
**Before:** `DocumentChangeBatcher.test.ts` fails with `ReferenceError: setTimeout is not defined` on Node 24 (`current`), causing CI Complete to fail on all PRs.

**After:** Tests pass on Node 20, 22, and 24.

### DIFF
--- a/packages/lsp-compliant-services/test/services/DocumentChangeBatcher.test.ts
+++ b/packages/lsp-compliant-services/test/services/DocumentChangeBatcher.test.ts
@@ -166,9 +166,6 @@ describe('DocumentChangeBatcher', () => {
 
   describe('concurrency limit', () => {
     it('should limit concurrent processor calls to maxConcurrentParses', async () => {
-      // Use real timers for this test since we need promise resolution
-      jest.useRealTimers();
-
       let concurrentCount = 0;
       let maxObservedConcurrent = 0;
       const resolvers: Array<() => void> = [];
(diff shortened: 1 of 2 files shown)


packages/lsp-compliant-services/test/services/DocumentChangeBatcher.test.ts
```diff
@@ -197,31 +194,31 @@ describe('DocumentChangeBatcher', () => {
...
+      await Promise.resolve();
 
       // Only maxConcurrentParses (2) should be running
       expect(concurrentCount).toBe(2);
       expect(blockingProcessor).toHaveBeenCalledTimes(2);
 
       // Resolve one — should trigger the next queued flush
       resolvers[0]();
-      await new Promise((resolve) => setTimeout(resolve, 10));
+      await Promise.resolve();
```
kylewalke, line 207: So this double resolve flushes the promises twice?

kylewalke (approved):

PR #416, opened 2026-06-03
feat(apex-lsp-compliant-services): inject IWorkspaceLoadCoordinator into ReferencesProcessingService - W-22692414

kylewalke:
## Summary

Story 6.02 of the Find References offload epic. Replaces the direct
`LSPConfigurationManager.getInstance().getConnection()` lookup in
`ReferencesProcessingService` with an injected `IWorkspaceLoadCoordinator`
so the same service runs in both the coordinator process and enrichment
workers without branching on the runtime topology.

- **`LocalWorkspaceLoadCoordinator`** wraps `ensureWorkspaceLoaded()` and is
  the default when `ServiceFactory` has a `Connection` (coordinator process).
- **`RemoteWorkspaceLoadCoordinator`** forwards to the coordinator over the
  assistance bus via `coordinator:EnsureWorkspaceLoaded`. Worker bootstraps
  will inject this in story 6.05; story 6.03 lands the coordinator-side
  handler that services it.
- `ServiceFactory` builds the Local default when a connection is present;
  workers will inject Remote via the new optional
  `ServiceDependencies.workspaceLoadCoordinator` field.

POC ref: design doc §3.5.

@W-22692414@

## Test plan
- [x] Existing `ReferencesProcessingService` tests updated to drive the injected coordinator
- [x] New `WorkspaceLoadCoordinator.test.ts` covering both Local and Remote impls
- [x] Full `lsp-compliant-services` test suite green (560/570 passed, 10 pre-existing skips)
- [x] Repo-wide pre-commit gate green (4242 passed, 0 failed)


peternhale (dismissed):
Overall this is a clean, well-structured dependency-inversion. The `IWorkspaceLoadCoordinator` interface is the right abstraction and the test migration from mocked globals to an injected mock is exactly what should happen here. Approving with nits below.

---

**Nit — bare `catch {}` in `HandlerFactory.createReferencesHandler`**

The try/catch around `LSPConfigurationManager.getInstance().getConnection()` swallows *all* exceptions, including programming errors. If the manager throws for a reason other than "not yet registered", the coordinator is silently dropped and references degrade without any log trace. Prefer catching only the expected sentinel (e.g. `NotYetRegisteredError`), or at minimum log at `warn` level in the catch body so the silent degradation is observable.

---

**Nit — `isLoadingRef` stays set permanently on the worker**

`RemoteWorkspaceLoadCoordinator.ensureLoaded` sets `isLoadingRef = true` but never clears it (comment acknowledges this; 6.12 adds the completion broadcast). This is fine architecturally, but the tests call `reset()` between each `describe` block without explaining why — a short comment linking to the permanent-loading-state design would prevent future "why do we need reset() here?" confusion.

---

**Nit — `catch: (err) => err` in `Effect.tryPromise`**

```ts
Effect.tryPromise({
  try: () => proxy(...),
  catch: (err) => err,   // propagates err as-is
})
```

This re-uses the thrown value as the "error channel value", which works because `catchAll` absorbs it, but it's an unusual pattern. `catch: (err) => new Error(String(err))` (or a typed error class) makes the intent clearer and avoids leaking an `unknown` through the error channel.

---

**Nit — `null` / `undefined` dual nullability in `ServiceFactory`**

`getWorkspaceLoadCoordinator()` returns `IWorkspaceLoadCoordinator | null` internally, but is passed as `?? undefined` to the `ReferencesProcessingService` constructor which takes `IWorkspaceLoadCoordinator | undefined`. This is correct TypeScript but the two representations add a small mental tax. Minor — could normalize to `undefined` throughout or `null` throughout, whichever matches the broader project convention.

PR #417, opened 2026-06-03
feat(apex-lsp-extension): handle coordinator:EnsureWorkspaceLoaded in primary mediator - W-22692415

kylewalke:
## Summary

Story 6.03 of the Find References offload epic. Adds the coordinator-side
branch in LCSAdapter's `CoordinatorAssistanceMediator` primary handler that
services worker requests for workspace load.

- Workers (enrichment / data-owner) have no LSP `Connection` of their own.
  They forward through the assistance bus, and the coordinator fires the
  `apex/requestWorkspaceLoad` notification on its own connection.
- Routed **before** the catch-all `connection.sendRequest`, since
  `coordinator:EnsureWorkspaceLoaded` is not a real LSP method and the
  fall-through would surface as a "Method not found" error back to the worker.
- Re-uses the existing `ensureWorkspaceLoaded` Effect from
  `apex-lsp-compliant-services` so the local state guard
  (`isLoadedRef` / `isLoadingRef`) is shared between coordinator-process
  callers and the worker-driven path.

POC ref: design doc §3.3.

@W-22692415@

## Test plan
- [x] New `LCSAdapter-ensureWorkspaceLoaded.test.ts` exercises the handler through the real `CoordinatorAssistanceMediator` wire path
- [x] Happy path: `workDoneToken` propagated, response posted with `correlationId`
- [x] State guard: second request while loading does not double-fire `apex/requestWorkspaceLoad`
- [x] Full `apex-ls` test suite green (132/136 passed, 4 pre-existing skips)
- [x] Repo-wide pre-commit gate green (4240 passed, 0 failed)


peternhale (approved):
Extracting the primary handler into its own file was the right call — the inline lambda was untestable and the production branching logic is now exercised directly. The test using the real `CoordinatorAssistanceMediator` wire path is exactly the right fidelity for a stateful router. Approving with nits.

---

**Nit — `Date.now()` in `makeRequest` correlationId**

```ts
correlationId: `test-${Date.now()}`,
```

Non-deterministic IDs are harmless here but are a latent flakiness risk if these tests are ever snapshotted or compared across runs. A simple counter (`let seq = 0; ... correlationId: \`test-${++seq}\``) or a fixed string per test site is more robust.

---

**Nit — `setTimeout`-based assertions**

```ts
await new Promise((r) => setTimeout(r, 50));
expect(sendNotification).toHaveBeenCalledTimes(1);
```

Timing-based flush is a flakiness risk under CI load. If `CoordinatorAssistanceMediator` processes messages asynchronously, check whether it exposes a drain/flush mechanism. If the handler resolution is microtask-based a `await Promise.resolve()` loop may be sufficient. If `50ms` is genuinely required, a comment explaining why would help future readers avoid reducing it to `10ms` and introducing intermittent failures.

---

**Nit — unchecked cast `connection as Connection`**

`PrimaryAssistanceHandlerDeps.connection` is typed as `Pick<Connection, 'sendRequest' | 'sendNotification'>`, which is correct for the handler's own branches. But `ensureWorkspaceLoaded` receives it cast to the full `Connection`. If `ensureWorkspaceLoaded` only ever calls `sendNotification`, widening the `Pick` in the deps type to include whatever `ensureWorkspaceLoaded` actually calls, and removing the cast, would make the contract explicit and catch future drift at compile time.

PR #418, opened 2026-06-03
feat(apex-lsp-parser-ast,apex-lsp-extension): data-owner ResolveDependentUris handler - W-22692417

kylewalke:
## Summary

Story 6.04 of the Find References offload epic. Adds the data-owner-side
handler that, given a target URI, returns serialized symbol tables for
every file whose declared symbols reference any symbol declared in the
target. Used by Find References on the enrichment worker to load
caller-side symbol tables before the algorithm runs.

- New helper `resolveDependentUris(symbolManager, uri, symbolName?)` in
  `apex-lsp-parser-ast`: iterates declared symbols in `uri`, calls
  `findReferencesTo` on each, dedups dependent file URIs, serializes
  matching symbol tables. Self-references (target file referencing its
  own symbols) are skipped — the caller already owns the target's table.
  Honors optional `symbolName` narrowing accepted on the wire by 6.01.
  Skips dependents whose symbol table is missing on this data-owner;
  cross-worker resolution is the 6.12 follow-up.
- Mirror handlers in `worker.platform.ts` and `worker.platform.web.ts`
  delegate to the helper after wire-cloning, alongside the existing
  `ResolveDepUris` handler.
- `ResolveDependentUris` added to `AllWorkerRequests` schema unions on
  both platforms so the @effect/platform runner can decode incoming
  requests.

POC ref: design doc §3.3.

@W-22692417@

## Test plan
- [x] Unit tests cover the design's verification matrix:
  - A declares Foo, B/C reference Foo, D unrelated
  - `symbolName` narrowing
  - dedup across multiple symbols
  - self-reference exclusion
  - missing-table skip (cross-worker case deferred to 6.12)
  - empty target file
- [x] Full `apex-lsp-parser-ast` test suite green (2768/3236, pre-existing skips)
- [x] Full `apex-ls` test suite green (130/134, pre-existing skips)
- [x] Repo-wide pre-commit gate green (4244 passed, 0 failed)


peternhale (dismissed):
Solid implementation. The URI normalization for self-reference exclusion is thoughtful, the `symbolName` narrowing is correctly threaded from wire to helper, and the test matrix matches the design doc verification matrix. Approving with nits.

---

**Nit — identical handler bodies in both worker platforms**

The `ResolveDependentUris` handler block in `worker.platform.ts` and `worker.platform.web.ts` is 100% identical. This matches the existing `ResolveDepUris` copy-paste convention, so it may be intentional, but it means future changes (e.g. error handling tweaks) must be made in both files in sync. If a shared internal helper is feasible given the platform split, it would prevent drift. At minimum, consider a comment like `// keep in sync with worker.platform.web.ts` above each block.

---

**Nit — `cloneForWire` on already-serialized data**

`serializeSymbolTable` returns plain-data (`ApexSymbol[]`, `SymbolReference[]`, etc.) and the result is then passed through `cloneForWire`. If `cloneForWire` is a structured-clone, this is redundant. If it strips something non-serializable that can survive in the plain-data types, a short comment stating what it's guarding against would prevent someone from removing the "redundant" call in a future cleanup.

---

**Nit — two-pass URI normalization comment**

```ts
const normalizeForCompare = (uri: string): string => {
  const stripped = extractFilePathFromUri(uri);
  return stripped.startsWith(FILE_URI_PREFIX)
    ? stripped.slice(FILE_URI_PREFIX.length)
    : stripped;
};
```

The second `startsWith(FILE_URI_PREFIX)` check implies `extractFilePathFromUri` can return a string that still carries `file://`. A one-line comment on what `extractFilePathFromUri` returns (and why the second strip is needed) would prevent "wait, doesn't `extractFilePathFromUri` already do this?" confusion for future readers. The test covering the LSP-shaped vs path-shaped asymmetry is great and covers the correctness — the comment would cover the *why*.

---

**Nit — `makeWorkerDispatcher` export for testing**

The new test imports `makeWorkerDispatcher` directly from `WorkerCoordinator`. If this is exported solely to enable testing, an `@internal` JSDoc marker would signal that it's not part of the stable public API and prevent accidental external use.

PR #419, opened 2026-06-03
fix(apex-lsp-compliant-services): address PR 407 completions review findings

kylewalke:
## Summary

1. **Relationship strategy gating** — `RelationshipCompletionStrategy.canHandle` no longer fires after a `.` trigger, and its per-symbol `findRelatedSymbols` fanout is bounded (25 source symbols, 100 candidates) to stay within LSP latency budget.
2. **Strategy overlap** — `General` and `SystemNamespace` `canHandle` predicates tightened (reject trailing-dot lines; require a non-empty word prefix), cutting the number of strategies firing per plain keystroke.
3. **Case-insensitive variable resolution** — `findVariableInScope` matches identifiers case-insensitively, so `myvar.` resolves a field declared `myVar`.
4. **Column-aware, scope-bounded position filter** — variable resolution compares `(line, column)` tuples (translating 0-based LSP positions to 1-based symbol ranges) and requires the match's enclosing scope to contain the cursor; drops the strict-equality fallback that leaked sibling-method variables.
5. **Literal-dot rejection** — `MemberAccessCompletionStrategy.canHandle` rejects numeric (`1.`) and string-literal dots by requiring the preceding char to be a valid receiver (identifier / `)` / `]`).
6. **Cycle guards** — `resolveMethodReturnType`, `resolveFieldType`, `getMembersOfTypeEffect`, and `OverrideCompletionStrategy.collectOverridableMethods` track visited type ids to prevent infinite recursion / stack overflow on cyclic inheritance chains.
7. **Tautological tests removed** — the two completion tests in `queue-integration.test.ts` that asserted a mock was called (never crossing production code) are removed; handler registration remains covered in `index.test.ts`.
8. **Readiness path implemented** — `CompletionProcessingService.processCompletionWithReadiness` is now a real implementation surfacing `isIncomplete` based on enrichment readiness, replacing the previously dead optional-handler branch.

## Test plan

- [x] `lsp-compliant-services`: 694 passed, 10 skipped
- [x] `apex-ls`: 132 passed, 4 skipped, 7 snapshots
- [x] New regression tests: case-insensitive resolution, sibling-scope exclusion, numeric-literal/call-result/index `canHandle`, prefix-gated strategies, `processCompletionWithReadiness` readiness states
- [x] Per-commit pre-commit hooks (lint + tests) ran on each of the 8 commits

### DIFF
--- a/packages/apex-ls/test/queue-integration.test.ts
+++ b/packages/apex-ls/test/queue-integration.test.ts
@@ -260,41 +260,11 @@ describe('Queue Integration in apex-ls', () => {
     });
   });
 
-  describe('completion request processing', () => {
-    it('should process completion requests through queue', async () => {
-      const params = {
-        textDocument: { uri: 'file:///test.cls' },
-        position: { line: 5, character: 10 },
-      };
-
-      const expectedItems = [
-        { label: 'doSomething', kind: 2 },
-        { label: 'toString', kind: 2 },
-      ];
-      mockQueueManager.submitCompletionRequest.mockResolvedValue(expectedItems);
-
-      const result = await mockQueueManager.submitCompletionRequest(params);
-
-      expect(mockQueueManager.submitCompletionRequest).toHaveBeenCalledWith(
-        params,
-      );
-      expect(result).toEqual(expectedItems);
-    });
-
-    it('should handle completion request failures', async () => {
-      const params = {
-        textDocument: { uri: 'file:///test.cls' },
-        position: { line: 0, character: 0 },
-      };
-
-      const error = new Error('Completion processing failed');
-      mockQueueManager.submitCompletionRequest.mockRejectedValue(error);
-
-      await expect(
-        mockQueueManager.submitCompletionRequest(params),
-      ).rejects.toThrow('Completion processing failed');
-    });
-  });
+  // Note: completion request flow is exercised via the wired LCSAdapter
+  // onCompletion handler in index.test.ts ('Protocol Handler Integration').
+  // Mock-only round-trip tests for completion were removed because they
+  // never crossed production code; they asserted the mock framework, not
+  // queue routing.
 
   describe('queue statistics', () => {
     it('should provide queue statistics for monitoring', () => {
--- a/packages/lsp-compliant-services/src/services/CompletionProcessingService.ts
+++ b/packages/lsp-compliant-services/src/services/CompletionProcessingService.ts
@@ -177,11 +177,33 @@ export class CompletionProcessingService implements ICompletionProcessor {
   public async processCompletion(
     params: CompletionParams,
   ): Promise<CompletionItem[]> {
+    const result = await this.processCompletionInternal(params);
+    return result.items;
+  }
+
+  /**
+   * Process a completion request and report whether the result is fully
+   * resolved. The CompletionHandler surfaces the `isIncomplete` flag back to
+   * the editor so it knows to re-query as enrichment completes.
+   */
+  public async processCompletionWithReadiness(
+    params: CompletionParams,
+  ): Promise<{ items: CompletionItem[]; isIncomplete: boolean }> {
+    return this.processCompletionInternal(params);
+  }
+
+  private async processCompletionInternal(
+    params: CompletionParams,
+  ): Promise<{ items: CompletionItem[]; isIncomplete: boolean }> {
     this.logger.debug(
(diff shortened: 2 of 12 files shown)


peternhale (approved):
## Review Summary

All 8 fixes are well-implemented and should be accepted:

✅ **Issue #1**: Relationship fanout - bounds + gating  
✅ **Issue #2**: Strategy overlap - trailing-dot rejection  
✅ **Issue #3**: Case-insensitive variable resolution  
✅ **Issue #4**: Column-aware position filtering & scope validation  
✅ **Issue #5**: Literal-dot rejection  
✅ **Issue #6**: Cycle guards with visited sets  
✅ **Issue #7**: Tautological tests removed  
✅ **Issue #8**: Readiness path implemented  

The fixes demonstrate solid understanding of the issues with appropriate solutions and good test coverage.

---

## Follow-up Recommendation: Remove RelationshipCompletionStrategy

While reviewing issue #1, a deeper concern emerged about **RelationshipCompletionStrategy's fundamental value proposition**.

### The Problem

**RelationshipCompletionStrategy is the ONLY strategy that lacks contextual filtering:**

All other strategies filter based on user context:
- **MemberAccessCompletionStrategy**: Parses expression before `.`, resolves type, returns only that type's members
- **GeneralCompletionStrategy**: Gets current word, uses `resolveSymbol(word, context)` with scope/imports/namespace
- **SystemNamespaceCompletionStrategy**: Filters namespaces/types by prefix match
- **TriggerCompletionStrategy**: Only fires in trigger files
- **OverrideCompletionStrategy**: Returns only overridable methods from parent/interfaces

**RelationshipCompletionStrategy:**
```typescript
getCompletions(context) {
  const fileSymbols = findSymbolsInFile(uri);        // All symbols in file
  const sourceSymbols = fileSymbols.slice(0, 25);    // First 25
  
  for (symbol of sourceSymbols) {
    const related = findRelatedSymbols(symbol, 'method-call');
    suggestions.push(...related);                     // Up to 100 total
  }
  return suggestions;  // NO filtering based on:
  // ❌ Cursor position
  // ❌ Current scope (which block/method cursor is in)
  // ❌ What user is typing
  // ❌ Variables in scope
  // ❌ Parent scopes
  // ❌ Expected type
}
```

### Conceptual Mismatch

**Relationships are for refactoring/navigation:**
- `findRelatedSymbols(symbol, 'method-call')` finds "where is this symbol called?"
- Used correctly in `ReferencesProcessingService` for "Find References"
- Answers: "Show me all call sites" (refactoring context)

**Completions are for coding assistance:**
- Should answer: "What can I type at this cursor position given my context?"
- Requires scope awareness, user input correlation, type information

**RelationshipCompletionStrategy mixes these concepts** - uses refactoring infrastructure for completions where it doesn't fit.

### Real-World Impact

**Scenario: User opens new line in a method**

Proper completion (GeneralCompletionStrategy):
- Suggests symbols in scope: local variables, parameters, fields, methods
- Filtered by scope chain (current block → method → class → imports)
- Matched against what user is typing

RelationshipCompletionStrategy:
- Suggests 100 random call sites from elsewhere in the file
- No correlation to cursor context or what user needs

**Even with PR 419's dot-gating**, it still fires during general typing and provides noise rather than contextually relevant suggestions.

### Evidence

1. **No contextual filtering** - zero consideration for cursor position, scope, or user input
2. **Conceptual mismatch** - relationship graphs inform refactoring, not completions
3. **Not in original jorje LSP** - this is a new, unvalidated feature
4. **Redundant** - GeneralCompletionStrategy already handles general typing with proper context awareness

### Recommendation

**Remove RelationshipCompletionStrategy entirely** in a follow-up PR.

Two options:
- **Option A (safer)**: Disable via `canHandle() { return false; }` - keeps code for potential future investigation
- **Option B (cleaner)**: Delete the file and remove from strategy list

I recommend **Option A** - it's reversible and documents the decision if someone later identifies a valid use case.

### Questions for the Team

1. What developer problem was RelationshipCompletionStrategy intended to solve?
2. Was this feature driven by user research or is it speculative?
3. Has anyone validated that the current implementation provides value to users?

If there's a valid use case we haven't identified, it would need:
- Clear documentation of the developer value
- Contextual filtering (like all other strategies)
- User input correlation (filter based on what they're typing)

---

**Bottom line:** PR 419's fixes are all correct. Approve and merge. But let's discuss removing RelationshipCompletionStrategy in a follow-up.

PR #451, opened 2026-06-13
feat(apex-lsp-extension): worker references enrichment + LSP cancellation - W-22692418

kylewalke:
### What does this PR do?

Story 6.05 of the Find References offload epic. Replaces the bare `DispatchReferences` passthrough on both worker platforms (`worker.platform.ts`, `worker.platform.web.ts`) with a full enrichment-style handler that mirrors hover/definition:

`loadSymbolDataForEnrichment` → `loadDependentsForReferences` → `processReferences` → `writeBackEnrichedSymbols`, at `'protected'` detail level.

- **`loadDependentsForReferences(svc, uri, symbolName?)` helper** on both platforms. Where `loadSymbolDataForEnrichment` pre-fetches *outbound* deps (files this file references), Find References needs the *inbound* direction — the caller-side tables whose declared symbols reference the target. It loads them via the `dataOwner:ResolveDependentUris` assistance call (added in 6.04). Best-effort: a failed resolve leaves the graph partial and the search proceeds on already-loaded tables.
- **`context.includeDeclaration`** honored end-to-end (already threaded through `processReferences` → `getReferenceLocations`).
- **Coordinator-side LSP `CancellationToken` support** so rapid cursor moves don't queue dead work behind active requests. The token is threaded from `onReferences` through `submitReferencesRequest` into `submitRequest`, which short-circuits a request cancelled before dispatch and interrupts the in-flight Effect fiber on mid-flight cancellation. A new `RequestCancelledError` surfaces as the `null` fallback in `LCSAdapter.handleLspRequest` without an error log. (Cross-boundary worker cancellation is intentionally out of scope — it would require a wire-schema change.)

@W-22692418@

Builds on #418 (6.04). Part of epic [IDE Apex Language Support - findReferences].


peternhale (approved):
Approving — solid work that faithfully mirrors the established hover/definition enrichment pattern, and the cancellation layer correctly handles the Effect interrupt semantics. Verified locally: both platform helper bodies are byte-identical, the `RequestCancelledError`/`CancellationLike` export chain is sound, `token` is backward-compatible (optional, single caller), and the 4 new cancellation tests pass.

A few non-blocking nits for a follow-up (or now, your call):

**nit — cancellation pollutes observability success-rate.** In `LCSAdapter.runWithSpanAndRecord`, the catch sets `success = false` and records the span as a failure *before* rethrowing, so `handleLspRequest` swallowing the `RequestCancelledError` happens too late — the metric is already counted as a failure. Since cancellation is expected control flow here, rapid cursor movement will inflate the references failure rate. Consider treating `RequestCancelledError` as a non-failure (or a distinct "cancelled" outcome) in the recorder.

**nit — `ingested` count computed then discarded.** `loadDependentsForReferences` returns a count that the caller ignores. Either `logger.debug` it (handy "loaded N dependent tables" diagnostic) or drop the return value so it doesn't imply it's consumed.

**nit — duplicated `any`-heavy raw shape.** The `raw` cast (`symbols: any[]; references?: any[]; ...`) is now duplicated verbatim in `loadDependentsForReferences` and the existing `ingestEntries`, across both platform files. A shared named type (e.g. `SerializedSymbolTableData`) feeding `SymbolTable.fromSerializedData` would remove the duplication and the `any`s. Pre-existing pattern, just doubled here.

**nit — silent best-effort swallow.** The bare `catch {}` in `loadDependentsForReferences` returns 0 on any failure (intentional, matches the existing dep-prefetch pattern). A `logger.debug` on the swallow would help diagnose future "cross-file references missing" reports.

**nit — test gap.** Cancellation is well-covered at the queue layer, but nothing exercises `loadDependentsForReferences` / the new `DispatchReferences` enrichment flow — neither the success path (dependents ingested → cross-file refs found) nor the best-effort failure path. The riskier change is the less-tested one; an integration-level test asserting cross-file references appear after enrichment would be worth adding.

Out-of-scope-but-noted: cross-boundary worker cancellation isn't handled (acknowledged in the PR body) — the worker still runs the now-heavier reference search to completion and only the result is discarded. Reasonable follow-up candidate given enrichment just got more expensive.

kylewalke:
I'll address nits on the next feature branch.

PR #462, opened 2026-06-16
feat(apex-lsp-extension): route references through enrichment pool + #451 nits - W-22692419

kylewalke:
### What does this PR do?

Activates the 6.06 references-offload path and folds in the PR #451 review follow-ups on the same change set.

**6.06 — references dispatch**
- Flip `references` routing from `coordinatorOnly` to `enrichmentPool` in `WorkerCoordinator`, threading `includeDeclaration` end-to-end.
- Consult the worker dispatcher in `LSPQueueManager` before falling back to the coordinator-local handler.
- Wait for the batch-ingestion dispatcher (bootstrap race) in `WorkspaceBatchHandler`, with an injectable timeout for tests.

**PR #451 review nits**
- Skip the observability metric for `RequestCancelledError` so superseded cursor moves no longer inflate the references failure rate.
- Log the dependent-table count loaded for Find References.
- Extract a shared `SerializedSymbolTableData` type, removing the duplicated `any`-heavy raw cast across both worker platform files.
- Log the best-effort dependent-prefetch swallow at debug.
- Add unit coverage for `loadDependentsForReferences` (ingest contract + best-effort failure paths). The live end-to-end path (real topology + coordinator assistance bus) remains the skipped test owned by W-22692429 (6.13).

@W-22692419@


peternhale (dismissed):
Solid, well-tested PR. The non-test changes are small and surgical; the bulk is well-constructed test coverage. No correctness blockers — findings below are minor and none are blocking.

### What this does well

- **Cancellation metric fix** ([LCSAdapter.ts:482-501](packages/apex-ls/src/server/LCSAdapter.ts#L482-L501)): skipping `aggregator.record` entirely for `RequestCancelledError` is correct — a superseded request is neither success nor failure, and recording it would inflate the failure rate or pollute duration percentiles. `throw e` is preserved so cancellation still propagates. Excellent explanatory comment.
- **`SerializedSymbolTableData` extraction** ([symbol.ts:824-830](packages/apex-parser-ast/src/types/symbol.ts#L824-L830)): removes four duplicated `any`-heavy inline casts and gives `fromSerializedData` a named contract.
- **`loadDependentsForReferences` injectable fetcher**: textbook way to make an assistance-bus-dependent function unit-testable without faking the bus; the default parameter keeps callers unchanged.
- **Test coverage is genuinely thorough** — the new `loadDependentsForReferences.node.test.ts` pins the ingestion contract (null-entry skipping, missing `entries`, both best-effort failure paths). The dispatcher-consultation tests cover all four branches (willing, declines, unavailable, failure).

### Suggestions (non-blocking)

**1. Add `afterEach` dispatcher reset in `LSPQueueManager.test.ts` — cheapest/highest value.**
[LSPQueueManager.test.ts:888+](packages/lsp-compliant-services/test/queue/LSPQueueManager.test.ts) — each test calls `manager.setWorkerDispatcher({...})` but there's no teardown resetting it to `null`. Since `getInstance()` returns a shared singleton, a leftover dispatcher could leak into a later test. The four new tests each set their own dispatcher so they're internally safe, but an `afterEach(() => manager.setWorkerDispatcher(null))` would prevent cross-test bleed — the way `WorkspaceBatchHandler.test.ts` correctly resets in its `afterEach`.

**2. Introduce a `WorkerDispatcher` interface to replace the `any`.**
[LSPQueueManager.ts:217-220](packages/lsp-compliant-services/src/queue/LSPQueueManager.ts#L217-L220) — the new dispatch branch leans on `workerDispatcher?.isAvailable?.() && workerDispatcher.canDispatch?.(type)` against an `any`-typed field. The optional-chaining defends against shape drift at runtime, but a real interface (`isAvailable()`, `canDispatch(type)`, `dispatch(type, params)`) would catch mismatches at compile time and document the protocol. The PR didn't introduce the `any`, but it's now load-bearing.

**3. Consider `Deferred`-based wiring (or `info`-level timeout log) for the batch dispatcher wait.**
[WorkspaceBatchHandler.ts:563-572](packages/apex-ls/src/server/WorkspaceBatchHandler.ts#L563-L572) — the 50ms poll loop over a module-level mutable works and is easy to reason about, but the more idiomatic Effect approach would be having `setBatchIngestionDispatcher` complete a `Deferred` that this awaits, eliminating the poll and the up-to-50ms latency after wiring. Separately: when the 5s window elapses, batches silently fall through to coordinator-local processing — the safe fallback, but it defeats the offload silently. Consider logging the timeout-elapsed case at `info` rather than `debug`, since it signals a missed optimization rather than routine flow.

### Risk note

The dispatch layer itself is well-tested, but the **live cross-file resolution** path is only exercised by the skipped integration test. The PR body and test comments are appropriately honest about this — real coverage is deferred to W-22692429 (6.13). The skipped test is correctly documented with its reason (assistance bus not wired in the isolated topology) and a tracked follow-up. Make sure W-22692429 lands before relying on cross-file references in production.

**Verdict: Approve** with the minor suggestions above.

peternhale (approved):

PR #463, opened 2026-06-16
fix(apex-lsp-compliant-services): identifierRange + fileUri output-shape fixes - W-22692420

kylewalke:
Fixes the output-shape bugs that caused find-references to silently drop every `Location` it produced.

### What changed
- `ReferencesProcessingService` and `WorkspaceSymbolProcessingService` location helpers now read `.location.identifierRange.{startLine,startColumn,endLine,endColumn}`. The previous code read flat fields off `SymbolLocation` that don't exist there, yielding `NaN` → `null` after the worker JSON round-trip.
- File-URI helpers read the canonical `fileUri` first, falling back to `file://${filePath}` for legacy wire-shape variants (fallback order unchanged).
- **Removed the blanket `any` typing on these helpers** — the root cause of the original bug was that the wrong-field-name read was invisible to the compiler. They're now typed against `ApexSymbol` / `ReferenceResult` with small, documented wire-variant types for the legacy `filePath` field, so a future wrong-field-name fails at compile time.

### Tests
- Unit tests covering same-file and cross-file location shapes plus `fileUri`-vs-`filePath` fallback ordering, in both services. 718 passing in the package.

@W-22692420@


peternhale (changes requested):
## Requesting changes — `filePath` fallback is dead code mislabeled as "legacy"

The `identifierRange`/`fileUri` bug fix itself is correct and well-tested — no concerns there. But the `filePath` fallback added alongside it needs to be removed before merge.

### Issue (must fix)

The PR introduces a fallback to a `filePath` field on the file-URI helpers and frames it as a "legacy wire-shape variant." There is **no producer of that shape**, in this repo or out of it (confirmed with the author):

- Neither canonical type declares `filePath`. `ApexSymbol` has no `filePath` field; `ReferenceResult.fileUri` is the canonical field, no `filePath`.
- The only reason `filePath` compiles in these helpers is the PR manually widening the types with `& { filePath?: string }`.
- No producer in the repo writes `filePath` onto these objects. The `filePath` on `WireResolvedQualifier` (`wireSchemas.ts`) is an unrelated type that never reaches this code path.
- Its only historical basis is a 2023 commit (`23dcec60`) where the symbol type was structured differently — since renamed to `fileUri`.
- The new "falls back to filePath" tests only pass because the literals are cast through `(service as any)`; they can't be constructed as real `ApexSymbol`/`ReferenceResult`, so they prove the branch *runs*, not that any caller hits it.

So this is defensive code against a contract that does not exist, and the "legacy" label implies a backward-compat obligation that isn't real. Two concrete harms:

1. It weakens the exact protection this PR set out to add. The point of removing the blanket `any` was "a wrong field name fails at compile time." The `& { filePath?: string }` intersection re-opens a hole in that guarantee for a field that should never be read.
2. "Legacy" is self-perpetuating: the next maintainer assumes an unseen producer exists and won't delete it either.

### Requested change

Remove the dead fallback:
- Drop the `& { filePath?: string }` intersections (the `WireSymbol` / `WireReference` types can alias `ApexSymbol` / `ReferenceResult | ApexSymbol` directly, or be inlined).
- Drop the `if (...filePath)` fallback branches in `getSymbolFileUri` and `getReferenceFileUri` in both services.
- Drop the four "falls back to filePath" tests that cover the removed branches.

Rely solely on the canonical `fileUri`. If a real cross-process producer of `filePath` ever appears, it will arrive with a type, a schema, and a failing test — add the fallback then, against something concrete, rather than pre-paying for a wire shape we can't name.

peternhale (approved):

PR #464, opened 2026-06-16
feat(apex-lsp-parser-ast): cross-file edges + kind-aware findReferencesTo - W-22692421

kylewalke:
Populates cross-file incoming edges on the data-owner worker and makes `findReferencesTo` dispatch by symbol kind — the foundation of find-references algorithm completeness. Without this, `findReferencesTo` only sees same-file edges.

### What changed
- **`dataOwner:UpdateSymbolSubset`** now calls `resolveCrossFileReferencesForFile(uri)` after `addSymbolTable` (both `worker.platform.ts` and `worker.platform.web.ts`).
- **New `DrainDeferredReferences` wire request**: `WorkspaceBatchCompile` invokes `dataOwner:DrainDeferredReferences` at the end of the batch (both platforms). Best-effort — a drain failure does not fail the batch compile.
- **`ApexSymbolRefManager.drainAllDeferredReferencesSync`**: in-place, scheduler-free walk of the deferred map; an async `Effect` wrapper is retained for Effect-genned call sites.
- **Kind-aware `findReferencesTo` dispatch**: locals walk the declaring file; field/property/static-method/constructor/type use the graph reverse index; instance methods use an override-aware traversal (superclass → interfaces → children → implementors) with a visited-types cycle guard and de-duped results.
- **Related-type discovery is single-pass**: a reverse-adjacency map built once + BFS — O(N + edges) per request, not a multiplicative fixpoint scan.

### Tests
- `CrossFileEdgePopulation.test.ts`: deferral → later file added → drain populates the incoming edge; locals path; 3-level override union (asserts the cross-file caller ref surfaces when querying both a base and a derived override); inheritance-cycle guard. 11/11 in the file.

### Notes
- The WI mentioned a `DYNAMIC` symbol kind that does not exist in this codebase; the graph path covers field/property/static-method/constructor/type as intended.
- The override-union test seeds a resolved caller edge directly: a qualified, receiver-typed call (`b.doWork()`, `b: Base`) does not yet resolve to the cross-file member — that receiver-typed resolution is upstream of `findReferencesTo` and tracked by later epic items (6.09 / 6.11). The seeded edge exercises the in-scope union logic against non-empty data.
- `findMethodOnType` does not disambiguate overloaded methods sharing a name on one type (returns first match); documented as a known imprecision deferred to a later story.

@W-22692421@


peternhale (commented):
## Review note — architectural intersection with W-23006798 (implementor-references)

Flagging a cross-WI intersection that should be reconciled before this lands, because both pieces of work compute the **same** inheritance/implementor relationship by **different** mechanisms, and they converge on the same data-owner hooks.

### The overlap

This PR's instance-method path discovers related types **at query time** by scanning every type in the workspace:
- `collectRelatedTypeNames` -> `collectAllTypeSymbols()` builds a parent->child reverse-adjacency map by reading the `superClass`/`interfaces` **string arrays** on every `TypeSymbol`, then BFS. This is the O(workspace types) per-request cost noted in the perf section.

W-23006798 (`feature/W-23006798-implementor-references`, Phase 1 commit `25815e4f`) makes implements/extends **first-class reverse-index edges at ingest time**, specifically so `findReferencesTo(interface/superclass)` returns implementors/subclasses **directly from the reverse index** — no scan:

> Make implements/extends relationships first-class reverse-reference edges so `findReferencesTo(interface)` returns implementors ... Previously implements/extends were stored only as string arrays on TypeSymbol and emitted as generic TYPE_DECLARATION references.

It adds `ReferenceContext.INHERITANCE` / `INTERFACE_IMPLEMENTATION`, maps them to `ReferenceType.INHERITANCE` / `INTERFACE_IMPLEMENTATION` in `ApexSymbolManager`, and resolves them through the same `resolveCrossFileReferencesForFile` hook this PR wires into the worker. Phase 2 (`c66212aa`) reuses `loadDependentsForReferences` + `resolveCrossFileReferencesForFile` — the exact hooks added here.

### Why it matters

1. **Two sources of truth for one relationship.** After both land, the inheritance graph exists as reverse-index edges (W-23006798) *and* as a query-time scan of string arrays (this PR). They can diverge.
2. **The O(N) scan is dead weight against W-23006798.** Once inheritance edges live in the reverse index, the scan recomputes in O(N) what the index answers in O(result). W-23006798 is the actual fix for the perf nit — not a "later" optimization.
3. **Double-counting risk.** W-23006798 emits inheritance as `INHERITANCE`/`INTERFACE_IMPLEMENTATION` reference edges. This PR's `findReferencesViaGraph` reads the same reverse index, so once both are present the instance-method union may surface the inheritance edge itself as a "reference to the method." Needs verification that `findReferencesViaGraph` filters by reference type, or it will return inheritance edges as method references.
4. **Merge collision.** Both branches modify `ApexSymbolManager.ts` resolution, the worker platform files, and `workerWireSchemas.ts`.

### Request

Rebase this PR's related-type discovery onto W-23006798's reverse-index edges rather than reimplementing discovery by scanning `superClass`/`interfaces` string arrays. Concretely: once W-23006798's edges exist, `collectRelatedTypeNames`/`collectAllTypeSymbols` should consume the reverse index (treating W-23006798 as the canonical inheritance source) instead of the per-request whole-graph scan. If sequencing means this PR lands first, please add a tracking item to remove the scan when W-23006798 merges, and confirm `findReferencesViaGraph` won't return the new INHERITANCE/INTERFACE_IMPLEMENTATION edges as method references.

(Raising for visibility — the bug fix and override-union logic themselves look sound; this is about not committing to a query-time scan the implementor-references work already replaces with maintained edges.)

kylewalke:
CC: "Thanks for flagging the W-23006798 intersection, Peter. Both PRs are still open against `main` (this is #464; implementor-references is #472), so sequencing isn't yet fixed. Addressed per your conditional in 647df089c (comment-only, no behavior change):

**Concern #3 (double-counting) — verified it cannot occur.** W-23006798's `INHERITANCE`/`INTERFACE_IMPLEMENTATION` edges target the **type** symbol (implementor→interface / subclass→superclass), so they land in the *type's* reverse-index bucket. `findInstanceMethodReferences` only ever reads **method** buckets — `findReferencesViaGraph(symbol)` and `findReferencesViaGraph(methodOnType)`, both keyed on a method declaration. Inheritance edges therefore can't surface here as method references, so no type-based filtering of `findReferencesViaGraph` is needed. Documented this on `findInstanceMethodReferences`.

**Concerns #1/#2 (two sources of truth, O(N) scan as dead weight) — tracked in code.** Added a follow-up note on `collectRelatedTypeNames`: once W-23006798's maintained edges exist on `main`, the query-time whole-graph scan (`collectAllTypeSymbols`) should be replaced by a reverse-index lookup keyed on the declaring type, treating W-23006798 as the canonical inheritance source, so the scan doesn't outlive the edges and the two can't diverge. I can't do the actual rebase-onto-edges now since those edges aren't on `main` yet — whichever of #464/#472 merges first, the second will reconcile (the swap if #472 is second, or executing this tracked follow-up if #464 is second).

**Merge collision (#4):** noted — #464 and #472 both touch `ApexSymbolManager.ts`, the worker platform files, and `workerWireSchemas.ts`;"

kylewalke:
CC: "Follow-up done — the related-type discovery now consumes #472's edges directly (02dd224eb).

The merge from main pulled in #472 (W-23006798) — it merged ~12 min before this branch's merge-from-main — so the precondition in my earlier reply (\"once W-23006798's maintained edges exist on main\") now holds, and I executed the swap rather than leaving it tracked.

**What changed:** `collectRelatedTypeNames` no longer scans `superClass`/`interfaces` string arrays across all workspace types (`collectAllTypeSymbols` is deleted). It now walks #472's INHERITANCE / INTERFACE_IMPLEMENTATION reverse-index edges:
- **ancestors** = a type's OUTGOING such edges (`findSupertypesViaEdges`)
- **descendants** = a type's INCOMING such edges (`findSubtypesViaEdges`)

This resolves all four of your concerns:
1. **Two sources of truth** → gone. The edges are the only inheritance source; nothing recomputes the graph from string arrays.
2. **O(N) scan as dead weight** → gone. Each traversal step is O(edges incident to the type), not O(workspace types) per request.
3. **Double-counting** → still cannot occur, now verified in code: `findSupertypesViaEdges`/`findSubtypesViaEdges` filter to INHERITANCE/INTERFACE_IMPLEMENTATION and to type-group symbols; the instance-method union still reads only *method* buckets.
4. **Merge collision** → resolved by the merge from main; this commit sits on top of #472.

Verification: 11/11 CrossFileEdgePopulation (incl. the 3-level override up-walk and implementor down-walk, which now exercise the real resolved edges) + 400 symbols/references tests green; full pre-commit suite 4499 passed / 0 failed."

peternhale (changes requested):
## Review — cross-file edges + kind-aware `findReferencesTo`

Strong PR overall: the logic is correct, the wire plumbing is consistent across both platforms, and the tests are genuinely good (the sibling-implementor direction-pinning test would actually catch a supertype/subtype swap). Routing is clean, and the coordinator dispatcher case correctly forwards `DrainDeferredReferences` to the data owner without transforming anything.

I'm requesting changes around **one** thing: parser-ast inheritance-graph logic now exists in two divergent implementations at two architectural layers. The rest below are non-blocking notes.

---

### 🔴 Blocking: duplicated inheritance traversal — `apex-parser-ast` vs `lsp-compliant-services`

This PR adds edge-based inheritance traversal to `ApexSymbolRefManager` (`collectRelatedTypeNames` → `findSupertypesViaEdges` / `findSubtypesViaEdges`), and its own JSDoc argues edges are "the canonical inheritance source… one source of truth for the graph and no per-request whole-workspace scan."

But `ImplementationProcessingService` (in `lsp-compliant-services`) **independently reimplements the same traversal**, using `superClass`/`interfaces[]` string matching plus a full-workspace `getAllSymbolsForCompletion()` scan:

| Concern | New, in `apex-parser-ast` (this PR) | Pre-existing, in `lsp-compliant-services` |
|---|---|---|
| interface → sub-interface walk | `collectRelatedTypeNames` via `findSupertypesViaEdges` (graph edges) | `collectInterfaceHierarchy` via `interfaces[]` string match |
| find implementors / subclasses | `findSubtypesViaEdges` (incoming INHERITANCE / INTERFACE_IMPLEMENTATION edges) | `findDirectSubclasses` / `findImplementingClasses` via `superClass` / `interfaces[]` string match |
| transitive descent | `collectRelatedTypeNames` down-walk, cycle-guarded | `findExtendingClasses` BFS |

Consequences:

- **Two inheritance graphs of truth, two layers.** A future change to how inheritance is modeled (generics, namespace-qualified supertypes, …) must be made in both `findSupertypesViaEdges`/`findSubtypesViaEdges` *and* `collectInterfaceHierarchy`/`findDirectSubclasses`, in two packages.
- **The two paths can disagree.** The ref-manager path is edge-driven (and now drains deferred edges on read); the service path is string-array-driven with a full-scan fallback. For the same hierarchy, go-to-implementation (service path) and find-references-on-an-interface-method (ref-manager path) can return different implementor sets — a subtle, hard-to-diagnose user-visible inconsistency.
- **The redundant fallback this PR's premise eliminates is still running.** `ImplementationProcessingService.findImplementingClasses` / `findDirectSubclasses` still call `getAllSymbolsForCompletion()` and scan **every class in the workspace** — exactly the O(workspace) scan this PR claims to have removed, sitting one layer up. Index completeness is added here without retiring the now-redundant scan, so both run.

**Requested change.** Not asking to rewrite the world, but the boundary should hold and this PR is the right moment since it establishes the canonical mechanism. Either:

1. **(Preferred)** Expose the edge-based traversal from `apex-parser-ast` as first-class `ISymbolManager` API — e.g. `findSupertypes(type)` / `findSubtypes(type)` / `findImplementors(interface)` built on `collectRelatedTypeNames` (currently `private`) — have `ImplementationProcessingService` call those, and delete `collectInterfaceHierarchy` / `findDirectSubclasses` / `findExtendingClasses` plus the `getAllSymbolsForCompletion()` fallbacks. One inheritance graph; find-references and go-to-implementation provably agree.
2. **(Minimum)** If consolidation is out of scope for the WI, add a cross-reference comment in both implementations noting they must stay in sync, **and** a tracking work item to unify — so the drift doesn't silently widen with the next inheritance change.

---

### 🟡 Non-blocking

**Drain-on-read runs on the request worker's local graph, not the data owner's — and the doc implies otherwise.** `references` routes to `requestPool`, so `findReferencesTo` → `findInstanceMethodReferences` → `drainAllDeferredReferencesSync()` mutates the *request worker's* transient subset graph (built from `loadDependentsForReferences`), not the data owner's. That's defensible (it doesn't corrupt the owner), but the JSDoc's "near-free… once the post-batch drain has run the deferred map is empty / this is a no-op" only holds for the **data owner**; the post-batch `DrainDeferredReferences` hit the owner, not this worker. Please correct the comment to state which graph it operates on so a future reader doesn't assume the wrong invariant.

**Two parallel deferred-resolution mechanisms.** `drainAllDeferredReferencesSync` hand-rolls the resolve-source / pick-target / create-edge logic rather than reusing the existing Effect/scheduler path. Notably `addReference` normalizes block-symbol sources to their containing declaration before `createAndAddReference`; `drainAllDeferredReferencesSync` does not — a latent divergence. Consider factoring the shared "resolve + place edge" core, or cross-linking the two so they don't drift.

**`getEnclosingScopeRange` → `undefined` widens the locals fallback.** When scope can't be determined the name-match fallback stops being scope-bounded, re-opening the cross-contamination this feature guards against. Low risk (resolved refs match by id), but worth a one-line comment that it's a deliberate precision/recall tradeoff.

**`findMethodOnType` first-overload match** is documented as deferred — fine, but it means find-references on an overloaded method can union edges from the wrong overload across the hierarchy. The honesty is appreciated.

**Override-union test seeds the caller edge directly** (receiver-typed `b.doWork()` resolution is upstream / 6.09–6.11). So the union *logic* is tested with synthetic input; the end-to-end "rename a virtual method, find all real call sites" path isn't user-observable yet. Fine for the WI scope — just flagging it isn't exercised end-to-end.

---

### 🟢 Follow-up (docs, not this PR)

The architecture skill's routing table lists `references` as `coordinatorOnly`; the code says `requestPool`. The doc is stale and should be corrected — the data-ownership analysis hinges on which it is.

peternhale (approved):
## Re-review — concerns addressed ✅

Re-reviewed at `2b81a36`. The blocking item is fully resolved, and you took the **preferred** path (consolidate), not the minimum. Approving.

### 🔴→✅ Inheritance traversal drift — resolved by consolidation

- New first-class API in `apex-parser-ast`: `ISymbolManager.findSubtypes(type)` / `findSupertypes(type)`, delegating `ApexSymbolManager` → `ApexSymbolRefManager`. Both are transitive, cycle-guarded, and built on a single shared `collectTransitiveTypes` spine over `findSubtypesViaEdges` / `findSupertypesViaEdges`. (e3daf50b)
- `ImplementationProcessingService` now calls `findSubtypes(...)` and **deletes** `findImplementingClasses`'s hierarchy BFS, `collectInterfaceHierarchy`, `findExtendingClasses`, and `findDirectSubclasses` — along with every `getAllSymbolsForCompletion()` whole-workspace scan. The only remaining mention of that scan is a comment describing what was removed. (a254cef8)
- Critically, the find-references path was refactored too: `collectRelatedTypeNames` no longer has its own up/down walk — it now consumes the same public `findSupertypes` / `findSubtypes`. So there is genuinely **one** transitive-inheritance implementation feeding both go-to-implementation and find-references; they provably agree. That closes the correctness-divergence and the redundant-fallback concerns together.

### 🟡→✅ Non-blocking items

- **Drain-on-read graph locality** — the `findSubtypes` / `findSupertypes` JSDoc now states plainly that the drain operates on *this* manager's graph, which in the worker topology is the request worker's transient subset (built per request from `loadDependentsForReferences`), not the data owner's, and that cost scales with the local deferred map rather than the workspace. Exactly the clarification I asked for. (0b3fde3a)
- **Tests** — `ImplementationProcessingService.test.ts` re-expressed against the `findSubtypes` delegation (sub-interface hierarchy, grandchild transitivity preserved); new coverage for the `DrainDeferredReferences` coordinator relay and web dependent prefetch. (2b81a367)

### Still-open (as agreed, not blocking)

- `findMethodOnType` first-overload imprecision and the synthetic-seed override-union test remain deferred to the receiver-typed-resolution stories (6.09 / 6.11) — unchanged scope, fine.
- Architecture-skill routing-table doc (`references` listed as `coordinatorOnly`, code says `requestPool`) is a docs follow-up on our side, not this PR.

Nice work — the consolidation made the codebase smaller and removed the O(workspace) scan rather than just papering over the duplication. LGTM.

PR #465, opened 2026-06-17
feat(apex-lsp-compliant-services): chainNodes + position-based reference resolution - W-22692425

kylewalke:
Resolves cursor-on-inner-name in chained references like `GeocodingService.GeocodingAddress`, where the fully-qualified dotted name has no entry in `findSymbolByName` and references were silently empty.

> **Stacked on #463 (6.07).** Base branch is `feature/W-22692420-output-shape-fixes`; merge after #463. Builds on 6.07's `fileUri`-canonical / `identifierRange` output-shape fix.

### What changed
- `ReferencesProcessingService.findReferences` now picks the name under the cursor by walking `chainNodes` (new `pickNameUnderCursor` / `isPositionWithinIdentifier`), falling back to the leaf identifier of a dotted name.
- Resolution order: `getSymbolAtPosition(uri, pos, 'precise')` (position-based, identifier-anchored — no containing-scope fallback) → context-aware `resolveSymbol` on the name under the cursor.
- `isPositionWithinIdentifier` reuses the shared `isPositionWithinLocation` helper (newly exported from the `apex-parser-ast` barrel) rather than a private bounds copy.

### Tests
- Cursor positions: leaf type, qualifier segment, and plain unqualified name, plus a dotted-name leaf-fallback (`pickNameUnderCursor`) and a containing-scope-leak guard that asserts the `'precise'` strategy is requested. 26 passing in the references suite.

### Review-driven simplification (Pete's #465 review)
Two scaffolding paths from the initial cut were **removed** as part of addressing the review, so they are intentionally **not** on the branch:
- **Overload selection by signature** (`disambiguateOverload` / `methodSignatureKey`) — deleted. It ran a workspace-wide `findSymbolByName` on every method find-references but had no observable effect on the returned locations (see data-layer note below). The real per-overload separation will land with the 6.11 signature-keyed `findReferencesTo` work.
- **Dotted-name last-segment retry** — deleted. It was unreachable: `symbolName` always comes from `pickNameUnderCursor`, which only ever returns dot-free names, so the `symbolName.includes('.')` guard could never fire. Removing it also removes the latent mis-resolution-to-an-unrelated-same-named-symbol hazard.

The overloaded-method fixture (`process(Integer)` / `process(String)`) remains in `ChainedRefTestClass.cls` as a method-call site for the plain-name resolution test.

### Known limitation (data layer)
Returning fully distinct reference *Location sets* per overload isn't achievable in this service today: `findReferencesTo` is keyed by `(name, fileUri)` (cache key `refs_to_${name}`; target re-resolved via `findSymbolInFileByName`), so both overloads of a method in the same file resolve to the same reference set at the data layer regardless of which declaration instance the service selects. End-to-end per-overload reference separation needs parser/symbol-manager changes (call-site arg types on `SymbolReference` + signature-keyed `findReferencesTo`) — tracked as a 6.11 follow-up.

@W-22692425@


peternhale:
## Code review findings (W-22692425)

Reviewed at medium effort. The candidate correctness bugs were all **refuted** on verification (`getSymbolAtPosition` correctly returns `ApexSymbol | null` so the direct assignment is fine; the no-params signature key produces a distinct `foo()`; and a leaf-segment keyword can't occur since Apex forbids reserved words as member names). What remains is quality / altitude / reuse — listed most-actionable first.

### 1. `disambiguateOverload` runs a workspace-wide lookup for zero observable effect — `ReferencesProcessingService.ts:507`
Its own JSDoc states it "has NO observable effect on the returned Locations" until the 6.11 signature-keyed `findReferencesTo` lands. Until then, every find-references on a method pays for a global `findSymbolByName` lookup + filter on a hot path that cannot change the result. Suggest gating it behind the downstream work (feature branch / flag) rather than shipping a no-op.

### 2. `rangeContainsPosition` re-implements the exported `isPositionWithinLocation` — `ReferencesProcessingService.ts:466`
`apex-parser-ast/src/symbols/ops/positionUtils.ts` already has `isPositionWithinLocation` with identical inclusive-bounds logic in the same parser coordinate system. It isn't currently exported from the package barrel (`index.ts`), so reuse means adding that export, then having the two call sites (`symbolIdentifierContainsPosition`, `isPositionWithinIdentifier`) call it via the symbol/node `location`. Removes a maintenance fork. (Note: `MemberAccessCompletionStrategy` has a third copy, but it's pre-existing and uses 0-based LSP coords + `symbolRange`, a different contract — not in scope here.)

### 3. The resolution cascade + identifierRange guard are a service-layer bandaid — `ReferencesProcessingService.ts:318`
The guard rejecting step-1 results whose `identifierRange` doesn't contain the cursor exists to compensate for `getSymbolAtPositionWithinScope`'s Step-4 fallback returning any containing-scope symbol. But `getSymbolAtPosition` already accepts a `'precise'` strategy that does identifier-only resolution with **no** containing-scope fallback — which is exactly what this guard hand-rolls, and what `HoverProcessingService` / `DefinitionProcessingService` already pass. Passing `'precise'` here would let you delete the guard and `symbolIdentifierContainsPosition`. (Do **not** tighten the shared Step-4 fallback itself — `ImplementationProcessingService` intentionally relies on the permissive `'scope'` behavior.)

### 4. The dotted-name leaf retry can mis-resolve to an unrelated same-named symbol — `ReferencesProcessingService.ts:351`
When the cursor is on `Leaf` in `Outer.Leaf` and position/context resolution fail, the retry calls `resolveSymbol('Leaf')` with no constraint that the match is a member of `Outer`. A public top-level `Leaf` elsewhere in the workspace can be returned, surfacing the wrong references. The `leafPicked` gate prevents the qualifier case but not this cross-symbol collision — the comment already acknowledges the risk. Consider constraining the retry to members of the resolved qualifier, or dropping the retry when the qualifier doesn't resolve.

Items 1–4 are quality/altitude, not crashes. Happy to discuss any of them.

peternhale (changes requested):
Requesting changes on the quality/altitude findings below (no correctness bugs survived verification — the candidate crashes were refuted). Details in the inline review comment already posted.

1. **`disambiguateOverload` is a no-op on a hot path** (`ReferencesProcessingService.ts:507`) — its own JSDoc says it has no observable effect until the 6.11 signature-keyed `findReferencesTo` lands, yet it runs a workspace-wide `findSymbolByName` on every method find-references. Gate it behind the downstream work.

2. **`rangeContainsPosition` duplicates the exported `isPositionWithinLocation`** (`:466`) — reuse the parser-ast helper (export it from the barrel) instead of a third copy of the same inclusive-bounds logic.

3. **Service-layer bandaid for the resolution cascade** (`:318`) — pass the existing `'precise'` strategy to `getSymbolAtPosition` (as Hover/Definition already do) and delete the hand-rolled identifierRange guard. Do not tighten the shared Step-4 fallback — Implementation relies on it.

4. **Leaf retry can mis-resolve to an unrelated same-named symbol** (`:351`) — constrain the dotted-name last-segment retry to members of the resolved qualifier, or drop it when the qualifier doesn't resolve.

Items 1–4 are quality/altitude, not crashes — happy to discuss any of them.

peternhale (commented):
Thanks Kyle. The `pickNameUnderCursor` substring/leaf-bias logic and the `isPositionWithinIdentifier` null-guard both check out, and `isPositionInIdentifierRange` is genuinely consumed (via the `posInIdRange` alias in `ApexSymbolManager.ts`), so no dead export. Two concerns to investigate before merge:

> Reminder: base branch is `feature/W-22692420-output-shape-fixes` (stacked on #463) — merge #463 first.

### 1. [Medium] PR description overclaims vs. what's on the branch
The body lists three behaviors that aren't in the diff:
- "Overload selection by signature (`disambiguateOverload` / `methodSignatureKey`)"
- "last-segment retry for dotted names that still fail"

A `git grep` against `origin/feature/W-22692425-chained-typeref-resolution` finds **none** of `disambiguateOverload`, `methodSignatureKey`, or a third retry stage. The diff has exactly two resolution stages (`getSymbolAtPosition('precise')` → `resolveSymbol`), and the body's own "Known limitation" section admits per-overload separation isn't implemented. The overloaded-method fixture (`process(Integer)` / `process(String)`) is added but exercises behavior the code doesn't have.

**Could you confirm** whether a commit was dropped during the rebase onto #463's branch, or whether the description should be trimmed to match? Want to make sure we're not losing intended work.

### 2. [Low] `precise` can resolve a single-segment name to a *different* declaration than the old name-lookup, with no fallback
The new resolution order prefers `getSymbolAtPosition(uri, pos, 'precise')` and only falls back to `resolveSymbol(symbolName, context)` when the result is **null** — the `if (!resolvedSymbol)` guard doesn't trigger on a non-null-but-different result.

For the common single-segment case these usually agree, but where they diverge (e.g. an overloaded `process` call, or an inherited-vs-override member) `precise` now picks the position-resolved declaration rather than the name-keyed one, potentially yielding a different reference-location set than pre-PR. Small blast radius given the data-layer limitation you noted, but **could you confirm** no existing single-segment reference test regressed?

Happy to pair on either.

kylewalke:
Thanks Pete — both confirmed, details below.

### 1. Description overclaim — no commit was dropped; the code was removed intentionally
`disambiguateOverload` / `methodSignatureKey` and the last-segment retry were deleted in `75291be8c` as part of addressing your #1 and #4. So `git grep` finding none of them on the branch is correct and expected — nothing was lost in the rebase. I've trimmed the description to match the two-stage resolution that's actually on the branch (`getSymbolAtPosition('precise')` → `resolveSymbol`), moved the removed paths into an explicit "Review-driven simplification" section, and reconciled the "Known limitation" wording (it no longer references `disambiguateOverload`). The overload fixture stays as a plain method-call site for the unqualified-name test.

On #4 specifically: the leaf retry turned out to be **unreachable** rather than merely risky — `symbolName` always comes from `pickNameUnderCursor`, which only ever returns dot-free names, so `symbolName.includes('.')` was never true. I deleted it outright (which also eliminates the mis-resolution hazard you flagged) rather than constraining a branch that can't run. Happy to revisit with a member-scoped retry if there's a real cursor scenario that needs it.

### 2. `precise` vs. name-lookup divergence — confirmed no regression, and it's bounded at the data layer
You're right that the order now prefers `precise` and only falls back on a **null** result, so a non-null-but-different result wins. But the choice of declaration *instance* can't change the returned location set, because `findReferencesTo` is keyed on `(name, fileUri)`, not on the symbol instance / signature / overload:
- `ApexSymbolManager.findReferencesTo` cache key — `refs_to_${symbol.name}` (`ApexSymbolManager.ts:746`)
- target re-resolved by name within the file — `findSymbolInFileByName(normalizedSymbolFileUri, symbol.name)` (`ApexSymbolRefManager.ts:1947`)

So where `precise` and the old name-lookup diverge but land on the same `(name, fileUri)` (overloads, inherited-vs-override member in the same file), the result is identical. The only theoretical divergence is a same-named symbol in a *different* file — and there `precise` is strictly better, since it's anchored to the cursor's actual file while the old name-lookup had no qualifier constraint at all.

Verification: references unit + handler + upserter suites green (35/35), including the plain-unqualified-name test that drives the real `precise` path against a live symbol manager (no mock). Full `apex-parser-ast` and `lsp-compliant-services` suites also green (0 failed) from the pre-commit runs on both commits.

peternhale (approved):

PR #466, opened 2026-06-17
fix(apex-lsp-parser-ast): walk block source to non-block before deferral - W-22692422

kylewalke:
Stops silently dropping cross-file deferrals whose source is a synthetic block symbol (`block_*`) inside a method body. When a deferred reference's source resolved to a block, the post-batch drain couldn't match it back to a real declaration, so the cross-file edge was lost.

> **Stacked on #464 (6.08).** Base branch is `feature/W-22692421-cross-file-edge-population`; merge after #464. Depends on 6.08's deferred-reference drain.

### What changed
- `processSymbolReferenceToGraphEffect` now walks to the enclosing non-block declaration (new `findContainingNonBlockSymbol` in `ops/positionUtils.ts`, parentId-walk with a cycle guard) before calling `enqueueDeferredReference`, so a deferral's source is the containing method/class — findable by the drain — not the synthetic block. Resolved (non-deferred) reference handling is untouched.

### Tests
- `BlockSourceDeferralWalk.test.ts`: dreamhouse-style flow — a cross-file ref inside a test method body, target file added later, drained — asserts the enqueue source is non-block and the incoming edge is populated. Verified load-bearing (reverting the source walk fails the non-block assertion). 2789 passing in the package.

@W-22692422@


kylewalke:
Closing unmerged. Review (see `pr-reviews/PR-466-6.09-review.md`) established empirically that the cross-file reference edge is already populated correctly by `enqueueDeferredReference`'s internal `findContainingSymbolForBlock` block→declaration walk. The call-site `findContainingNonBlockSymbol` walk added here is redundant for the end-to-end outcome through the public pipeline; the only scenario where it would be load-bearing (source file's `SymbolTable` not yet registered in `fileToSymbolTable` at enqueue time) is not reachable via the supported API. No downstream branch depends on this code (verified: not present in 6.10/6.11/6.12 or main). Closing rather than merging a defensive no-op.

PR #467, opened 2026-06-17
test(apex-lsp-parser-ast): cover clearReferenceStateForFile incoming-edge preservation - W-22692424

kylewalke:
The production fix for this story — `clearReferenceStateForFile` clearing only outgoing edges and preserving incoming reverse-index edges on re-add — already landed on `main` via **#472** (W-23006798, implementor-references), which removed the same `removeIncomingReferencesToSymbols` call as part of making implements/extends first-class reverse edges.

This PR therefore delivers the **documentation clarity + regression coverage** for that behavior, which #472 did not include at the generic reference-graph altitude.

### What changed
- **Comments** in `ApexSymbolRefManager.ts`: document that `removeReferenceKey` cleans both the forward and reverse index per edge, and clarify why incoming edges (target in this file, source elsewhere) must survive a re-add while only outgoing edges are rebuilt. No functional code change (the call removal is already on `main`).
- **`ClearReferenceStateForFile.test.ts`** (new): regression coverage at the generic reference-graph level. #472's `implementorReferences.test.ts` only covers inheritance/implementor edges; these assert the guarantee for arbitrary cross-file references plus two edge cases #472 does not touch.

### Tests
- Generic `B → A.Foo` survives A re-add; A's forward index is actually emptied (proves not a no-op). Verified to fail on the old behavior.
- **P2.1** — renamed target on A re-parse surfaces no phantom `findReferencesTo` result.
- **P2.2** — `removeFile` still purges incoming edges (re-parse-vs-remove asymmetry made explicit).

### Follow-up
- **F10-1 / N1** — stale reverse-index entries are reclaimed only when the *source* file is re-resolved. Pre-existing graph-staleness property, not introduced here; tracked separately.

@W-22692424@


peternhale:
## Review note

#467 itself (comments + `ClearReferenceStateForFile.test.ts`) reviewed **clean** — the added comments accurately describe `clearReferenceStateForFile` / `removeReferenceKey`, the tests call real `ApexSymbolRefManager` methods, and the P2.1 "no phantom under renamed target" assertion is deterministic (symbol IDs incorporate the name, so `Foo` ≠ `Baz`).

The actionable code-review findings are all in **#465**'s `ReferencesProcessingService.ts` (this PR touches a different package). They've been posted there for the author: forcedotcom/apex-language-support#465 (comment).

peternhale (approved):
LGTM. Comments on `clearReferenceStateForFile` / `removeReferenceKey` accurately describe the code, the tests exercise real `ApexSymbolRefManager` methods, and the P2.1 "no phantom under renamed target" assertion is deterministic (symbol IDs incorporate the name, so Foo ≠ Baz). Regression coverage for incoming-edge preservation is a welcome addition.

PR #468, opened 2026-06-17
feat(apex-lsp-extension): cross-worker DataOwnerQuerySymbolByName with worker fallback - W-22692427

kylewalke:
Resolves qualified TypeReferences whose target lives in a file not loaded by the enrichment worker handling the request. When a worker's local name index misses, it can now ask the data-owner (which holds all workspace symbols) by name.

> Independent of the other open epic PRs — branched off `main` (its only dependency, 6.04, is already merged).

### What changed
- **New wire schema `DataOwnerQuerySymbolByName`** (`workerWireSchemas.ts`, exported from `index.ts`, added to `DataOwnerTags` + `DataOwnerRequest`). Payload `{ name, namespace? }`; success `{ matches: [{name, fileUri, kind?}], entries }`; failure `{ _tag, message }` — modeled on `ResolveDepUris` / `QueryGraphData`.
- **Data-owner handler** (both `worker.platform.ts` and `worker.platform.web.ts`, byte-identical): `findSymbolByName` → matches with owning `fileUri` + serialized symbol tables in `entries`.
- **Coordinator dispatcher** case in `WorkerCoordinator.createDispatcher`.
- **Worker fallback**: in `loadSymbolDataForEnrichment`, after the existing `ResolveDepUris` prefetch, `resolveMissingNamesViaDataOwner` checks the local index first and only on a miss routes `dataOwner:QuerySymbolByName` through the assistance proxy, ingesting returned tables (best-effort, injectable fetcher for testability).

### Tests
- `CrossWorkerQuerySymbolByName.integration.node.test.ts`: live node topology, write back a symbol table via `UpdateSymbolSubset`, query by name, assert resolution to owning `fileUri` + table in `entries`; negative case for unknown name. Plus dispatcher switch-coverage cases. shared 433 / apex-ls suites green.

@W-22692427@


peternhale (commented):
Thanks Kyle — the wiring is clean (registration complete across all unions, dispatcher routing correct, both worker handler bodies byte-identical, best-effort error handling sound). Two **medium** concerns I'd like you to investigate before we merge:

### 1. Ingested table is never cross-file-resolved on the hot paths
`resolveMissingNamesViaDataOwner` ingests the owning file's table via `addSymbolTable`, but `loadSymbolDataForEnrichment`'s callers for **Hover / Completion / SignatureHelp / CodeAction** go straight to the provider without calling `resolveCrossFileReferencesForFile` on the requesting URI. The sibling helper `loadDependentsForReferences` *does* call it (worker.platform.ts:891), with a comment explaining that ingesting a table alone "never materializes" the cross-file edges.

So on those paths the table lands in the symbol manager, but the requesting file's unresolved `TypeReference` is never re-materialized — meaning the feature may be a no-op for its stated purpose. The integration test only asserts the data-owner *returns* the table; nothing asserts the reference actually resolves after ingestion.

**Could you confirm** whether a downstream resolve pass picks these up? If not, the helper (or its callers on the hover/completion paths) needs to trigger `resolveCrossFileReferencesForFile`, and it'd be worth an integration test asserting end-to-end resolution rather than just the returned table.

### 2. Sequential blocking cross-worker queries on every enrichment hot path
The helper runs after `ResolveDepUris` inside `loadSymbolDataForEnrichment` (called on every Hover/Completion/Definition/SignatureHelp/CodeLens), issuing one **blocking** IPC round-trip per unresolved name, sequentially (`for … await`).

For a file referencing N namespaced / managed-package types the workspace doesn't own (e.g. `System.Database`), every keystroke fires N sequential blocking round-trips to the single-writer data-owner — each returning empty — adding N×RTT to completion latency. The local-skip check (`findSymbolByName`) doesn't help here since those names never resolve.

**Proposed mitigations to consider:**
- Batch all unresolved names into a single `DataOwnerQuerySymbolByName` query rather than one-per-name.
- Make the query non-blocking / fire-and-forget given the result isn't consumed synchronously today (ties to #1).
- Dedup against names `ResolveDepUris` already attempted so we don't re-query.

Happy to pair on either if useful.

kylewalke:
Thanks Peter — both addressed in the latest push (`2f00faae6`, on top of a `main` merge that also pulls in #488's typecheck fixes). Findings + what changed:

### 1. Ingested table never cross-file-resolved — confirmed gap, now fixed
You're right that ingestion alone doesn't materialize the edge: `addSymbolTable` runs `processSameFileReferencesToGraphEffect` only and explicitly defers cross-file refs (`ApexSymbolManager.ts` ~L2049), so the requesting file's `TypeReference` kept no `resolvedSymbolId` / reverse-index edge.

One nuance worth recording: hover/completion/definition resolve via an **on-demand** `findSymbolByName` at query time (`resolveSymbolReferenceToSymbol` → `chainResolution.ts` L3360), and since we ingest *before* the provider runs (and `addSymbolTable` invalidates the `symbol_name_*` cache), those three paths *did* pick up the symbol. So it wasn't a total no-op — but the **edge** was never materialized, which starves reverse-index consumers and left the feature half-wired.

Fix: after a successful cross-worker ingest, `loadSymbolDataForEnrichment` now calls `resolveCrossFileReferencesForFile(uri)` on the requesting file (gated on `ingested > 0`; the method is re-entrancy-guarded + `addReference` de-dupes, so it's near-free otherwise).

And the test gap you flagged: added an end-to-end integration test that ingests the data-owner's returned table into a fresh symbol manager and asserts the requester's `CrossWorkerTarget` reference actually resolves (`resolvedSymbolId` set) after the resolve pass — not just that the data-owner returned a table.

### 2. Sequential blocking queries on the hot path — fixed
Collapsed the `for … await` per-name loop into a **single** batched round-trip:
- Added an additive `names[]` to the `DataOwnerQuerySymbolByName` wire schema (single `name` preserved; the data-owner resolves `names ?? [name]` and unions owning-file tables into one `entries` map).
- `resolveMissingNamesViaDataOwner` now pre-filters locally-resolved + duplicate names, then sends the residual set in one hop. A file referencing N unowned/managed-package types costs **1** IPC hop per request, not N.
- **Dedup against `ResolveDepUris`** falls out of the same pre-filter: any name `ResolveDepUris` already resolved is in the local index, so it's dropped before any query — the residual is exactly the set `ResolveDepUris` couldn't map.

On fire-and-forget: I kept it blocking for now because #1's fix makes the result load-bearing (we resolve the edge synchronously after ingest). The batching removes the N×RTT cost that motivated going async, so this felt like the lower-risk shape — happy to revisit if you'd still prefer non-blocking.

node/web worker bodies + data-owner handlers kept byte-identical. typecheck/lint clean; unit (15), integration (3), and related worker tests (12) green.

peternhale (approved):
Approved — verified both review concerns are genuinely addressed in `2f00faae`, not just acknowledged.

**Concern #1 (cross-file edge never materialized):** `loadSymbolDataForEnrichment` now calls `resolveCrossFileReferencesForFile(uri)` after a real ingest (gated on `ingested > 0`), and the new integration test asserts the requester's `TypeReference.resolvedSymbolId` is set after ingest + resolve — the edge, not just the symbol. ✓

**Concern #2 (sequential blocking per-name queries):** collapsed into a single batched round-trip via the additive `names[]` wire field (single `name` preserved, no wire break); dedup against `ResolveDepUris` falls out of the local-index pre-filter. Keeping it blocking is a reasonable call given #1 makes the result load-bearing and batching removes the N×RTT cost. ✓

Verified locally: node/web worker bodies byte-identical, unit tests 15/15, integration 3/3 (after a fresh `apex-lsp-shared` build).

Two minor, non-blocking follow-ups for whenever:
1. The e2e integration test exercises the symbol-manager directly rather than through `loadSymbolDataForEnrichment`, so the `if (ingested > 0)` hot-path gate itself isn't covered end-to-end (low risk — trivial gate).
2. The integration test loads the *built* `apex-lsp-shared` in the worker child process, so it has a real build-ordering dependency — CI is fine, just noting it.

Nice work.

PR #472, opened 2026-06-18
feat(apex-lsp-extension): activate worker dispatch for go-to-implementation - W-23006798

peternhale:
### What does this PR do?

Activates the multi-worker LSP topology so go-to-implementation (and the other symbol-reading features) work end-to-end, and fixes the correctness bugs that surfaced once requests actually ran across workers.

**Worker dispatch + data-owner accumulation**
- Move the file-scoped LSP-request features off the coordinator onto the request pool: `completion`, `documentSymbol`, `codeLens`, `signatureHelp`, `codeAction` (each handler loads its symbol subset, runs the service, returns the finished result). `workspaceSymbol` intentionally stays coordinator-only (workspace-wide, not file-scoped).
- Route document-lifecycle (`open`/`change`/`save`/`close`) to the single-writer data-owner, so symbol compilation accumulates in one place. The coordinator still tracks live document text for in-flight edits.
- Rename the worker role `enrichment`/`enrichmentSearch` → `lspRequest` (it is the LSP-request worker, not an enrichment-only one).

**Go-to-implementation**
- Emit graph edges at `implements`/`extends` sites (`INTERFACE_IMPLEMENTATION` / `INHERITANCE`), so `findReferencesTo(interface)` returns implementors and `findReferencesTo(superclass)` returns subclasses — the basis for discovering implementors that live in unopened files.
- Load inbound implementor/subclass tables and resolve the cross-file supertype edges in the worker before searching; thread cancellation through the request.

**Correctness fixes (found running live)**
- Dropped write-back + cold-start read race: globally-unique assist correlation IDs (per-worker collisions were coalesced away) and a deterministic per-URI readiness latch on the data-owner (replaces a spin gate).
- Write-back wire crossing: sanitize the symbol payload with `cloneForWire` (a real class's symbols carry function thunks that `postMessage` structured-clone throws on).
- Stale-version ready signal and version-vs-detail-level write-back skip in the data-owner.
- Cold-open `documentSymbol` empty outline: thread document text to the pool worker (the provider re-compiles from text, not the graph).
- Multiple-implementor collapse: `findReferencesTo` cached under `refs_to_<name>` was never invalidated on add, so an interface queried with only its first implementor loaded stayed pinned to that one; and `clearReferenceStateForFile` wiped inbound edges on re-add. Invalidate the relationship cache on every add; clear only a file's own outbound edges on re-add (inbound edges drop only on true deletion). Applies symmetrically to subclasses.

Regression coverage added at the unit level and through the live worker topology (write-back/readiness, enrichment round-trip, cold-open document symbol, incremental multi-implementor/multi-subclass discovery).

@W-23006798@, @W-22629627@


kylewalke (changes requested):
What it does

  1. Topology activation — routes file-scoped readers (completion, documentSymbol, codeLens, signatureHelp, codeAction) to the request pool, document lifecycle to the data-owner; renames worker role enrichment* →
  lspRequest.
  2. Go-to-implementation — emits INHERITANCE/INTERFACE_IMPLEMENTATION edges at extends/implements sites so findReferencesTo(type) returns subclasses/implementors, including in unopened files.
  3. Correctness fixes found running live — cloneForWire on write-back, unique assist correlation IDs, a per-URI readiness latch replacing a spin gate, version-aware detail-level skip, cold-open documentSymbol
  text threading, relationship-cache invalidation.

  Issues
  
  🔴 Node/web data-owner divergence. worker.platform.web.ts is missing two fixes its Node twin received, despite a comment pointing to Node "for the full rationale":
  - Web's readiness peek and post-wake re-peek short-circuit on st != null; Node keys on the merged documentVersion reaching the latch version (symbolsAreCurrent). Web can hand v1 symbols for v2 content.
  - Web's write-back skip is still enrichedOrder <= currentOrder; Node guards sameOrOlderVersion && enrichedOrder <= currentOrder, so web drops a newer version's symbols at equal/lower detail level.
  
  Port both guards to web, or add a comment stating web intentionally lags. Root cause is ~250 lines of latch logic hand-duplicated across the two files — a shared module would prevent the drift. Non-blocking only
  if web topology is off in production; confirm that.

  🟡 QuerySymbolReadiness looks unused. Schema + both handlers + queryDataOwner case added, but the gate uses AwaitSymbolReadiness and I found no caller. Confirm with knip; remove or add a "reserved for X"
  comment.

  🟡 yieldToEventLoopAsync added to effectUtils.ts with no consumer in this diff. Confirm it's used or drop it.

  🟢 Awaiting the store in the serial WRITE handler. DispatchDocumentOpen/Change now await setDocument before armReadiness — correct for the write-back race, but it serializes lifecycle throughput on the
  data-owner. Fine if setDocument stays memory-fast; worth confirming.

  🟢 Full-scan implementation fallback. findImplementingClasses/collectInterfaceHierarchy call getAllSymbolsForCompletion() and scan all classes per request, now gated requiresWorkspaceLoad. Acceptable as a
  correctness fallback; candidate for the reverse-index treatment used in #464.

  🟢 Verbose [impl] debug logs in ImplementationProcessingService — gated behind logger.debug(() => …) so cost is fine, but trim the play-by-play before merge.

  Strengths
  
  - cloneForWire regression test documents exactly why the trivial-class GREEN test masked the bug (no function-valued thunks) and uses a stdlib-referencing class that carries them.
  - Readiness latch (arm in serial handler, await off-runner, resolve on every terminal write-back branch including rejections) is well-reasoned, and the self-deadlock avoidance is correct on the Node side.
  - clearReferenceStateForFile clearing only outbound edges (inbound dropped only on true deletion) is the right fix for implementor collapse on re-add.
  - Multi-implementor incremental test reproduces the reported sequence and guards the cache-invalidation fix.

  Verdict

One requested change: resolve the Node/web divergence — port the two guards or document the lag. Everything else is non-blocking. Confirm CI is green on the final merge commit (the *.node.test.ts
  suites spawn real worker threads with 120s timeouts).

peternhale:
Addressed the requested change and both 🟡 items.

**🔴 Node/web data-owner divergence — fixed.** Ported both guards into `worker.platform.web.ts`:
- Readiness peek + post-wake re-peek now call `symbolsAreCurrent(uri, version, st != null)` (the same predicate the Node side uses), keying on the merged `documentVersion` reaching the latch version instead of mere presence. Web no longer hands v1 symbols for v2 content.
- Write-back skip now guards `sameOrOlderVersion && enrichedOrder <= currentOrder`, so a newer version's symbols merge even at an equal/lower detail level.

Logic is a verbatim port of the tested Node path. The ~250-line shared-module extraction (the root cause of the drift) is real but out of scope here — filing a follow-up so this PR stays scoped to the fix.

**🟡 QuerySymbolReadiness — removed.** Confirmed unused: the gate uses `awaitSymbolDataReady` → `AwaitSymbolReadiness`; the only `QuerySymbolReadiness` reference was the unreachable `queryDataOwner` case (nothing dispatches that string). Dropped the schema, both handlers, the `queryDataOwner` case, the `DataOwnerTags`/`DataOwnerRequest` entries, and the exports.

**🟡 yieldToEventLoopAsync — removed.** No consumer anywhere; the Effect-valued `yieldToEventLoop` it sat beside is still used and stays.

**🟢 Verbose [impl] logs / full-scan fallback / awaited setDocument** — left as-is. The `[impl]` logs are lazy-gated (`logger.debug(() => …)`, zero cost when off) and useful for a brand-new feature; the full-scan fallback is the acknowledged correctness path pending the #464-style reverse index; `setDocument` is memory-fast.

Verification: `apex-lsp-shared` (433 tests) and `apex-ls` (168 tests, incl. the real-worker `*.node.test.ts` integration suites) green; node/browser/worker tsconfigs compile; lint clean. Note `worker.platform.web.ts` has no direct test harness (which is why the drift went unnoticed) — the port mirrors the Node behavior exactly.

kylewalke (dismissed):

kylewalke (approved):

PR #486, opened 2026-06-19
fix(apex-lsp-extension): align web data-owner readiness with node + drop dead code - W-23006798

kylewalke:
### What does this PR do?

Addresses three findings from the code review of the worker-topology go-to-implementation work (#472). Cut from `feature/W-23006798-implementor-references` so it lands alongside the rest of that effort.

**1. Node/web data-owner readiness divergence (the headline fix)**
The web data-owner (`worker.platform.web.ts`) had diverged from its Node twin on two correctness paths, despite its own comment pointing at Node "for the full rationale":
- Its `AwaitSymbolReadiness` peek and post-wake re-peek short-circuited on a bare "symbol table present" check (`st != null`) instead of Node's `symbolsAreCurrent`, which keys readiness on the **merged** `documentVersion` reaching the armed latch version (and explicitly does *not* trust `latch.settled`, since a latch also settles on a rejected/superseded write-back). → web could hand v1 symbols for v2 content.
- Its write-back skip guarded on detail level alone (`enrichedOrder <= currentOrder`) rather than Node's `sameOrOlderVersion && enrichedOrder <= currentOrder`. → web dropped a newer version's symbols whenever its compile landed at an equal/lower detail level.

Ported `symbolsAreCurrent` into web and applied it to both peeks; added the `sameOrOlderVersion` condition to the write-back guard. Web now matches Node behavior.

**2. Remove dead `QuerySymbolReadiness`**
The presence-only `QuerySymbolReadiness` wire schema, both data-owner handlers, the `queryDataOwner` dispatch case, and the shared index re-exports had no caller — the cold-read gate uses `AwaitSymbolReadiness`. Removed the lot.

**3. Remove orphaned `yieldToEventLoopAsync`**
Added to `effectUtils.ts` in this series with zero consumers. Removed.

### Verification
- `tsc -b` clean across shared / parser-ast / lsp-compliant-services / apex-ls.
- eslint clean on changed files.
- Integration suites pass (19/19): `WorkerConcurrencyInterop` (stale-version readiness), `ColdStartWriteBackGate`, `EnrichmentRoundTrip`, `WorkerCoordinator.queryDataOwner`.
- Full workspace lint+test green via pre-commit hook.

> Note: the underlying smell is that ~250 lines of latch/readiness logic are hand-duplicated across `worker.platform.ts` and `worker.platform.web.ts` — this PR re-syncs them but doesn't extract a shared module. Worth a follow-up.

@W-23006798@


kylewalke:
Superseded by 28d650a4e on the base branch (feature/W-23006798-implementor-references), which lands the same fix independently — porting `symbolsAreCurrent` and the `sameOrOlderVersion` write-back guard into the web data-owner, and removing the dead `QuerySymbolReadiness` schema/handlers and orphaned `yieldToEventLoopAsync`. Verified the end-state code is identical (diff is comments-only), so closing rather than resolving the conflicts.

peternhale:
Heads-up: this overlaps a fix already on the #472 branch. Before #486 was opened, I'd pushed `28d650a4` to `feature/W-23006798-implementor-references` resolving the same three findings — the web `symbolsAreCurrent` port, the `sameOrOlderVersion` write-back guard, the `QuerySymbolReadiness` removal, and the `yieldToEventLoopAsync` removal. The two implementations are functionally identical (same six files, same code); #486 differed only in comment wording.

I've adopted #486's comment wording onto the #472 branch (`bd20d175`, a comment-only `style` commit), so your prose is preserved. Your separate `0c96acd0` cloneForWire write-back fix was already on the branch and is untouched — I rebased on top of it.

So #486's changes are now fully on the #472 branch. Suggest closing #486 as superseded unless you'd rather land it the other way — your call.

PR #496, opened 2026-06-25
fix(apex-lsp-parser-ast): evict renamed/deleted declarations on re-parse - W-23133526

kylewalke:
## What does this PR do?

`findReferences` reverse-index staleness fix for the re-parse path. On an
authoritative re-parse (newer `documentVersion` → REPLACE), `ApexSymbolManager.addSymbolTable`
now diffs the file's previous declarations against the fresh symbol table and
evicts the gone ones from the symbol indexes **and** the incoming reverse-index
edges that targeted them (new `ApexSymbolRefManager.evictStaleFileDeclarations`).

### Why

`registerSymbolTable` swaps the file's `SymbolTable`, but the graph indexes
(`fileIndex`/`nameIndex`/`symbolIdIndex`/`fqnIndex`/`symbolQualifiedIndex`) and
the reverse index are add-only. A declaration renamed away (`Foo` → `Baz`) or
deleted therefore lingered: its stale id and any incoming reverse-index edge
keyed on it stayed resolvable, so `findReferencesTo(Foo)` kept surfacing a
phantom reference — and the dead reverse-index bucket kept accumulating — until
the *referencing* file was itself re-resolved.

### Scope — F10-1 vs N1 (explicit)

- **N1 (re-parse merge-not-replace eviction) — FIXED.** Eviction runs **only** on
  the `accepted-replace` registration decision. The MERGE path (enrichment passes:
  same document version, higher detail level) is **deliberately skipped** so old
  private/protected symbols that are absent from a partial fresh table are still
  preserved (the existing W-22692424 contract). After renaming `Foo`→`Baz` in
  file A and re-parsing, A's index no longer contains `Foo` and B's stale incoming
  edge into `Foo` is reclaimed, so `findReferencesTo(Foo)` returns nothing.
- **F10-1 (unbounded reverse-index accumulation / compaction) — ADDRESSED for the
  rename/delete case; no general compaction pass added.** Reclaiming the dead
  reverse-index bucket at the moment the target declaration disappears caps the
  accumulation for renames/deletes (the scenario F10-1 calls out). A separate
  periodic/full compaction pass is **not** introduced and was judged unnecessary
  given this eviction: the only known source of dead entries is renamed/deleted
  targets, which are now reclaimed at re-parse time. **Residual gap deliberately
  left:** dead entries originating from a *referencing* file that is never
  re-parsed AND whose target is never re-parsed are still only reclaimed when one
  of the two files is re-resolved; if a workload is found that accumulates these
  in practice, a follow-up compaction sweep can be added.

### Tests

Added to `ClearReferenceStateForFile.test.ts` (3 new tests; verified failing
without the fix by temporarily disabling the eviction branch):
- `findReferencesTo(Foo)` returns nothing and the reverse-index bucket for Foo's
  old id is reclaimed after Foo is renamed away and A is re-parsed (newer version).
- Direct reverse-index reclamation assertion (F10-1): the dead bucket count drops
  rather than accumulating across a rename.
- Over-reach guard: unrelated declarations and live incoming edges (a sibling type
  that was NOT renamed) are preserved on re-parse.

### Verification

- `npm run typecheck -w packages/apex-parser-ast` — clean.
- `npm run lint` (repo, via commit hook) — clean.
- Targeted: `ClearReferenceStateForFile` (6/6), `implementorReferences`,
  `ApexSymbolManager`, `ApexSymbolRefManager`, `test/references` — all pass.
- Full husky pre-commit suite: 4526 passed, 0 failed.

@W-23133526@

Deferred from 6.10 (W-22692424); see `pr-reviews/FOLLOW-UPS-2026-06-19.md` for the
F10-1 / N1 framing.

### DIFF
--- a/packages/apex-parser-ast/src/symbols/ApexSymbolManager.ts
+++ b/packages/apex-parser-ast/src/symbols/ApexSymbolManager.ts
@@ -2016,6 +2016,30 @@ export class ApexSymbolManager implements ISymbolManager, SymbolProvider {
         );
       }
 
+      // N1/F10-1 (W-23133526): on an authoritative re-parse (REPLACE), evict
+      // declarations the fresh parse no longer carries. registerSymbolTable
+      // swaps the file's SymbolTable but the graph indexes are add-only, so a
+      // renamed/deleted declaration (e.g. Foo -> Baz) would otherwise linger:
+      // its stale id and any incoming reverse-index edge keyed on it stay
+      // resolvable, leaving findReferencesTo(Foo) surfacing a phantom reference
+      // until the referencing file is itself re-resolved. The MERGE path is
+      // deliberately skipped — enrichment passes (same document version, higher
+      // detail) intentionally preserve old private/protected symbols that are
+      // absent from the partial fresh table.
+      if (registration.decision === 'accepted-replace') {
+        const liveSymbolIds = new Set<string>();
+        for (const symbol of symbols) {
+          const liveId = symbol.key?.unifiedId || symbol.id;
+          if (liveId) {
+            liveSymbolIds.add(liveId);
+          }
+        }
+        self.symbolRefManager.evictStaleFileDeclarations(
+          normalizedUri,
+          liveSymbolIds,
+        );
+      }
+
       // Invalidate cache for all symbol names that were added
       // This ensures that findSymbolByName() will see the newly added symbols
       for (const symbolName of symbolNamesAdded) {
--- a/packages/apex-parser-ast/src/symbols/ApexSymbolRefManager.ts
+++ b/packages/apex-parser-ast/src/symbols/ApexSymbolRefManager.ts
@@ -3372,6 +3372,104 @@ export class ApexSymbolRefManager {
   }
 
   /**
+   * Evict declarations that no longer exist in a file after a re-parse.
+   *
+   * `removeFile` purges an entire file; this is the surgical analogue used on
+   * the re-add path: given the symbol ids that the fresh parse still declares
+   * for `fileUri`, it drops every other id that the file previously declared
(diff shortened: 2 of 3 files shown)


peternhale (approved):
Approve — correct, well-scoped, well-tested staleness fix. Scoping notes (N1 fixed, F10-1 partially addressed) are accurate and honest about the residual gap.

Correctness verified:
- **ID matching is sound** (the key risk): `fileIndex` stores `getSymbolId()` (= `symbol.id`); eviction builds `liveSymbolIds` from `symbol.key?.unifiedId || symbol.id`. These agree because `SymbolTable.addSymbol` synchronizes `symbol.id = symbol.key.unifiedId` (symbol.ts:937-938). No risk of evicting live symbols or missing stale ones.
- **Mirrors `removeFile`** exactly across all indexes (`symbolFileMap`/`symbolIdIndex`/`symbolQualifiedIndex`/`fqnIndex`/`nameIndex`/`fileIndex`) plus incoming reverse edges; adds a small write-skip guard that's functionally equivalent.
- **No cross-file over-reach**: reverse-index keys are qualified by `(targetFileUri, targetSymbolId)` and name/FQN buckets filter by globally-unique `symbolId`, so identically-named symbols in other files are untouched.
- **Scope gating correct**: fires only on `accepted-replace` (newer version + complete parse); MERGE/enrichment path correctly excluded, preserving the W-22692424 contract.

Verified locally: typecheck clean; ESLint clean on all 3 changed files; ClearReferenceStateForFile 6/6, ApexSymbolManager/ApexSymbolRefManager 324 pass, implementorReferences/test/references 65 pass.

Non-blocking: the alternate `symbolIndexStoreLive.ts` store also returns `accepted-replace` but doesn't invoke eviction — different Ref-based abstraction, consistent with scope, but worth a tracking note if it ever becomes the live path.

PR #497, opened 2026-06-25
feat(apex-lsp-parser-ast): findReferences signature-keyed + scope/namespace resolution follow-ups - W-23133640

kylewalke:
### What does this PR do?

Follow-up work from 6.11 (PR #465) / 6.12 (PR #468) addressing the family of bugs rooted in **name-keyed** find-references/go-to-implementation resolution that is neither signature- nor scope-aware. Lands the self-contained, provably-correct items in full and the foundational schema slice of the hard signature-keying core, with the large/risky remainder clearly deferred.

> **Note on basing:** this work was originally intended to stack on `feature/W-22692427-cross-worker-query` (PR #468, story 6.12). #468 has since **squash-merged into `main`** (commit `4e8e6fbbd`) and its branch was deleted, so this PR is now based on `main` and rebased onto current `main` — it carries only its own three commits and builds on the already-merged 6.12 cross-worker batching work.

#### Sub-item status

| Item | Status | Notes |
|------|--------|-------|
| **F11-4** — go-to-implementation scope-fallback guard | **DONE** | `ImplementationProcessingService` used an unguarded `getSymbolAtPosition('scope')` result, so the cursor on whitespace/keyword inside an abstract/interface method body returned implementations for the **enclosing** symbol. Now mirrors the 6.11 find-references guard: only honor the scope fallback when the cursor lands on the symbol's own declaration identifier (`identifierRange`), not merely inside its scope body (`symbolRange`). Positive + negative regression tests added. |
| **F12-2** — namespace through cross-worker query | **DONE** | `resolveMissingNamesViaDataOwner` sent only `{ names }`. Threaded an optional `namespace`/qualifier hint from the call site through the helper into the `DataOwnerQuerySymbolByName` query payload for same-name disambiguation (the schema + `WorkerCoordinator` already accepted it). The `namespace` key is added only when supplied, so unqualified queries keep the exact prior wire shape. Applied identically to `worker.platform.ts` and `worker.platform.web.ts` — the enrichment/data-owner bodies remain **byte-identical** (verified). Two tests (present + absent namespace). |
| **F11-1** — generic-overload collapse | **DONE (verified + locked)** | Signature keying must use each param's `originalTypeString`, not erased `type.name`, so `f(List<String>)` and `f(List<Integer>)` are distinct. The canonical comparator `areMethodSignaturesIdentical` already keys on `originalTypeString`; added generic / array(varargs-style) / arity tests to lock that behavior so the future signature-keyed `findReferencesTo` builds on a proven base. |
| **F11-2** — per-overload reference separation | **PARTIAL (schema foundation)** | `SymbolReference` carried **no** call-site arg types, so signature-aware resolution had nothing to key on. Added an optional positional `argumentTypes` string field to `SymbolReference` / `EnhancedSymbolReference`, included it in `toJSON` (round-trips across the worker boundary as a plain field), and threaded it as an optional param through `createMethodCallReference`. Holds each arg's `originalTypeString` (so `List<Integer>` is not collapsed to `List`). Additive and non-breaking: absent `argumentTypes` ⇒ today's name-only keying. **Deferred:** parser *population* of `argumentTypes`, the signature-aware reverse-index re-keying in `ApexSymbolRefManager` (`addReference`/`findReferencesViaGraph` currently collapse overloads at insert time via name-only `findSymbolInFileByName`), and the overload-selecting resolver. These are the large/risky core changes; landing them here would produce an unreviewable diff and risk regressing the cross-file graph. |
| **F11-3** — qualifier-scoped last-segment retry | **DEFERRED** | A clean fix needs a new **public** `ISymbolManager` qualifier-scoped-member resolution method (resolve the qualifier, then look up the leaf as its member). `resolveMemberInContext` exists only as an internal op against the manager, not on the public interface `ReferencesProcessingService` consumes. Wiring it cleanly is itself meaningful core work that interacts with the same deferred resolution machinery as F11-2. Deferred rather than land a rushed partial that risks regressing leaf resolution. |

#### Residual work for a future PR
- F11-2 core: parser emission to populate `argumentTypes` from call-site argument expressions; signature-aware reverse-index keying in `ApexSymbolRefManager` (overload-distinct target resolution at `addReference` time + at `findReferencesViaGraph`); overload-selecting resolver; signature-keyed `findReferencesTo` cache key in `ApexSymbolManager` (currently `refs_to_${name}`).
- F11-3: public qualifier-scoped member-resolution API + `ReferencesProcessingService` retry that resolves the qualifier first.
- F12-2: optional data-owner-side consumption of the `namespace` hint to actually filter same-named matches (currently threaded but not yet consumed).

#### Verification
- `npm run typecheck` clean for all touched files in `apex-parser-ast`, `apex-ls`, and `lsp-compliant-services`. Pre-existing, unrelated typecheck failures remain in `lsp-compliant-services/test/queue/LSPQueueManager.test.ts` and `test/settings/ApexLanguageServerSettings.test.ts` (confirmed identical with these changes stashed — mock-shape drift from earlier 6.x work, not from this PR).
- `npm run lint` clean for touched files.
- Targeted suites pass: `ReferencesProcessingService`, `ImplementationProcessingService`, `resolveMissingNamesViaDataOwner`, `WorkerCoordinator.queryDataOwner`, `CrossWorkerQuerySymbolByName`, `implementorReferences`, `ApexSymbolRefManager`, `signatureKeying` (new), `MethodSignatureEquivalence`, `DuplicateMethodValidator`.
- Full husky pre-commit/pre-push suite ran green on every commit (no `--no-verify`).

#### Tests added
- `ImplementationProcessingService.test.ts`: F11-4 positive (cursor on declaration identifier ⇒ implementations) + negative (cursor in body ⇒ `[]`, no discovery).
- `resolveMissingNamesViaDataOwner.test.ts`: F12-2 namespace threaded into the batched query; namespace key omitted entirely when not supplied.
- `references/signatureKeying.test.ts` (new): F11-1 generic/array/arity distinctness on `areMethodSignaturesIdentical`; F11-2 `argumentTypes` flows through the factory and survives JSON serialization.

@W-23133640@

### DIFF
--- a/packages/apex-ls/src/worker.platform.ts
+++ b/packages/apex-ls/src/worker.platform.ts
@@ -868,6 +868,12 @@ async function loadSymbolDataForEnrichment(
  * @param queryByName Coordinator-assistance fetcher; injectable so the
  *   ingestion contract can be unit-tested without a live assistance bus.
  *   Defaults to {@link requestCoordinatorAssistancePromise}.
+ * @param namespace Optional namespace/qualifier hint (e.g. the leading
+ *   qualifier of a qualified TypeReference such as `MyNs` in `MyNs.Foo`).
+ *   Threaded through to the {@link DataOwnerQuerySymbolByName} query so the
+ *   data-owner can disambiguate same-named matches across namespaces. Omitted
+ *   from the wire payload when absent so unqualified queries are byte-identical
+ *   to before.
  * @returns Count of owning files ingested (0 on failure or no matches).
  */
 export async function resolveMissingNamesViaDataOwner(
@@ -878,6 +884,7 @@ export async function resolveMissingNamesViaDataOwner(
     params: unknown,
     blocking: boolean,
   ) => Promise<unknown> = requestCoordinatorAssistancePromise,
+  namespace?: string,
 ): Promise<number> {
   // Drop duplicates and names the LOCAL name index already resolves before any
   // IPC. The local-skip also dedups against ResolveDepUris: any name it already
@@ -901,12 +908,18 @@ export async function resolveMissingNamesViaDataOwner(
     // per keystroke; batching makes it a single hop. The success `entries` map
     // is keyed by owning file URI, so it carries every matched name's table.
     //
-    // NOTE: the schema/coordinator also accept an optional `namespace`, but we
-    // intentionally omit it for now (forward-compat hint — qualifier
-    // disambiguation is not wired yet).
+    // Thread the optional namespace/qualifier hint through to the data-owner
+    // for same-name disambiguation. Only add the key when a namespace is
+    // supplied so unqualified queries keep the exact prior payload shape.
+    const queryParams: { names: string[]; namespace?: string } = {
+      names: residual,
+    };
+    if (namespace) {
+      queryParams.namespace = namespace;
+    }
     const response = (await queryByName(
       'dataOwner:QuerySymbolByName',
-      { names: residual },
+      queryParams,
       true,
     )) as {
       matches?: ReadonlyArray<{ name: string; fileUri: string }>;
--- a/packages/apex-ls/src/worker.platform.web.ts
+++ b/packages/apex-ls/src/worker.platform.web.ts
@@ -800,6 +800,12 @@ async function loadSymbolDataForEnrichment(
  * @param queryByName Coordinator-assistance fetcher; injectable so the
  *   ingestion contract can be unit-tested without a live assistance bus.
  *   Defaults to {@link requestCoordinatorAssistancePromise}.
+ * @param namespace Optional namespace/qualifier hint (e.g. the leading
+ *   qualifier of a qualified TypeReference such as `MyNs` in `MyNs.Foo`).
+ *   Threaded through to the {@link DataOwnerQuerySymbolByName} query so the
+ *   data-owner can disambiguate same-named matches across namespaces. Omitted
+ *   from the wire payload when absent so unqualified queries are byte-identical
+ *   to before.
  * @returns Count of owning files ingested (0 on failure or no matches).
  */
 export async function resolveMissingNamesViaDataOwner(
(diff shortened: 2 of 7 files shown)


peternhale (approved):
Reviewed via /review (medium effort, multi-angle finder + verify pass).

No correctness or regression findings survived verification:
- The `identifierRange` guard in ImplementationProcessingService is safe — the 'scope' strategy only returns Interface and abstract/virtual Method symbols (Block symbols filtered at ApexSymbolManager.ts:2932), and both reliably carry an identifierRange covering the declaration name, so no previously-working cursor-on-name case regresses.
- EnhancedSymbolReference constructor param order matches the padded args in createMethodCallReference; argumentTypes lands in the correct final slot.
- The two worker files' new blocks are byte-identical, and the `namespace` key is correctly omitted (not undefined) when absent, preserving the wire shape.

Non-blocking cleanup note: the long optional-positional EnhancedSymbolReference constructor is becoming fragile (createMethodCallReference now pads 9 commented `undefined` args; the round-trip test pads 15). Worth considering an options-object/builder when the deferred F11-2 core work adds the next field — fine to defer.

Well-tested, deferrals clearly documented, additive changes non-breaking. LGTM.

peternhale:
### Review findings (non-blocking cleanup)

The behavioral changes verified clean — no correctness or regression issues. The two items below are cleanup observations, both rooted in the same growing-positional-constructor smell. Fine to defer; flagging for when the deferred F11-2 core work adds the next field.

1. **`packages/apex-parser-ast/src/types/symbolReference.ts` (~L204-217)** — `createMethodCallReference` threads `argumentTypes` into the final constructor slot by passing 9 explicitly-commented positional `undefined` args. Any future field inserted before `argumentTypes` forces re-counting/editing every padded call site, and the inline `// literalValue`, `// literalType`, … comments are themselves a sign the positional constructor has outgrown its arity. An options-object or builder would make new optional fields zero-touch. (Pre-existing pattern the PR extends rather than introduces.)

2. **`packages/apex-parser-ast/test/references/signatureKeying.test.ts` (L370-391)** — the JSON round-trip test constructs `EnhancedSymbolReference` with 15 positional `undefined` args before the one it actually exercises. The test encodes positional knowledge it doesn't test, so a constructor reorder would break it spuriously. Using the factory or a named-field builder would isolate it from unrelated constructor churn.

Neither blocks merge — the constructor fragility is just a good candidate to address alongside the next additive field.

PR #500, opened 2026-06-25
feat(apex-lsp-shared): promote QueueState/GraphData types to shared - W-23163173

peternhale:
## Summary

- Adds QueueStateParams/Result and GraphDataParams/Result protocol types to apex-lsp-shared for the apex/queueState and apex/graphData custom LSP requests
- Defines shared structural shapes (SchedulerMetricsShape, GraphNodeShape, GraphEdgeShape, etc.) that model the canonical runtime types from apex-parser-ast and lsp-compliant-services without importing them (keeps apex-lsp-shared lightweight)
- Adds compile-time assignability assertions in lsp-compliant-services/test/protocol/sharedTypesAssignable.test.ts to verify local runtime shapes remain compatible with the shared contract

## Plan

See .claude/plans/W-23163173.md

## Review follow-up (committed in 1d809eb2)

The four issues raised in review are now fixed and committed (the earlier "applied in the working tree but uncommittable" note is obsolete — the merge-conflict blocker was resolved by the later merge from main):

1. **Sound assignability gate.** The gate previously materialised each assertion as `null as unknown as <never>`, which always type-checks — so it caught nothing. Replaced with `Expect<IsAssignable<To, From>>`: `IsAssignable` resolves to the literal `true`/`false`, and `Expect<T extends true>` makes a `false` a concrete `tsc` error on that assertion's line.

2. **Corrected `GraphEdgeShape.type`.** Fixing the gate (#1) uncovered a real type bug the broken gate had masked. `GraphEdgeShape.type` was `string | number`, but the runtime `GraphEdge.type` is `EnumValue<typeof ReferenceType>`, which widens to `EnumPrimitive | undefined` (string | number | boolean | symbol | undefined) — not assignable to `string | number`. Changed to `EnumPrimitive | undefined` so the shared shape genuinely mirrors the runtime type. (5 assertions — GraphEdge/GraphData/FileGraphData/TypeGraphData/GraphDataResult — were silently passing before.)

3. **Gate now runs in the pipeline.** Added `typecheck` to `lsp-compliant-services` `wireit.test.dependencies`, so the assertions are enforced on `npm test` (and pre-commit/pre-push), not just on a `typecheck` target nothing invoked.

4. **Unblocked #3** by fixing 3 pre-existing, unrelated typecheck errors that would otherwise turn the suite red: `LSPQueueManager.test.ts` (mock missing `drainAllDeferredReferences`/`findSubtypes`/`findSupertypes` after `ISymbolManager` grew) and `ApexLanguageServerSettings.test.ts:218,309` (partial fixtures missing newer `apex` sub-settings — `loadWorkspace`/`queueProcessing`/`scheduler`).

**Non-vacuousness verified:** with the #2 type fix reverted, the gate fails on exactly those 5 assertions (`Type 'false' does not satisfy the constraint 'true'`); with it in place, typecheck + jest + the full monorepo suite are green (4581 passed, 0 failed).

## Test plan

- [x] Compile apex-lsp-shared; verify no apex-parser-ast import
- [x] Verify module lives at src/queueStateGraphData.ts (root); confirm no src/protocol/ dir was created
- [x] Verify index.ts re-export uses a named export type { … } block matching the wireSchemas/workerWireSchemas style; new symbols resolve from @salesforce/apex-lsp-shared root
- [x] Test lsp-compliant-services; assignability test passes
- [x] Compile lsp-compliant-services; local shapes still build
- [x] Grep confirm apex-lsp-shared/package.json deps unchanged (no parser-ast added)
- [x] Lint clean on touched files
- [x] Assignability gate is sound (non-vacuous) and wired into the test target; full monorepo suite green
- [ ] NOT e2e-covered: pure type-level promotion; no runtime/protocol behavior change → no e2e on branch exercises this. Compile + type-assertion test are the gate.

@W-23163173@

---
🤖 Generated by auto-build pipeline. Original WI: https://gus.my.salesforce.com/a07EE00002d1jRyYAI


kylewalke:
Do not merge as-is.

**1. The assignability gate catches nothing.** `test/protocol/sharedTypesAssignable.test.ts:79-122` — `null as unknown as AssertX` always type-checks even when the assertion resolves to `never` (verified with the repo `tsc`). Replace with a sound, error-producing pattern:
```ts
type IsAssignable<To, From> = [From] extends [To] ? true : false;
type Expect<T extends true> = T;
type _AssertQueueStateParams = Expect<IsAssignable<SharedQueueStateParams, LocalQueueStateParams>>;
```

**2. `GraphEdgeShape.type: string | number` is a wrong model of the runtime type.** `src/queueStateGraphData.ts:184` — runtime `GraphEdge.type` is `EnumValue<typeof ReferenceType>`, which resolves to `EnumPrimitive | undefined` (not `string | number`); not assignable (verified with `tsc`). Declare `type: EnumPrimitive | undefined` (import `EnumPrimitive` from `./enumUtils`).

**3. The gate never runs in any pipeline.** Assertions are only enforced by the `typecheck` target, but `wireit.test` depends only on `compile`, CI never runs `npm run typecheck`, and pre-commit runs only `lint` + `test`. Add `typecheck` to `wireit.test.dependencies` for `lsp-compliant-services` (and/or a CI step). This requires fixing 3 pre-existing unrelated type errors first: `LSPQueueManager.test.ts:116`, `ApexLanguageServerSettings.test.ts:218,309`.

**4. Diff does not match the PR description.** The "Reviewer notes" claim fixes #1 and #2 were applied, but the committed diff contains neither. Land the fixes or correct the description before merge.

kylewalke (changes requested):
^^

kylewalke (approved):

PR #501, opened 2026-06-25
feat(apex-lsp-parser-ast): findReferences overload separation + qualifier-scoped resolution (F11-2 core, F11-3) - W-23133640

kylewalke:
## What does this PR do?

**F11-2 core: per-overload reference separation.** find-references on one overload of an overloaded method used to return references to *all* same-named overloads. Name-keyed resolution collapses every overload onto a single target declaration, and both the cache and the reverse-index lookup keyed on name alone — so overload identity was discarded.

This PR keys reference resolution on **call-site arity**, the discriminator that is statically knowable at parse time (no type resolution needed):

- **`SymbolReference.argumentCount`** — new optional field, stamped on every `METHOD_CALL` reference by the parser. Added as a **plain post-construction field**, not another positional constructor parameter — directly addressing the #497 review note that the long optional-positional `EnhancedSymbolReference` constructor is getting fragile, rather than worsening it.
- Stamped at **all 6 `createMethodCallReference` sites** across *both* the symbol-collection pass and the worker reference pass (`ApexReferenceCollectorListener`). A **listener-drift guard test** locks the worker path so the two listeners can't silently diverge and disable separation live.
- Threaded through the reverse-index `RefStoreEntry.context` (and every deferred/ingest path) so it survives to query time.
- **`findReferencesTo`** now runs a guarded filter (`separateOverloadReferences`) that narrows the union to the requested overload's parameter count — but **only when the target method actually has same-named siblings** on its declaring type. Non-methods and non-overloaded methods pass through untouched, so the common case is byte-for-byte unchanged.
- The `findReferencesTo` **cache key** is now overload- and file-aware (was name-only), so an overloaded method's per-arity results no longer alias each other — or collide with same-named members in other files.

## Scope & limits

- **Catches:** arity-distinct overloads — `f()` vs `f(x)` vs `f(x, y)` — the common real-world case.
- **Out of scope (documented follow-up):** same-arity/different-type overloads (`f(String)` vs `f(Integer)`) cannot be separated by arity alone and remain unified; that needs call-site *type* capture at ingestion.
- References whose `argumentCount` is undefined (pre-existing edges, non-call references) are **never dropped**, preserving today's behavior.

## Testing

- New `overloadSeparation.test.ts`: end-to-end proof that the zero-arg overload no longer picks up the one-arg call site and vice-versa; non-overloaded pass-through; same-arity-stays-unified limit; worker-pass drift guard. Verified non-vacuous (fails with the filter disabled).
- `signatureKeying.test.ts`: `argumentCount` field + JSON round-trip.
- Full `@salesforce/apex-lsp-parser-ast` suite green (2823 passed); typecheck + lint clean; downstream `apex-ls` typecheck clean.

@W-23133640@

---

## F11-3: qualifier-scoped member resolution (added in this PR)

Investigated the "last-segment retry mis-resolve" — `Qualifier.member` resolving to an unrelated same-named member from a different class. **It is no longer reproducible**: the #465 chain-resolution restructure already made `resolveMemberInContext` resolve members strictly within the qualifier type's own scope (plus its superclass / Object / stdlib chain), returning `null` when absent rather than leaking to a global same-named symbol. Confirmed empirically across property, method, class-type, and full-chain (`resolveQualifiedReferenceFromChain`) vectors.

Rather than add a guard for an already-closed path, this PR adds **regression tests** that lock the correct behavior in (`ApexSymbolManager.qualifierScopedResolution.test.ts`), so a future change to the member-resolution fallback cannot silently reintroduce the cross-class mis-resolve.


peternhale (changes requested):
**Request: normalize `fileUri` in `buildReferencesToCacheKey`**

`buildReferencesToCacheKey` keys the cache on the raw `symbol.fileUri`:

```ts
const filePart = symbol.fileUri ?? 'no-file';
```

Everywhere else in this layer, file URIs are run through `extractFilePathFromUri(...)` before they're used as a key — symbol IDs (`ApexSymbolRefManager` `getSymbolId`/`makeInternalKey`), the reverse-index keys, and the `findReferencesViaGraph` lookup the results come from all normalize first. Using the raw URI here means the new file-aware cache key can disagree with the rest of the system on what "the same file" is.

Failure scenario: the same symbol reached via two URI spellings (e.g. `file:///Users/x/Foo.cls` vs `/Users/x/Foo.cls`, or differing percent-encoding/case) produces two distinct cache keys — `refs_to_m@file:///Users/x/Foo.cls:1` vs `refs_to_m@/Users/x/Foo.cls:1`. Results stay correct, but it's a silent cache miss / duplicate entry, and the arity-aware key the PR introduces is precisely the path where we don't want spelling drift.

Suggested change:

```ts
const filePart = symbol.fileUri
  ? extractFilePathFromUri(symbol.fileUri)
  : 'no-file';
```

(`extractFilePathFromUri` is already imported in this file.) This keeps the cache key consistent with how the reverse index and symbol IDs are normalized.

kylewalke:
Addressed in fb6bc05 — `buildReferencesToCacheKey` now runs `symbol.fileUri` through `extractFilePathFromUri` before keying, matching the reverse index and symbol IDs:

```ts
const filePart = symbol.fileUri
  ? extractFilePathFromUri(symbol.fileUri)
  : 'no-file';
```

Thanks for the catch.

peternhale:
**Follow-up (separate from the fileUri fix, which looks good): align the parallel `findReferencesTo` cache key in `referenceOps.ts`.**

This PR upgraded `ApexSymbolManager.findReferencesTo` to the overload/file-aware `buildReferencesToCacheKey` (and you just normalized the fileUri there — thanks). But there's a second, parallel implementation that was not touched:

`packages/apex-parser-ast/src/symbols/ops/referenceOps.ts`
```ts
// line 24
const cacheKey = `refs_to_${symbol.name}`;
// line 40
const cacheKey = `refs_from_${symbol.name}`;
```

These are exported from the package index as `findReferencesToOp` / `findReferencesFromOp` and surfaced through `symbolManagerFacade`, so they're part of the public surface. Their underlying call (`refs.findReferencesTo`) *does* go through the RefManager and gets overload-separated by `separateOverloadReferences` — but the results are then cached under a **name-only** key. So per-overload results alias each other in the facade cache: a consumer querying `f()` then `f(x)` on an overloaded method gets the first call's wrong-arity results back from cache.

This is the exact aliasing `buildReferencesToCacheKey` was introduced to prevent — just in the sibling implementation. To keep the two `findReferencesTo` paths from disagreeing, the cleanest fix is to share one key builder rather than maintain two key formats (e.g. extract the key logic so both `ApexSymbolManager` and `referenceOps` call it). The `findReferencesFrom` twin at line 40 has the same name-only key and should move together.

Not blocking the fileUri fix — flagging so the two implementations don't drift before merge.

peternhale (changes requested):
## Review Summary

This PR implements F11-2 (per-overload reference separation) and F11-3 (qualifier-scoped resolution regression tests). The approach of keying reference resolution on call-site arity is sound, and the implementation correctly stamps `argumentCount` on METHOD_CALL references and threads it through the reverse index and cache keys.

## Findings

### 🔴 Changes Requested

**[referenceStore.ts:29-40](https://github.com/forcedotcom/apex-language-support/blob/feature/W-23133640-signature-keyed-core/packages/apex-parser-ast/src/symbols/services/referenceStore.ts#L29-L40)** — ReferenceStoreShape.addReference missing argumentCount field

The `ReferenceStoreShape` interface defines the `context` parameter without `argumentCount`:

```typescript
context?: {
  methodName?: string;
  parameterIndex?: number;
  isStatic?: boolean;
  namespace?: string;
  // argumentCount is missing here
}
```

But `ApexSymbolRefManager.addReference` (line 1807) expects it, and all three storage interfaces (`ReferenceEdge`, `ReferenceResult`, `RefStoreEntry`) include it.

**Impact:** Effect-wrapped callers constrained by the `ReferenceStoreShape` interface type cannot pass `argumentCount` without a compile error. This silently breaks overload separation (F11-2) for any code using the Effect-based ReferenceStore API instead of calling ApexSymbolRefManager directly.

**Fix:** Add `argumentCount?: number;` to the context type in ReferenceStoreShape.

---

### ⚠️ Performance Discussion

**[ApexSymbolRefManager.ts:2008](https://github.com/forcedotcom/apex-language-support/blob/feature/W-23133640-signature-keyed-core/packages/apex-parser-ast/src/symbols/ApexSymbolRefManager.ts#L2008)** — separateOverloadReferences uses O(workspace) lookup

The overload sibling detection calls `findSymbolByName(symbol.name)`, which returns ALL same-named symbols workspace-wide, then filters by `fileUri` and declaring type:

```typescript
const siblings = this.findSymbolByName(symbol.name).filter(
  (s) =>
    isMethodSymbol(s) &&
    s.fileUri === symbol.fileUri &&
    this.findDeclaringTypeForMember(s)?.id === declaringType.id,
);
```

**Impact:** This scales with workspace size rather than type member count. In a workspace with 10,000 `execute` methods across all classes, every `findReferencesTo` on an overloaded `execute` fetches all 10,000, filters them, and then discards 9,997. The cost grows as the codebase grows, even though we're only looking for 2-3 overloads on a single class.

**Question:** Can this be improved? Options might include:
- Caching the sibling-count check keyed on `(declaringTypeId, methodName)` since declaring type membership is stable between file changes
- Adding a type-scoped member lookup API to avoid the global `findSymbolByName` (though this might require broader SymbolTable changes)
- Pre-computing overload metadata at ingestion time

Not blocking for merge if the caching already mitigates this in practice, but wanted to surface the O(workspace) behavior for visibility.

peternhale (approved):
## ✅ Both Issues Addressed

Great work addressing both review findings!

### 🟢 Issue 1: ReferenceStoreShape.argumentCount - FIXED

**[referenceStore.ts:39](https://github.com/forcedotcom/apex-language-support/blob/d12d8362/packages/apex-parser-ast/src/symbols/services/referenceStore.ts#L39)**

```typescript
context?: {
  methodName?: string;
  parameterIndex?: number;
  isStatic?: boolean;
  namespace?: string;
  argumentCount?: number;  // ✅ Added
}
```

The `argumentCount` field is now properly included in the `ReferenceStoreShape` interface, fixing the type mismatch. Effect-wrapped callers can now pass `argumentCount` through the service layer without compile errors.

---

### 🟢 Issue 2: O(workspace) Lookup - FIXED

**[ApexSymbolRefManager.ts:2008](https://github.com/forcedotcom/apex-language-support/blob/d12d8362/packages/apex-parser-ast/src/symbols/ApexSymbolRefManager.ts#L2008)**

Changed from:
```typescript
const siblings = this.findSymbolByName(symbol.name).filter(...)  // O(workspace)
```

To:
```typescript
const siblings = this.getSymbolsInFile(symbol.fileUri).filter(  // O(file)
  (s) =>
    s.name === symbol.name &&
    isMethodSymbol(s) &&
    this.findDeclaringTypeForMember(s)?.id === declaringType.id,
);
```

**Impact:** This changes the complexity from O(workspace symbols) to O(file symbols). For a workspace with 10,000 `execute` methods across all classes, this now only searches the ~50-200 symbols in the declaring file instead of all 10,000. Much better! 🎉

The comment clearly explains the optimization:
> Source candidates from the declaring symbol's own file rather than a workspace-wide findSymbolByName(name) scan: overload siblings live in the same file as the declaration, so this is bounded by file size instead of by the count of same-named symbols across the whole workspace.

---

## Commit Message

The commit message clearly captures both fixes:
> `fix(apex-lsp-parser-ast): argumentCount in ReferenceStoreShape, file-scoped overload lookup - W-23133640`

Perfect! ✅

PR #502, opened 2026-06-26
feat(apex-lsp-client): new package + RpcConnection port + ApexClientCore skeleton - W-23163181

peternhale:
## Summary

- Scaffold new `@salesforce/apex-lsp-client` package with Effect-based architecture (Effect internal, Promises/plain types exported)
- Define `RpcConnection` port interface (transport abstraction) with Disposable-returning handlers, onError/onClose hooks
- Implement `ApexClientCore` skeleton with Concern 1 lifecycle (`initialize`/`shutdown`/`dispose`), default logging middleware (Effect.log), and default `apex/findMissingArtifact` responder (`{ notFound: true }`)

## Plan

See `.claude/plans/W-23163181.md` for full work-item specification (phases, verification, Effect idiom rules, out-of-scope items).

## Reviewer notes

**Intentional deferred concerns (per plan, NOT bugs):**

1. **Inert middleware chain** (HIGH) — `ApexClientCore.ts` populates `middlewareRef` but the middleware chain is not yet wired into request/notification routing. All LSP traffic currently bypasses interceptors. This is explicitly deferred to Concern-4 by the work-item plan (`.claude/plans/W-23163181.md:68`). Implementing the full `next`-based routing here would pull future-WI scope forward and is a behavior-changing design decision affecting all LSP traffic. The standalone `loggingMiddleware` runtime-loss bug WAS fixed (now provides `EffectLspLoggerLive` before `runFork`) so logs land correctly once the chain is wired.

2. **Test logger console.log** (LOW) — `EffectTestLoggerLive` uses `console.log` (`effectLspLoggerLayer.ts:108`), flagged by `effect-best-practices`. NOT applied: this is an internal test-only logger (not exported from `index.ts`) whose explicit purpose is console fallback when the LSP wire is unavailable. Removing it would require leaking Effect types into test helpers or building alternate infra — disproportionate to zero user-facing impact. The production logger (`EffectLspLoggerLive`) routes to the LSP wire correctly.

3. **Effect.gen vs Effect.fn style preference** (LOW) — `initializeEffect`/`shutdownEffect` use `Effect.gen`. The finding noted the current form is defensible for compute-once cached primitives. The lifecycle effects were rewritten anyway (cache-of-failure fix) into mutex-guarded forms; the zero-param `Effect.fn` wrapper noted as awkward was not adopted. No functional impact.

## Test plan

All verification is unit/build-level (Phase 5 tests). This WI is intentionally NOT e2e-covered — the SDK has no runtime consumer until adapters/headless host land (WI 2.3+) and conformance/integration suites (WI 5.1). The lifecycle ordering invariants asserted here (initialize→initialized, dispose LIFO) are the unit-level proxy for what 5.1 will exercise end-to-end against a real server.

- [x] `npm install` updates lockfile with new workspace package; no errors
- [x] `npm run compile -w @salesforce/apex-lsp-client` — clean tsc build (composite refs resolve to shared `out`)
- [x] `npm run typecheck -w @salesforce/apex-lsp-client` — test tsconfig clean
- [x] `npm run lint -w @salesforce/apex-lsp-client` — BSD headers present, no unused
- [x] `npm run test -w @salesforce/apex-lsp-client` — Phase-5 unit tests green:
  - Fresh Core registers default logging middleware + `{ notFound: true }` responder at construction BEFORE `initialize`
  - `initialize` ordering: request sent and `InitializeResult` resolves BEFORE `initialized` notification (LSP spec)
  - Idempotent `initialize`: second call returns memoized result, does NOT re-send
  - `shutdown()` sends `shutdown` then `exit` in order; idempotent
  - Dispose ordering: middleware/handler Disposables disposed BEFORE `connection.dispose()` (LIFO via finalizers)
  - `use(mw)` returns Disposable; disposing removes it from registered set
- [x] `npm run bundle -w @salesforce/apex-lsp-client` — esbuild cjs+esm emit (smoke)
- [x] Type-boundary check: no exported signature in `out/index.d.ts` references an Effect type (`Effect`, `Ref`, `Scope`, `Layer`)
- [x] Core imports neither vscode nor a transport entry (only port/shared/effect allowed; logging middleware does NOT import `vscode-jsonrpc` Logger)
- [x] Effect-idiom checks:
  - No mutable boolean lifecycle flags, no hand-rolled disposable arrays in `ApexClientCore` (state in `Ref`/`Scope`)
  - `npx effect-language-service diagnostics --project tsconfig.json` — clean
- [x] Orphan-state assertion: `grep -rn "@salesforce/apex-lsp-client" packages/*/src packages/*/package.json | grep -v "packages/apex-lsp-client/"` → zero hits (no consumer yet; adapters land in WI 2.3/4.1)
- [x] No-regression for existing port consumers: `grep -n "RpcConnection" packages/apex-lsp-vscode-extension/src/language-server.ts` → zero hits; `npm run compile -w @salesforce/apex-lsp-vscode-extension` clean (`ClientInterface` unchanged)

@W-23163181@

---
🤖 Generated by auto-build pipeline. Original WI: [W-23163181](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002d13gfYAA/view)


kylewalke:
**1. `forkLog` does not catch the forked fiber's cause.** `src/middleware/loggingMiddleware.ts:1318-1324` — `Effect.runFork(...)` runs with no failure handling, unlike the core's handler at `src/apexClientCore.ts:735-739` which adds `.pipe(Effect.catchAllCause(() => Effect.void))`. A fire-and-forget log that throws leaves an uncaught cause on a detached fiber. Add the same `catchAllCause` for parity.

**2. README contradicts `use()` behavior.** `src/apexClientCore.ts:858-870` — `use()` throws `ApexClientDisposedError` synchronously, but the README says lifecycle methods after `dispose()` reject. Document that `use()` throws (not rejects) so consumers wrap it in try/catch.

kylewalke (approved):
only a couple nits

PR #503, opened 2026-06-26
fix(apex-lsp-extension): cross-file find-references through the worker pool - W-22692429

kylewalke:
## What does this PR do?

6.13 verification capstone for `textDocument/references`. The end-to-end verification work surfaced — and this PR fixes — a real defect: **cross-file find-references dispatched to the enrichment worker pool returned `[]`**. Five compounding root causes, all fixed:

1. **Content not threaded** — the `references` dispatch never carried document text to the stateless pool worker (unlike documentSymbol/hover/completion), so the worker's storage missed and `ReferencesProcessingService` hard-returned `[]`. Added `content` to the `DispatchReferences` wire schema + coordinator + both worker platforms.
2. **Cursor file at public-api** — the pool held the cursor file with method bodies stripped, so an in-body usage resolved to nothing. Recompile the cursor file at full detail from its text (`recompileCursorFileAtFullDetail`), as documentSymbol already does.
3. **Wrong dependents loaded** — the handler loaded dependents of the *cursor* file, but find-references needs callers of the *target symbol's declaring file*. Resolve the cursor symbol's declaring file and load its dependents (`declaringFileForCursorSymbol`).
4. **URI-scheme mismatch** — `ResolveDepUris` keyed wire entries by the schemeless URI from `findFilesForSymbol` (`/test/X.cls`), so ingestion never matched the references' `file:///test/X.cls` targets and cross-file edges failed to bind. Key entries by the table's canonical `getFileUri()`.
5. **Resolved types skipped** — the Phase-2 prefetch skipped types whose references were already resolved (fine for hover/definition, fatal for find-references, which needs the target type's *table* present to enumerate its references). Load referenced types regardless of resolution state (`loadReferencedTypesForFile`).

## Verification (6.13 deliverables)

- **`ReferencesThroughWorkerTopology.node.test.ts`** — asserts cross-file location counts through the live worker topology (the named deliverable).
- **`referenceEnrichmentRecipe.node.test.ts`** — verifies the handler recipe end-to-end at the service level with full observability, and proves the declaring-file step is load-bearing.
- **`apex-find-references.spec.ts`** — Playwright E2E (intra-file, cross-file, no-results, responsiveness) + `findReferences()`/`closePeek()` on `ApexEditorPage`.
- **Regression**: locals-only resolution test; cancellation already covered by `LSPQueueManager.test.ts`.

Full monorepo suite green via pre-commit hook (4555 passed, 0 failed).

### Deferred (out of scope for this PR)
- **Telemetry on the references path** — blocked on Jorje-parity #04 (telemetry plumbing, W-22629622, not started); tracked in the W-22692429 story.
- **Parity-table doc flip** — the design-doc parity table is an external (Quip/GUS) artifact, not an in-repo markdown table; flip handled there.

## Follow-on fallback cleanup (merged from #515)

Additional find-references fixes landed on this branch via #515, applying the standard "fix the cause, don't add a fallback." A companion analysis under `docs/design/` records the full fallback sweep (FB-1..FB-6) and what was deferred.

- **Constructor overloads now separate.** Root went deeper than a cache-key tweak: constructor-call references never carried call-site arity. Stamped `argumentCount` (`countConstructorArguments`) on `CONSTRUCTOR_CALL` refs in both listeners, added `isMethodOrConstructorSymbol`, and keyed the `findReferencesTo` cache + overload separation on it. Previously `Foo()` and `Foo(String)` aliased onto one result set.
- **Qualified cross-file references now bind (FB-1 root).** `Outer.Inner` was stored under its dotted name but bound only via leaf-keyed `findSymbolByName`, so it never bound and `findReferencesTo(Inner)` missed the qualified caller. Resolve dotted names through the FQN index and bind to the leaf symbol; `resolvedSymbolId` is now set so precise resolution works without a by-name guess. By-name fallbacks kept (documented) for the genuine partial-graph case; node/web `declaringFileForCursorSymbol` drift closed by porting the node fallback to web.
- **Duplicate reference locations removed.** `getReferenceLocationsEffect` now collapses exact `(uri, range)` duplicates across declaration / references-to / references-from / relationship edges.
- **Degraded results now attributable.** Best-effort enrichment helpers that silently swallowed failures (subset load logged nothing; recompile / referenced-type / dependent fetch at `debug`) now `warn`, and the no-document-text case is flagged. Mirrored node + web.

Deferred with rationale (see `docs/design/findreferences-pr503-deep-analysis.md` §7): FB-3 (data-owner inner-class index population; the qualified-binding root behind it is fixed, escalation kept as documented safety net), FB-5 (not on the references path — uses `'precise'`), FB-6 (canonical symbol-id rewrite — ~70 id sites + worker wire format, its own work item).

@W-22692429@

### DIFF
--- a/docs/design/findreferences-poc-vs-pr503-findings.md
+++ b/docs/design/findreferences-poc-vs-pr503-findings.md
@@ -1,0 +1,86 @@
+# Find References: POC branch vs. PR #503 — Findings
+
+**Date:** 2026-06-26
+**Author:** Kyle Walker (drafted with Claude assistance)
+
+Comparison of two efforts on `textDocument/references`:
+
+- **POC branch** — `feature/find-references-algorithm-POC` (off an older `main`); design docs `docs/design/find-references-algorithm-completeness.md` and `docs/design/references-offload-to-enrichment-pool.md`.
+- **PR #503** — `feature/W-22692429-references-verification` (@W-22692429@), title *"fix(apex-lsp-extension): cross-file find-references through the worker pool"*.
+
+## TL;DR
+
+| | POC branch | PR #503 |
+|---|---|---|
+| Goal | **Discover & prototype** both (a) the algorithm-completeness fixes and (b) the enrichment-pool offload pipeline, end-to-end | **Verify** cross-file find-references through the worker pool, fixing the remaining *dispatch-layer* defects |
+| Touches `ReferencesProcessingService` | Yes — heavy rewrite (+151/−126) | **No** — unchanged from `main` |
+| Core data-layer algorithm fixes | Prototyped here first | **Already landed on `main`** (refined) — PR #503 builds on them |
+| Primary mechanism | `ReferencesProcessingService` chain-walk + position resolution + URI-field fixes; `drainAllDeferredReferencesSync`; `clearReferenceStateForFile` incoming-edge preservation; block-walk before deferral | 5 worker-pool dispatch fixes in `worker.platform.ts`/`.web.ts` + wire schema `content` field + canonical-URI keying |
+| Status | Manually verified vs `dreamhouse-lwc`; pre-commit hook bypassed; aspirational/experimental | Full suite green (4555 passed); adds integration + recipe + Playwright tests |
+
+The headline: **the POC's algorithm-completeness work has since been absorbed into `main`** (via the W-23133640 signature-keyed and W-23133526 eviction lines of work). PR #503 is the *offload-pipeline* capstone — it assumes the corrected data layer and fixes the five things that still broke when the request actually ran on a pool worker.
+
+## What the POC established
+
+The POC unmasked and fixed a *pre-existing algorithm-completeness bug*: Find All References returned only the declaration site, never the use sites. Root cause was missing cross-file edges in `ApexSymbolRefManager.reverseIndex`. POC fixes (see `find-references-algorithm-completeness.md`):
+
+1. **Cross-file edges populated on the data-owner** — `UpdateSymbolSubset` calls `resolveCrossFileReferencesForFile`; `WorkspaceBatchCompile` drains deferred refs post-batch.
+2. **Synchronous deferred drain** — `drainAllDeferredReferencesSync` bypasses the priority scheduler (the 5 tasks/sec rate limit caused 30s+ stalls).
+3. **Block sources walked to non-block before deferral** — `findContainingNonBlockSymbol`; otherwise `enqueueDeferredReference` silently dropped 118/146 deferrals on dreamhouse Test files.
+4. **`clearReferenceStateForFile` stops wiping incoming refs** — re-adding `Foo.cls` was deleting `Bar.cls → Foo.Member` edges on every enrichment write-back.
+5. **Position-based resolution for chained TypeReferences** — chain-node walk + `getSymbolAtPosition` fallback, in `ReferencesProcessingService.findReferences`.
+6. **URI field-name fix** — read `.fileUri` not `.filePath` in result extraction.
+
+The POC also built the *offload pipeline* (`references-offload-to-enrichment-pool.md`, M1–M11): new wire schemas (`CoordinatorEnsureWorkspaceLoaded`, `DataOwnerResolveDependentUris`), `IWorkspaceLoadCoordinator` injection (Local/Remote impls), `loadDependentsForReferences`, the beefed-up `DispatchReferences` handler, and the `references: 'enrichmentPool'` routing flip.
+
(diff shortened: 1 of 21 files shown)


peternhale (changes requested):
## Review Summary

This PR implements cross-file find-references through the worker pool, fixing 5 root causes that previously caused empty results: content not threaded to workers, cursor file detail level, wrong dependents loaded, URI-scheme mismatches, and resolved types being skipped. The implementation is comprehensive, with integration tests, recipe tests, and Playwright E2E coverage. It also incorporates constructor overload separation and qualified reference binding improvements.

## Findings

I've identified **3 issues** that should be addressed:

### 🔴 Changes Requested

**[worker.platform.ts:1083](https://github.com/forcedotcom/apex-language-support/blob/feature/W-22692429-references-verification/packages/apex-ls/src/worker.platform.ts#L1083)** — Falsy check rejects empty string as invalid content

```typescript
): Promise<boolean> {
  if (!content) return false;  // ❌ Treats '' as falsy
```

The guard `if (!content)` treats empty string (`''`) as invalid, but empty string is a valid zero-length file. The upstream caller at line 1522 explicitly uses `typeof req.content === 'string'` to distinguish "content present" (any string, including `''`) from "content absent" (`undefined`).

**Impact:** When a user opens an empty `.cls` file, the upstream check passes but this guard rejects it. The cursor file stays at public-api detail, in-body cursors fail to resolve, and find-references returns `[]` silently.

**Fix:** Change to `if (content == null)` or `if (content === undefined)` to only reject truly absent content, not empty strings.

**Note:** Same bug exists in web platform at [worker.platform.web.ts:997](https://github.com/forcedotcom/apex-language-support/blob/feature/W-22692429-references-verification/packages/apex-ls/src/worker.platform.web.ts#L997).

---

**[worker.platform.ts:1536](https://github.com/forcedotcom/apex-language-support/blob/feature/W-22692429-references-verification/packages/apex-ls/src/worker.platform.ts#L1536)** — Warning fires but execution continues unchecked

When `req.content` is undefined:
1. Line 1524-1529: Warning is logged about missing document text
2. Line 1536: `recompileCursorFileAtFullDetail` is called anyway
3. Line 1083: Function returns `false` (content guard)
4. **No code checks the return value**
5. Steps 3-6 proceed with degraded state
6. Find-references returns `[]` indistinguishable from "genuinely no matches"

**Impact:** The warning gives a false sense that the failure is handled, but execution continues on broken state. Users see "no references found" with no indication the result is degraded vs genuinely empty.

**Fix:** Either:
- Abort early and return an error result when `!cursorTextAvailable`
- Attempt to fetch content from the data-owner as a fallback
- Check the return value of `recompileCursorFileAtFullDetail` and abort if false

The current warn-then-continue pattern masks silent degradation rather than preventing it.

---

**[ReferencesProcessingService.ts:537](https://github.com/forcedotcom/apex-language-support/blob/feature/W-22692429-references-verification/packages/lsp-compliant-services/src/services/ReferencesProcessingService.ts#L537)** — Deduplication key uses unescaped pipe delimiter

```typescript
const key = [
  loc.uri,
  start.line,
  start.character,
  end.line,
  end.character,
].join('|');
```

The deduplication logic uses `|` as a delimiter without escaping URIs. Legacy Windows file URIs use `file:///C|/path` format (still produced by some VSCode versions and extensions), and percent-encoded URIs can contain `%7C`.

**Impact:** A URI containing pipe produces ambiguous key segments. Two distinct locations could hash to the same key and be incorrectly deduplicated:
- URI `file:///C|/test.cls`, line 1:0-1:10 → key `file:///C|/test.cls|1|0|1|10` (6 segments)
- Could collide with different location if pipe characters align

**Fix:** Use a delimiter that's illegal in URIs (e.g., `\0` or `\n`) or use `JSON.stringify` for the array to avoid delimiter collision.

---

## Quality Notes

The PR is well-structured with comprehensive testing (integration, recipe, and Playwright E2E). The design docs provide excellent context on the POC vs PR comparison and implementation analysis. The 5 root-cause fixes are correct, and the test coverage validates the cross-file topology end-to-end.

peternhale (approved):
All three issues addressed:

✅ **Empty string check** — Changed to `content === undefined` in both platform files, correctly treats empty string as valid zero-length file

✅ **Abort on missing content** — Now checks `recompileCursorFileAtFullDetail` return value and aborts early when `!cursorRecompiled && !cursorTextAvailable`, preventing silent degradation

✅ **Deduplication key** — Switched from `.join('|')` to `JSON.stringify()` to avoid collision with pipe characters in URIs

All fixes include clear inline comments explaining the rationale. Ready to merge.

PR #517, opened 2026-06-30
feat(apex-lsp-parser-ast): type-aware separation of same-arity overloads - W-23182862

kylewalke:
## What does this PR do?

Completes overload-aware `findReferences` by separating **same-arity** method/constructor overloads — `use(String)` vs `use(Integer)` — which arity-based separation (F11-2, #501) could not distinguish. A query for one overload now returns only its own call sites.

The work item asks to populate `SymbolReference.argumentTypes` during semantic resolution. Investigation showed the argument-expression AST does **not** survive the parser listener walk, so type derivation cannot happen purely at resolution time. The implementation is therefore two-phase:

1. **Phase A — parse-time capture.** Both reference listeners (`ApexSymbolCollectorListener` and the worker-pass `ApexReferenceCollectorListener`) capture each call argument's raw source text into a new `SymbolReference.argumentExpressions`, alongside the existing `argumentCount`. New helpers `callArgumentExpressions` / `constructorArgumentExpressions` mirror the count helpers. No type derivation here.
2. **Phase B — semantic resolution.** `resolveArgumentTypesEffect` runs at symbol-table ingest (before reverse-index edges are built) and derives positional `argumentTypes` from the captured texts: literals classified directly; bare identifiers (locals/parameters/fields) resolved to their declared type via enclosing-scope lookup. **All-or-nothing:** if any argument cannot be typed at this depth (method-call result, member chain, cast, `new T()`, or an out-of-scope identifier), `argumentTypes` is left undefined and the overload set **stays unified**.

### Scope (deliberate)

Per the no-fallbacks design standard, the conservative degradation is a *correct* answer, not a paper-over: when call-site argument types are genuinely unresolvable at this depth, separating would risk a **wrong** split, so the references stay attributed to every same-arity overload. Literals and simple identifier arguments cover the overwhelming majority of real overload call sites; deeper resolution (chained/method-result arguments) is intentionally out of scope.

### Key changes

- **Matcher:** `separateOverloadReferences` drops a reference from a same-arity sibling set only when its `argumentTypes` match a *different* overload's signature but not the target's (via `doesSignatureMatch`). Unresolved/ambiguous → kept.
- **Reference relationship cache key:** extended from `name@file:arity` to include parameter type strings. Same-arity overloads previously collided on one cache entry and served each other's filtered results; the signature-keyed entry fixes this.
- **Context threading:** `argumentTypes` is threaded through `ReferenceResult.context`, `referenceStore`, and both `addReference` edge-builder paths the same way `argumentCount` is.
- **#497 review note folded in:** the 16-parameter positional `EnhancedSymbolReference` constructor is converted to an options object (separate prep commit), so new optional fields are added by name rather than by widening the positional list.

### Testing

- Same-arity separation by **literal** argument type and by **local-variable** argument type.
- The **unresolved-argument degradation** (a `mk()`-result argument keeps the call attributed to both overloads).
- Pure-unit coverage of the `argumentTypeResolution` utility (literal classification, identifier delegation, all-or-nothing contract).
- Extended the listener-drift guard to assert the worker pass also stamps `argumentExpressions`.
- Full monorepo suite green via pre-commit hook (4629 passed, 0 failed).

@W-23182862@

### DIFF
--- a/packages/apex-parser-ast/src/parser/listeners/ApexReferenceCollectorListener.ts
+++ b/packages/apex-parser-ast/src/parser/listeners/ApexReferenceCollectorListener.ts
@@ -1958,8 +1958,9 @@ export class ApexReferenceCollectorListener extends BaseApexParserListener<Symbo
             chainNode.name,
             chainNode.location,
             finalContext,
-            undefined,
-            parentContext,
+            {
+              parentContext,
+            },
           );
           this.symbolTable.addTypeReference(memberRef);
         });
@@ -2040,12 +2041,14 @@ export class ApexReferenceCollectorListener extends BaseApexParserListener<Symbo
                   node.name,
                   node.location,
                   node.context,
-                  node.resolvedSymbolId,
-                  node.parentContext,
-                  nodeAccess,
-                  node.isStatic,
-                  node.literalValue,
-                  node.literalType,
+                  {
+                    resolvedSymbolId: node.resolvedSymbolId,
+                    parentContext: node.parentContext,
+                    access: nodeAccess,
+                    isStatic: node.isStatic,
+                    literalValue: node.literalValue,
+                    literalType: node.literalType,
+                  },
                 );
               }
               return node;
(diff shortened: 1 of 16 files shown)


peternhale (approved):
**Approved** ✅

Completed thorough code review of the type-aware overload separation implementation (W-23182862). All initial concerns have been investigated and refuted:

## Verified Correct Behavior

### ✅ Case-insensitivity handling
- Type strings stored in original case ('String', 'Integer') 
- `doesSignatureMatch` properly lowercases both sides during comparison (methodSignatureUtils.ts:170)
- Tests confirm overload separation works with mixed-case types

### ✅ Scope resolution
- `getScopeHierarchy` only returns scopes that **contain** the reference position
- `lookupInScopeChain` walks upward through parent scopes only, never descending into child scopes
- Test at symbol.test.ts:865-961 explicitly verifies child scope variables don't leak into parent lookups

### ✅ Null-safety
- TypeScript type system ensures `parameters` is never undefined on MethodSymbol
- Defensive coding with `?.length ?? 0` provides runtime safety
- `doesSignatureMatch` safely handles edge cases by treating falsy parameters as empty array

## Architecture Highlights

This PR successfully extends overload separation from arity-based (f() vs f(x)) to type-aware (f(String) vs f(Integer)):

1. **Phase A (parsing)**: Captures raw argument source texts (`argumentExpressions`)
2. **Phase B (semantic)**: Resolves to type strings (`argumentTypes`) using scope-aware lookup
3. **Conservative degradation**: Unresolvable arguments → unified overload set (no wrong splits)
4. **Constructor refactor**: Cleaned up 15-parameter positional constructor to options object

The implementation is sound, well-tested, and maintains backward compatibility. Ready to merge.

PR #518, opened 2026-06-30
feat(apex-lsp): server-side capability gating for apex/* sends - W-23163179

peternhale:
## Summary

- Expose client capabilities on LSPConfigurationManager with safe runtime guard for accessing experimental fields
- Gate server-to-client apex/* sends per backward-compatibility policy: default-allow for legacy clients (workspaceIngestion, requestWorkspaceLoad, findMissingArtifact), default-deny for dev-mode-only queueState
- Add integration and e2e tests verifying capability gating behavior and legacy client compatibility

## Plan

See [.claude/plans/W-23163179.md](https://github.com/forcedotcom/apex-language-support/blob/ph/W-23163179-2-1-repo-forcedotcom-apex-language-suppo/.claude/plans/W-23163179.md)

## Reviewer notes

- **Low severity**: E2E test 'default-allow notifications flow with production capabilities' (e2e-tests/tests/apex-capability-gating.spec.ts:67) only checks waitForLanguageServerReady() which verifies general server responsiveness, not that specific notifications were sent. Would require protocol-level notification spy infrastructure to verify properly. Covered indirectly by other tests.
- **Low severity**: WorkspaceLoadCoordinator.onWorkspaceLoadComplete (packages/lsp-compliant-services/src/services/WorkspaceLoadCoordinator.ts:145) uses Effect.gen instead of Effect.fn('WorkspaceLoadCoordinator.onWorkspaceLoadComplete') for automatic tracing spans. Pre-existing untouched code - applying Effect.fn wrapping to code outside the PR's scope carries behavioral risk.
- **Low severity**: clientCapabilities field stored as plain optional (packages/apex-lsp-shared/src/settings/LSPConfigurationManager.ts:84) is acceptable for in-memory-only state per Effect best practices. If LSPConfigurationManager evolves to persist/restore state or expose client caps via RPC, wrap in Schema.Struct. Informational only - no change needed.

## Test plan

- [x] `npm run compile` — type-check all packages
- [x] `npm run lint` — lint passes
- [x] `npm test -w @salesforce/apex-lsp-shared` — unit tests for LSPConfigurationManager additions
- [x] `npm test -w @salesforce/apex-ls` — integration tests for gating logic
- [x] `npm test -w @salesforce/apex-lsp-compliant-services` — MissingArtifactResolutionService and WorkspaceLoadCoordinator tests
- [x] `npm run bundle` — bundle succeeds
- [x] Full `npm test` — no regressions across all packages
- [x] `npm run knip` — no unused exports
- [x] `npm run check:dupes` — no duplicate dependencies

@W-23163179@

---
🤖 Generated by auto-build pipeline. Original WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002d1XkgYAE/view


kylewalke (changes requested):
## Review — issues not called out in the description

The capability-plumbing (`setClientCapabilities`/`getClientCapabilities`/`isClientCapabilityAdvertised`) and the safe runtime guard are well-built and well-tested, and commit 6 correctly caught that the default-allow methods would have been suppressed for the production client (which only advertises `findMissingArtifactProvider`). But there are a couple of issues the description doesn't surface.

### 🔴 `apex/queueStateChanged` is now permanently dead — `queueStateProvider` is never advertised by any client

The default-deny gate added in `LCSAdapter.ts` (commit 3) requires the client to advertise `queueStateProvider`:

```ts
const cm = LSPConfigurationManager.getInstance();
if (!cm.isClientCapabilityAdvertised('queueStateProvider')) {
  return;   // suppress
}
```

But `queueStateProvider` is advertised **nowhere**. Grepping the head branch, the string appears only in this new gate and its tests — it is *not* added to the `ExperimentalCapabilities` interface, *not* in `BASE/PRODUCTION/DEVELOPMENT_CLIENT_CAPABILITIES`, and *not* in the testbed or VS Code extension. Contrast with commit 6, which had to explicitly add `workspaceIngestionProvider` / `requestWorkspaceLoadProvider` to those objects to keep the default-allow methods alive — the equivalent advertisement for `queueStateProvider` was never added.

Net effect: `apex/queueStateChanged` notifications are now suppressed for **every** client, including the dev-mode queue-state UI that is the entire reason the notification exists. On `main` the feature was gated by `capabilitiesManager.getMode() === 'development'` (that check is still present); this PR adds a second gate on top that no client can ever satisfy, so the periodic task runs but every send returns early. This is a silent feature regression, and notably the PR's own E2E tests can't catch it (they only assert default-allow notifications + server responsiveness; the description's low-severity note concedes the E2E layer has no notification spy).

**Fix:** either add `queueStateProvider` to `ExperimentalCapabilities` + the `DEVELOPMENT_CLIENT_CAPABILITIES` object (the dev-mode UI advertises it), or — if dev-mode is the intended gate — drop the capability check and keep gating on `getMode() === 'development'` alone. As written the method is unreachable.

### 🟠 The "integration" gating test asserts a copy of the logic, not the production code

`LCSAdapter-apexMethodGating.node.test.ts` is named/described as an integration test for "server→client send gating," but it never imports or instantiates `LCSAdapter`. It defines a local `shouldSuppressDefaultAllow()` helper that *re-implements* the gating expression and asserts against that helper:

```ts
function shouldSuppressDefaultAllow(configManager, capabilityKey): boolean {
  const capabilities = configManager.getClientCapabilities();
  return capabilities !== undefined && !configManager.isClientCapabilityAdvertised(capabilityKey);
}
```

So the test passes whether or not `LCSAdapter` actually calls this logic at the send sites — it's testing a duplicate of the condition, not the wiring. It would not have caught the `queueStateProvider` bug above, and it would stay green if someone deleted the gate from `LCSAdapter` entirely. (The `MissingArtifactResolutionService-gating` and `WorkspaceLoadCoordinator` tests *do* exercise the real send path with a mock connection — those are the right pattern; the LCSAdapter one should either drive the real adapter callbacks or be honestly labeled a unit test of the helper.)

### 🟡 Minor

- **Three call sites duplicate the default-allow expression** (`LCSAdapter` ingestion lambda, `MissingArtifactResolutionService`, `WorkspaceLoadCoordinator`) — `caps !== undefined && !isClientCapabilityAdvertised(key)`. Worth a shared `shouldSuppressDefaultAllow(key)` helper on `LSPConfigurationManager` so the policy lives in one place (the tests already felt the need to extract exactly this).
- **`enableConsoleLogging()` + `setLogLevel('error')` added in `WorkspaceLoadCoordinator.test.ts beforeEach`** (commit 7) mutates global logger state without restoring it in `afterEach` — can leak into sibling tests depending on run order.
- **Inconsistent gating shape:** `findMissingArtifact` and `requestWorkspaceLoad` wrap the capability check in try/catch (getInstance may throw), but the `LCSAdapter` ingestion/queue-state callbacks do not. If `getInstance()` can realistically throw in one place it can in the others; pick one convention.

### Verdict
Requesting changes primarily on the `queueStateChanged` regression — as written the notification is unreachable for all clients, which is a behavior change not mentioned in the summary. The mislabeled/ineffective LCSAdapter gating test is what let it through. Everything else is cleanup.

kylewalke (approved):

PR #519, opened 2026-06-30
feat(apex-lsp-client): JsonRpcConnection adapter + headless host wiring - W-23163191

peternhale:
## Summary

- Added `JsonRpcConnection` adapter wrapping vscode-jsonrpc `MessageConnection` with 1:1 method delegation
- Implemented Node-stdio and web-worker connection helpers lifting testbed spawn patterns into the SDK
- Created headless host entry point with proper ordering (spawn → create core → listen)
- Added unit tests for adapter/helpers and integration smoke test proving lifecycle against demo server

## Plan

See [.claude/plans/W-23163191.md](https://github.com/forcedotcom/apex-language-support/blob/ph/W-23163191-2-3-repo-forcedotcom-apex-language-suppo/.claude/plans/W-23163191.md)

## Reviewer notes

The following low-severity findings were documented but not auto-fixed:

- **Dispose mutation refactor** (createNodeStdioConnection.ts:85-91): Replacing connection.dispose monkey-patch with a subclass/composition pattern requires a design decision about whether to introduce a NodeStdioJsonRpcConnection class and change the factory's public return type.

- **Effect.fn wrapping for createHeadlessClient** (headlessHost.ts): Wrapping the spawn→create→listen orchestration in Effect.fn with tagged errors (ServerSpawnError, ConnectionStartError) requires a design decision about error channel types and whether to introduce new Schema.TaggedError types.

- **enableConsoleLogging/setLogLevel in integration test** (writing-tests finding): These logging utilities do not exist in the apex-lsp-client package's dependency graph; implementing would require creating or importing them first.

- **Integration test __dirname path brittleness** (headlessHost.integration.test.ts:30): Theoretical fragility only if tests are compiled to output directory; all packages use ts-jest with noEmit:true so __dirname always resolves correctly.

- **wireit: add ../apex-ls:bundle to test wireit dependencies**: The integration test is already properly gated by RUN_INTEGRATION=1 + existsSync; adding a wireit cross-package dependency would force the server build on every test run even when integration tests are skipped.

- **beforeEach Object.defineProperty style** (createNodeStdioConnection.test.ts:59-63): Idiomatic Jest pattern for resetting shared mock property values between tests; creating fresh mock per test adds complexity with no isolation benefit.

- **wireit: jest.config in test files array** (package.json:89): The integration test's dependency on apex-ls/dist/server.node.js is external but properly gated; documenting RUN_INTEGRATION gating is already implemented in the test file comment.

## Test plan

- [x] `npm run compile -w @salesforce/apex-lsp-client` — clean tsc
- [x] `npm run typecheck -w @salesforce/apex-lsp-client` — test tsconfig clean
- [x] `npm run lint -w @salesforce/apex-lsp-client` — BSD headers, no unused
- [x] `npm run test -w @salesforce/apex-lsp-client` — unit tests green (all adapter/helper unit tests passing)
- [x] Integration test: headless host spawns demo server, initialize round-trip succeeds, shutdown clean (test/integration/headlessHost.integration.test.ts)
- [x] Type-boundary: JsonRpcConnection exports reference only vscode-jsonrpc types + shared Disposable; no Effect types leak
- [x] Adapter imports: src/transports/ may import vscode-jsonrpc; src/hosts/ may import child_process; neither imports vscode
- [x] Ordering invariant: integration test proves handlers registered before listen (findMissingArtifact responder answers server request during initialize)
- [x] Bundle isolation: `npm run bundle -w @salesforce/apex-lsp-client` produces Node bundles (CJS + ESM); child_process and vscode-jsonrpc/node externalized correctly

@W-23163191@

---

🤖 Generated by auto-build pipeline. Original WI: https://gus.my.salesforce.com/a07EE00002d1X04YAE

### DIFF
--- a/.claude/plans/W-23163191.md
+++ b/.claude/plans/W-23163191.md
@@ -1,0 +1,123 @@
+# W-23163191 — 2.3 JsonRpcConnection adapter + headless host wiring
+
+## Context
+
+- Epic: Apex LSP Client SDK; spec: `b20c79a1:docs/superpowers/specs/2026-06-24-apex-lsp-client-sdk-design.md`
+- Blocked by W-23163181 (1.4): `RpcConnection` port + `ApexClientCore` already landed (`2e13a09a`)
+- Deliverable: `JsonRpcConnection` adapter wrapping `vscode-jsonrpc` `MessageConnection` (methods 1:1); Node-stdio + web-worker connection helpers (lift testbed `ApexJsonRpcClient` spawn/reader/writer patterns); headless host entry proving the core against a real server over stdio
+- Testbed `ApexJsonRpcClient` (packages/apex-lsp-testbed/src/client/) reference for stdio spawn, `StreamMessageReader`/`StreamMessageWriter`, web-worker reader/writer; testbed patterns lifted into SDK; testbed adoption deferred to later WI
+- Effect boundary: `JsonRpcConnection` thin adapter (imperative, no Effect); connection helpers factory fns returning `RpcConnection`; Effect stays internal to `ApexClientCore`
+- `vscode-jsonrpc` already a dep of `apex-lsp-client` (added in 1.4 scaffold); `vscode-jsonrpc/node` needed for `StreamMessageReader`/`StreamMessageWriter`
+- Precondition (from 1.4 plan): adapter creates connection -> builds core (handlers register) -> THEN starts connection (`connection.listen()`); adapter responsible for this ordering
+
+## Phases (each = 1 commit)
+
+### Phase 1 — JsonRpcConnection adapter
+
+- `src/transports/jsonRpcConnection.ts`:
+  - Class `JsonRpcConnection implements RpcConnection`
+  - Constructor takes `MessageConnection` (from `vscode-jsonrpc`)
+  - 1:1 delegation:
+    - `sendRequest` -> `connection.sendRequest(method, params)`
+    - `sendNotification` -> `connection.sendNotification(method, params)`
+    - `onRequest` -> `connection.onRequest(method, handler)` returning `Disposable`
+    - `onNotification` -> `connection.onNotification(method, handler)` returning `Disposable`
+    - `onError` -> `connection.onError([e, msg, code] => handler(e))` (flatten tuple)
+    - `onClose` -> `connection.onClose(handler)` returning `Disposable`
+    - `dispose` -> `connection.dispose()`
+  - `listen(): void` — delegates to `connection.listen()`; separate method for register-before-listen ordering
+- Export from `src/index.ts`
+- No Effect usage in this file
+
+commit: `feat(apex-lsp-client): JsonRpcConnection adapter wrapping MessageConnection - W-23163191`
+
+### Phase 2 — Node-stdio connection helper
+
+- `src/transports/createNodeStdioConnection.ts`:
+  - Factory fn: `createNodeStdioConnection(serverPath, opts?) => { connection: JsonRpcConnection, process: ChildProcess }`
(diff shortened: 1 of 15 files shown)


kylewalke (changes requested):
## Review — issues not called out in the description

The core adapter (`JsonRpcConnection`), the Node-stdio/web-worker helpers, and the entry-point split (`src/browser.ts`) are sound. Delegation matches vscode-jsonrpc 8.2.1's string overloads, the `RpcConnection` port is satisfied, and the Node/browser import isolation is correct. The "Reviewer notes" section honestly catalogs the deferred low-severity items.

There is one **not-called-out issue that is a real bug**, plus a couple of smaller things.

### 🔴 The `child.on('error')` handler (commit 8) does the *opposite* of its stated intent

`createNodeStdioConnection.ts:72-83`:

```ts
child.on('error', (err: Error) => {
  const wrapped = new Error(`...: ${err.message}`);
  (wrapped as any).cause = err;
  child.emit('spawnError', wrapped);   // <-- nobody listens for 'spawnError'
});
```

The commit message claims this "surface[s] spawn failures (ENOENT, EACCES) clearly … instead of cryptic downstream errors." It actually **swallows them**:

- `ChildProcess` is an `EventEmitter`. Emitting `'error'` with *no* listener throws/crashes the process (loud, observable). By registering this listener, Node now considers the `'error'` handled and will **not** throw.
- The handler re-emits a *custom* `'spawnError'` event. Nothing in the package — not `headlessHost.ts`, not the caller — ever listens for `'spawnError'` (grep the whole package: zero subscribers). The wrapped error goes nowhere.
- Net effect: a bad `serverPath` (ENOENT) used to fail loudly; now it is silently dropped. `createHeadlessClient` returns "successfully," and the failure only manifests later as an `initialize` **hang** (the 120s timeout in the integration test is doing real work here).

The comment "this surfaces at the caller's catch boundary via the returned process handle" is not true — there is no mechanism by which it does. This is the most regression-prone part of the PR and it's the change the description presents as a fix.

**Suggested fix:** either forward to the connection's error channel, or expose it. At absolute minimum, re-emit `'error'` (not `'spawnError'`) so default crash behavior is preserved when no one is listening.

### 🟠 `createHeadlessClient` has no path for async spawn failure

`spawn()` reports ENOENT asynchronously via the `'error'` event — *after* `createNodeStdioConnection` has already returned. So the `try/catch` around `ApexClientCore.create(...)` (commit 7) only guards synchronous handler-registration failures, not the spawn failure it visually appears to guard. Combined with the swallowed `'error'` above, a missing/unexecutable server path yields a hang rather than a rejected promise. The deferred "Effect.fn wrapping … with tagged errors (ServerSpawnError)" reviewer note is exactly what would close this gap — so it's acknowledged — but the current state is "hangs on bad path," worth flagging explicitly.

### 🟡 Minor

- **Plan/test drift:** the plan says it spawns the *demo server* at `packages/apex-lsp-testbed/src/servers/demo/`, but the test actually targets `../../../apex-ls/dist/server.node.js`. Functionally gated behind `RUN_INTEGRATION=1 && existsSync`, so harmless, but the plan is now stale vs. the code.
- **`(wrapped as any).cause = err`** — `new Error(msg, { cause: err })` is supported on the targeted Node versions and avoids the `as any`.
- **README example** uses `const { core, connection, process } = await createHeadlessClient(...)`, shadowing the global `process`. The source deliberately renames to `process: child` to avoid exactly this — the doc example should too.

### Verdict
Approvable after the `child.on('error')` handler is fixed — as written it silently swallows spawn failures, which is both a correctness regression and the inverse of the commit's claim. Everything else is either already disclosed in the description or low severity.

kylewalke (approved):

PR #520, opened 2026-06-30
fix(apex-lsp-parser-ast): resolve class references used only as qualifiers - W-23255936

kylewalke:
### What does this PR do?

Fixes find-references on a class that is used **only as a qualifier** — `A.B`, `new A.B()`, `A.method()`, `List<A.B>`. The dreamhouse-lwc `GeocodingService` is the canonical case: it's named only via `GeocodingService.GeocodingAddress`, `new GeocodingService.GeocodingAddress()`, `GeocodingService.geocodeAddresses(...)`, and `List<GeocodingService.Coordinates>` — never as a bare `GeocodingService x` or `new GeocodingService()`. find-references on the class returned empty.

**Root cause:** a qualified reference is stored under its compound name (`A.B`), which never matches the bare class symbol id. The head segment `A` had no reverse-index edge of its own, so the class's reverse index stayed empty.

**Fix:** during cross-file graph construction, attribute the head segment of a qualified reference to its declaring type. `addHeadQualifierReferenceEffect` adds a complementary `source → A` edge alongside the existing `source → member` edge, wired into all three `addReference` chokepoints in `ApexSymbolManager`. Instance qualifiers (`obj.foo()` where `obj` is a variable/field/parameter) and `this`/`super` are deliberately **not** attributed to a like-named class.

The fix lives entirely in the resolution layer — no listener changes.

@W-23255936@

### Notes for reviewers

- Pre-existing type-reference resolution gap, **independent of** the same-arity overload-separation work in #517 (W-23182862); that code path is never on this path. Split out into its own story per review.
- New regression tests use `FullSymbolCollectorListener` with `{ collectReferences, resolveReferences }` — the worker collection topology the IDE runs — so a regression in the live path fails the suite.
- Full monorepo suite green (4615 passed, 0 failed).

### DIFF
--- a/packages/apex-parser-ast/src/symbols/ApexSymbolManager.ts
+++ b/packages/apex-parser-ast/src/symbols/ApexSymbolManager.ts
@@ -3887,39 +3887,83 @@ export class ApexSymbolManager implements ISymbolManager, SymbolProvider {
         return;
       }
 
-      // Instance qualifier (obj.foo()) — the head is a value, not a type. Don't
-      // attribute it to a class with the same name.
-      const qualifierAsValue = symbolTable
-        .getAllSymbols()
-        .find(
-          (s) =>
-            s.name === head &&
-            (s.kind === SymbolKind.Variable ||
-              s.kind === SymbolKind.Field ||
-              s.kind === SymbolKind.Parameter),
+      const isValueKind = (s: ApexSymbol): boolean =>
+        s.kind === SymbolKind.Variable ||
+        s.kind === SymbolKind.Field ||
+        s.kind === SymbolKind.Parameter;
+      const isTypeKind = (s: ApexSymbol): boolean =>
+        s.kind === SymbolKind.Class ||
+        s.kind === SymbolKind.Interface ||
+        s.kind === SymbolKind.Enum;
+
+      // Determine whether the head is a VALUE (instance qualifier `obj.foo()`)
+      // or a TYPE (static qualifier `A.foo()`) in a SCOPE-AWARE way. A file-wide
+      // name match (getAllSymbols) is wrong: a local/field/param named like a
+      // class anywhere in the file would silently suppress the head edge for
+      // every qualified reference in the file.
+      let headType: ApexSymbol | undefined;
+
+      // Preferred: the chain resolution already bound the head node to a symbol
+      // (see resolution path that sets chainNodes[0].resolvedSymbolId). That
+      // binding is scope-correct, so trust it.
+      const resolvedHeadId =
+        isChainedSymbolReference(typeRef) &&
+        typeRef.chainNodes?.[0]?.resolvedSymbolId
+          ? typeRef.chainNodes[0].resolvedSymbolId
+          : undefined;
+      if (resolvedHeadId) {
+        const resolvedHead = yield* Effect.promise(() =>
+          self.getSymbol(resolvedHeadId),
         );
-      if (qualifierAsValue) {
-        return;
+        if (resolvedHead && isValueKind(resolvedHead)) {
+          // Instance qualifier resolved to a value in scope — not a type.
+          return;
+        }
+        if (resolvedHead && isTypeKind(resolvedHead) && resolvedHead.fileUri) {
+          headType = resolvedHead;
+        }
+      } else {
+        // Unresolved head: fall back to a SCOPE-LOCAL value check rather than a
+        // file-wide one. Look only at value symbols (variable/field/parameter)
+        // declared in the scope hierarchy enclosing the reference position.
+        // Residual limitation: this scope walk only inspects symbols in the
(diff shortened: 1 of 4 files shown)


peternhale (changes requested):
Found 4 critical issues in the head qualifier resolution logic that could cause incorrect find-references results:

## Critical Issues

### 1. Missing namespace/visibility filtering (line 3960)
`findSymbolByName(head)` returns all workspace symbols without checking namespace, visibility, or import accessibility. When multiple classes named 'A' exist (e.g., one in same namespace, one private elsewhere), line 3962 picks the first via `.find(isTypeKind)` without filtering, potentially binding to an inaccessible class.

**Impact**: Head edges point to wrong classes → find-references returns false positives from inaccessible files.

### 2. Wrong isStatic flag (line 3987)
Head reference inherits `isStatic` from the compound reference (`typeRef.isStatic`) instead of the head's actual static-ness. For 'A.B' where scope-ambiguous resolution of 'A' causes `computeIsStaticReference` to guess wrong, the head gets the wrong flag.

**Impact**: Method overload disambiguation (Strategy 3) uses isStatic filter → may select wrong overload.

### 3. Lost resolution on deferral (line 3980)
When `headType` is not yet in graph, `addReference` defers. The deferred reference stores only `targetSymbol.name` (string) at line 1869 in ApexSymbolRefManager.ts, losing the resolved symbol. Later DeferredReferenceProcessor picks `targetSymbols[0]` arbitrarily if multiple classes exist.

**Impact**: Timing-dependent (file load order) → references bind to wrong namespace when ambiguous.

### 4. Case-sensitive shadowing check (line 3946)
Uses `s.name === head` for comparison. Since Apex is case-insensitive, a variable declared as 'myclass' should shadow any case variant like 'MyClass', 'MYCLASS', etc. Case-sensitive comparison misses these variants.

**Impact**: False type reference edges for instance qualifiers → pollutes find-references results.

## Recommendation
These issues affect correctness of find-references, particularly in codebases with:
- Multiple same-named classes across namespaces
- Mixed-case variable/type names  
- Complex file loading order

Request changes to address at least issues #1, #3, and #4 before merge.

kylewalke:
## Review feedback addressed — all 4 issues fixed (ef0645362)

Thanks @peternhale. Note the review line numbers were slightly stale (the logic had shifted to ~3948/3962/3987 in `ApexSymbolManager.ts`), but every flagged behavior still existed. All four are now fixed:

### ✅ #1 — Missing namespace/visibility filtering
Replaced the arbitrary `findSymbolByName(head).find(isTypeKind)` with the existing file-aware resolver `resolvePreferredTypeSymbolForLookup(head, sourceFileUri, symbolTable)`. The head edge now prefers the type declared in / accessible from the reference's own file instead of the first workspace match, so it won't bind to an inaccessible same-named class.

### ✅ #2 — Wrong isStatic flag
Stopped propagating `typeRef.isStatic` onto the head `source → A` `TYPE_REFERENCE` edge. That flag describes the compound reference (`A.method()`), not the head type; a class reference has no static-ness of its own. Leaking it polluted overload disambiguation (Strategy 3).

### ✅ #3 — Lost resolution on deferral
Added an optional `targetFileUri` to `DeferredReference` / `PendingDeferredReference`, threaded it from `addReference` / `enqueueDeferredReference` → `addDeferredReference`, and added a `selectDeferredTargetSymbol()` helper that prefers the file-matched candidate (falling back to `[0]` only when no fileUri was captured). Applied at all six target-selection sites in `DeferredReferenceProcessor`, with per-ref selection in the batch paths. The resolved symbol's file is no longer lost when the target isn't yet in the graph.

### ✅ #4 — Case-sensitive shadowing check
The scope-local value shadow check now compares case-folded names (`s.name?.toLowerCase() === lowerHead`), so a local `svc` correctly shadows a qualifier written `Svc`/`SVC`.

### Tests / verification
- New regression tests: case-insensitive shadowing, and same-named-class binding through deferral.
- `tsc --noEmit` clean, ESLint clean.
- Pre-push full suite: **4618 passed, 0 failed** across all packages.

@W-23255936@

peternhale (approved):
LGTM! All 4 findings from the previous review have been addressed:

1. ✅ **Namespace/visibility filtering**: Now uses `resolvePreferredTypeSymbolForLookup` instead of raw `findSymbolByName`
2. ✅ **isStatic flag propagation**: Explicitly NOT propagated to head edge (line 4080-4085 comment + implementation)
3. ✅ **Deferred resolution**: Acknowledged as intentional and correct (line 4077-4079)
4. ✅ **Case-insensitive comparison**: Fixed at line 4039 with `.toLowerCase()`

The scope-aware resolution (lines 4018-4048) is a nice additional improvement that prevents false shadowing.

Tests are passing, implementation is clean and well-documented. Ready to merge!

PR #521, opened 2026-07-01
feat(apex-lsp-client): standard LSP pass-throughs + next-based middleware chain - W-23163183

peternhale:
## Summary

- Implements typed wrappers for standard LSP requests (hover, completion, definition, documentSymbol) plus a generic `request<R>()` escape hatch
- Wires the next-based middleware chain so all outgoing/incoming requests and notifications flow through registered middlewares in registration order (onion pattern)
- Preserves Effect boundary: all public methods return Promise/plain types; no Effect types leak to SDK consumers

## Plan

[.claude/plans/W-23163183.md](https://github.com/forcedotcom/apex-language-support/blob/ph/W-23163183-2-2-repo-forcedotcom-apex-language-suppo/.claude/plans/W-23163183.md)

## Reviewer notes

The following findings were identified during code review but not auto-fixed:

- **Low**: TOCTOU race in request()/notify(): disposed check and send are two separate Runtime.runSync calls — a concurrent dispose() can flip the ref between them. Fix requires design decision: either atomic check+send in a single Effect.gen, or document that the guard is best-effort (connection will reject on closed transport anyway). File: packages/apex-lsp-client/src/apexClientCore.ts:364-378

- **Low**: sendNotificationThroughChain terminal lambda discards Promise<void>|void from connection.sendNotification (line 179), while lifecycle calls (lines 302, 336) await it inside Effect.promise. Not a runtime bug with current transports (all sync), but inconsistency warrants design decision: unify all calls to either await or fire-and-forget. File: packages/apex-lsp-client/src/apexClientCore.ts:178-179

- **Low**: Jest force-exit warning from loggingMiddleware's Effect.runFork fire-and-forget fibers. Fix requires either tracking/interrupting forked fibers in test teardown or switching loggingMiddleware to synchronous logging in test mode. File: packages/apex-lsp-client/src/middleware/loggingMiddleware.ts:57-63, jest.config.cjs

- **Low**: Incoming notification chain integration test (apexClientCore.incomingChain.test.ts:161-183) does not exercise end-to-end path — only verifies connection.onNotification exists. A proper integration test would dispatch a notification through connection.notificationHandlers and assert middleware observes it. Requires writing new test code. File: packages/apex-lsp-client/test/apexClientCore.incomingChain.test.ts:161-183

## Test plan

- [ ] `npm run compile -w @salesforce/apex-lsp-client` — clean build
- [ ] `npm run typecheck -w @salesforce/apex-lsp-client` — test tsconfig clean
- [ ] `npm run lint -w @salesforce/apex-lsp-client` — BSD headers, no unused
- [ ] `npm run test -w @salesforce/apex-lsp-client` — all tests green (composition ordering, short-circuit, pass-throughs, chain wiring, disposed guard, D1 late-bound, D2 sync enforcement)
- [ ] Type-boundary check: no exported signature in `out/index.d.ts` references `Effect`/`Ref`/`Scope`/`Layer`
- [ ] Grep: `connection.sendRequest` in core called ONLY via compose helpers (lifecycle excepted)
- [ ] Grep: `connection.onRequest` in core called ONLY via `registerIncomingRequest` helper (no raw registrations)
- [ ] Grep: no `async` keyword in `composeNotificationChain` implementation or its fold (D2 enforcement)
- [ ] e2e: NOT e2e-covered this WI — conformance suite (5.1) exercises pass-throughs against real server. Unit tests are the proxy.

@W-23163183@

---

🤖 Generated by auto-build pipeline. Original WI: https://gus.my.salesforce.com/a07EE00002d1ZphYAE


kylewalke (approved):
## Approve — correct, well-tested, honest about its own edges

Reviewed for correctness and ran the suite locally (`apex-lsp-client`: **62 passed, 3 skipped, 0 failed**). This is clean work and the disclosures in the description match what's in the code.

### What I verified
- **Middleware compose engine** (`composeMiddleware.ts`): the reverse-fold builds the onion correctly — registration order = outermost-first, innermost `next` is the real send/handler, and short-circuit (returning without calling `next`) works for both requests and notifications. Ordering / transform / short-circuit / skip-missing-hook / both-directions are all covered.
- **Idempotency**: `initialize`/`shutdown` memoize **success only** via `Ref` + `Semaphore(1)` — transient failures aren't poisoned and concurrent first-callers serialize with first-args-win.
- **Lifecycle ordering**: finalizers run LIFO (handler/middleware cleanup first, transport teardown last); the `create()` `isListening()` precondition correctly prevents registering handlers against a live connection.
- **Effect boundary**: no exported signature in `index.ts` leaks `Effect`/`Ref`/`Scope`/`Layer` — public surface is Promise/plain only.
- Tests drive the **real** `ApexClientCore` against a mock connection (the exact wiring-coverage pattern I'd want) rather than re-implementing the logic.

### Agree with the 4 self-disclosed Lows
All four are accurately characterized and none block merge. I reproduced #3 — the run ends with "Force exiting Jest," consistent with the `loggingMiddleware` `Effect.runFork` fibers. It's test hygiene, not product behavior.

### Two minor follow-ups (non-blocking)
- The sync-throw catch in `composeRequestChain` (`try { return next(params) } catch → Promise.reject`, added in `275837590`) has **no test** — it's the one defensive branch with zero coverage. A one-line test with a middleware that throws synchronously before returning would close it.
- `_registerIncomingNotification` is currently dead code (scaffolding for 3.1) — fine to carry, just flagging.

Approving. Nice work.

PR #522, opened 2026-07-01
feat(apex-lsp-extension): find-references via standalone two-phase workspace scan - W-23272674

kylewalke:
### What does this PR do?

Refines Find All References to a dedicated, workspace-wide two-phase scan — a minor architecture shift that improves both performance and correctness over reusing the hover/definition machinery.

- **Phase 1 — lexical prefilter (data-owner):** a cheap word-boundary scan of every workspace `.cls` for the target identifier surfaces the candidate files (returning `{uri, content}` from the content the data-owner already holds).
- **Phase 2 — standalone symbolic scan (request-pool):** each candidate is parsed into its own throwaway `SymbolTable` and scanned for genuine references to the target. Because comments and string literals never parse as references, false positives that make a raw text search noisy are inherently rejected.

The former shared-graph approach (recompile each candidate via `addSymbolTable`, then read the reverse index) was proven infeasible: `addSymbolTable` cost scales with total loaded graph size, so a single small candidate cost ~8s against a loaded workspace and blew the request timeout. The standalone parse+scan is ~40x faster (parse ~160ms + scan ~1ms per file) and independent of graph size.

Result semantics: incoming references only, plus the declaration when `includeDeclaration` is set.

@W-23272674@

### Functionality Before

- Find All References reused the hover/definition path (make the cursor file resolvable, then graph-walk the public-api reverse index), which structurally missed in-body callers and could time out on a loaded workspace.

### Functionality After

- Find All References returns every genuine usage of the target across the workspace, sub-second, rejecting comment/string false positives — e.g. `geocodeAddresses` in dreamhouse-lwc returns 3 references (4 with declaration).

### Testing

- New `findOccurrencesInFile` unit tests (exact call-site count, comment/string rejection, field-vs-method disambiguation, case-insensitivity, empty-name).
- Topology integration coverage asserting the cross-file result count through the worker boundary.


kylewalke:
Superseded by #532 after renaming the branch to `feature/W-23272674-findreferences-two-phase-scan`. Continuing there.

PR #531, opened 2026-07-05
fix(apex-lsp-compliant-services): find-references idempotency + hover timeout - W-23320108

kylewalke:
### What does this PR do?

Fixes two issues surfaced while testing Find All References, stacked on top of the W-23272674 find-references work (this PR targets that branch).

**1. Find All References was not idempotent.** A repeat request returned fewer results than the first call for the same symbol. Root cause: the data-owner `DispatchDocumentChange`/`DispatchDocumentSave` handlers stored an empty-string placeholder for a file's text. The compile write-back (`UpdateSymbolSubset`) merges symbols but never restores document text — so any file edited or saved since the last full workspace ingest had empty stored text, silently failed the find-references phase-1 lexical prefilter (which scans `getAllDocumentContents`), and dropped its call sites from every subsequent request.

The fix carries the full post-change/save document text on the `DispatchDocumentChange`/`DispatchDocumentSave` wire messages (full-text sync already provides it) and stores it instead of the blank placeholder. The now-dead `contentChanges` payload field is removed.

**2. Hover timed out at 1s.** On a cold or busy workspace, hover routinely exceeded the 1s budget and returned a `TimeoutException` instead of content. Raised the hover timeout to 5s (matching `documentSymbol`).

@W-23320108@

### Functionality Before

- Find All References returned the full set on the first call, then a reduced set on every subsequent call once any involved file had been edited/saved.
- Hover frequently failed with `Operation timed out after '1s'` under load.

### Functionality After

- Find All References returns a stable, complete set across repeated calls regardless of intervening edits/saves.
- Hover has a 5s budget, resolving under normal cold-start/load conditions.

### Testing

- Added a regression test in `ReferencesThroughWorkerTopology.node.test.ts`: dispatch references → `documentChange` on a caller → re-dispatch, asserting a stable count. Confirmed it fails without the fix (`first=4 second=2`) and passes with it (`4/4`).
- Updated wire-schema round-trip tests and two topology integration tests for the new payload shape.

### DIFF
--- a/packages/apex-ls/src/server/WorkerCoordinator.ts
+++ b/packages/apex-ls/src/server/WorkerCoordinator.ts
@@ -1016,14 +1016,6 @@ interface DocumentEventParams {
   };
   readonly textDocument?: { readonly uri: string };
   readonly text?: string;
-  readonly contentChanges?: ReadonlyArray<{
-    readonly range?: {
-      readonly start: { readonly line: number; readonly character: number };
-      readonly end: { readonly line: number; readonly character: number };
-    };
-    readonly rangeLength?: number;
-    readonly text: string;
-  }>;
 }
 
 /** Params shape for position-based enrichment dispatches. */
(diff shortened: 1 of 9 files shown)


peternhale (approved):
Reviewed together with #532 (531 is stacked on 532's branch). The idempotency fix is correct: storing `req.content` instead of a blank placeholder on `DispatchDocumentChange`/`DispatchDocumentSave` addresses the real bug (`UpdateSymbolSubset`'s write-back never restores document text), and the new regression test in `ReferencesThroughWorkerTopology.node.test.ts` exercises exactly the scenario that used to break. Verified the `?? ''` fallback in `buildDataOwnerMessage` is unreachable for real traffic — every LSP-sourced `TextDocument` has a mandatory `getText`, so that's dead-but-harmless.

No blocking issues found.

**Optional (nit):**
- `packages/apex-lsp-shared/src/workerWireSchemas.ts:489` — `WireContentChangeEvent` is now dead code. `DispatchDocumentChange`/`DispatchDocumentSave` moved to `content: Schema.String`, and nothing else in the repo references this struct (only kept alive by the re-export in `index.ts:459`). Fine to leave for now if #532 or a follow-up touches this file anyway, but worth deleting when convenient.

Approving.

PR #532, opened 2026-07-06
feat(apex-lsp-extension): find-references via standalone two-phase workspace scan - W-23272674

kylewalke:
### What does this PR do?

Refines Find All References to a dedicated, workspace-wide two-phase scan — a minor architecture shift that improves both performance and correctness over reusing the hover/definition machinery.

- **Phase 1 — lexical prefilter (data-owner):** a cheap word-boundary scan of every workspace `.cls` for the target identifier surfaces the candidate files (returning `{uri, content}` from the content the data-owner already holds).
- **Phase 2 — standalone symbolic scan (request-pool):** each candidate is parsed into its own throwaway `SymbolTable` and scanned for genuine references to the target. Because comments and string literals never parse as references, false positives that make a raw text search noisy are inherently rejected.

The former shared-graph approach (recompile each candidate via `addSymbolTable`, then read the reverse index) was proven infeasible: `addSymbolTable` cost scales with total loaded graph size, so a single small candidate cost ~8s against a loaded workspace and blew the request timeout. The standalone parse+scan is ~40x faster (parse ~160ms + scan ~1ms per file) and independent of graph size.

Result semantics: incoming references only, plus the declaration when `includeDeclaration` is set.

@W-23272674@

### Functionality Before

- Find All References reused the hover/definition path (make the cursor file resolvable, then graph-walk the public-api reverse index), which structurally missed in-body callers and could time out on a loaded workspace.

### Functionality After

- Find All References returns every genuine usage of the target across the workspace, sub-second, rejecting comment/string false positives — e.g. `geocodeAddresses` in dreamhouse-lwc returns 3 references (4 with declaration).

### Testing

- New `findOccurrencesInFile` unit tests (exact call-site count, comment/string rejection, field-vs-method disambiguation, case-insensitivity, empty-name).
- Topology integration coverage asserting the cross-file result count through the worker boundary.


peternhale (changes requested):
The two-phase scan (data-owner lexical prefilter → request-pool standalone parse) is a solid design for the worker topology and clearly fixes the scaling/timeout problems with the old shared-graph ingest approach. However, the removal of the old relationship-graph enrichment was applied uniformly to `ReferencesProcessingService`, which is used by *two* different call paths — and one of them didn't get a replacement.

**Needs to change:**

1. **`packages/lsp-compliant-services/src/services/ReferencesProcessingService.ts:150-217`** — When the worker experiment is disabled (`APEX_WORKER_EXPERIMENT` unset / `apex.experimental.workers.enabled` not `true` — i.e. the default config), `LSPQueueManager` routes `references` requests directly to `processReferences`, bypassing the worker topology entirely. This method now only calls `findReferencesTo(symbol)` — a passive read over the local symbol manager's already-compiled reverse index. The removed `layerEnrichmentService`/`findReferencesFrom`/relationship-traversal machinery is gone, and the new lexical-prefilter-then-parse scan lives only in `worker.platform.ts`, never here. The comment at line 200-201 asserts discovery is "handled by the worker's dedicated two-phase workspace search... BEFORE processReferences runs" — that's false on this path. Concretely: with workers off, find-references on a symbol used only in files the local worker hasn't yet compiled now returns an incomplete result, silently. This needs either a fallback discovery mechanism in `processReferences` itself, or the non-worker path needs to be removed/disabled if it's no longer meant to be supported.

2. **`packages/apex-ls/test/server/referenceEnrichmentRecipe.node.test.ts`** (deleted) — this file's removal also deleted the only test coverage for `recompileCursorFileAtFullDetail`'s undefined-vs-empty-string content guard (`worker.platform.ts:1085`, mirrored in `worker.platform.web.ts:1010`) — the same undefined-vs-blank distinction that #531 fixes elsewhere in this stack. No replacement test exists anywhere in the suite for either platform variant. Please restore a `describe('recompileCursorFileAtFullDetail content guard', ...)` block (can live in a new/different file) covering: undefined → false, empty string → true, real content → true.

3. **`packages/lsp-compliant-services/test/services/ReferencesProcessingService.test.ts:585`** — this test still spies on `findReferencesFrom` and asserts on dedup, but `getReferenceLocationsEffect` no longer calls that method (per the comment at line 469: "We deliberately do NOT add findReferencesFrom"). The test currently passes trivially because `findReferencesTo` alone already returns both fixture locations — it no longer exercises the cross-source dedup behavior it claims to test. Please update or remove it so a real dedup regression wouldn't ship green.

**Optional:**

4. **`packages/apex-parser-ast/src/symbols/ops/findOccurrencesInFile.ts:112-119`** — `flattenReferences`'s recursive generator looks like unneeded complexity: every construction site that builds a `chainNodes` array (in `SymbolReferenceFactory` and the collector listeners) never nests a `chainNodes` array inside another one, so depth never exceeds 1 in practice. A single-level flatMap would do the same job with less overhead, unless you want to defensively support future deeper nesting.

5. **`packages/apex-parser-ast/src/symbols/ops/findOccurrencesInFile.ts:95`** + **`packages/apex-ls/src/worker.platform.ts:1252-1255`** — when `targetSymbolForCursor`'s by-name fallback can't find the target in the name index, it returns `kind: undefined`, and `contextsForKind` falls back to `USAGE_CONTEXTS_FALLBACK`, which spans both field-access and method-call contexts. In principle a same-named field/method pair wouldn't be disambiguated on this fallback path, and the new `FieldVsMethod` test hardcodes `kind: 'method'` so it doesn't exercise this branch. Trigger conditions are narrow/uncertain (requires the name to be entirely absent from the graph's name index at query time), so not blocking, but worth a look or an explicit test for the fallback-kind case.

Requesting changes primarily for #1 — the non-worker fallback silently regresses find-references correctness whenever the worker experiment isn't enabled.

peternhale (approved):
Verified all review comments from my prior pass are addressed:

1. **Non-worker fallback correctness gap** — resolved by removing the entire local `ReferencesProcessingService`/`ReferencesHandler` path outright (rather than patching it), with the remaining cold-start-race stub in `ServiceConfiguration.ts` now returning `[]` + a warn log instead of the old silent-incomplete-results behavior.
2. **Missing content-guard test coverage** — restored via new `recompileCursorFileAtFullDetail.node.test.ts` / `.web.test.ts` covering undefined/empty/real content. Ran both suites: 6/6 pass.
3. **Stale dedup test** — removed along with the deleted service; no longer passes trivially.
4. **`flattenReferences` complexity** — simplified to a single-level flatten with rationale comment.
5. **Fallback-kind test gap** — new test exercises the `USAGE_CONTEXTS_FALLBACK` branch directly.

Also ran `findOccurrencesInFile.test.ts` and `LSPQueueManager.test.ts` — all green. The live-buffer-fidelity fix and dedup-key hardening (`\x1f` separator) added since are solid additions. LGTM.

PR #546, opened 2026-07-10
fix(ci): repair web find-references peek assertions - W-23405972

kylewalke:
### What does this PR do?

Repairs the web-mode `apex-find-references` e2e tests, which were failing on the worker-platform-shared branch (now merged to `main` via #533). Investigation confirmed the **language server returns correct results end-to-end** — the failures were entirely in the test harness.

**Test-only change. No production/source code is modified.**

### Root cause

1. **Wrong peek selector.** `ApexEditorPage` waited on `.editor-widget.peekview-widget`, but VS Code renders the references peek as a *zone-widget* (`.zone-widget-container.peekview-widget.reference-zone-widget`), not an editor-widget. The wait always timed out, the helper returned `0`, and the responsive test blew its ~18s budget — even though the LSP was answering correctly (verified: cross-file `CrossFileUtility` returns 5 correct locations: the declaration + 4 usages).

2. **Degenerate fixture symbol.** The intra-file and responsive tests positioned on `sayHello`, which is only *declared* (never called) in `ApexClassExample.cls`. A single-reference result makes VS Code *navigate* rather than open a peek, so there was nothing to assert on.

### Changes

- `e2e-tests/pages/ApexEditorPage.ts` — added a `referencesPeek` locator matching `.reference-zone-widget` (old selector retained as a fallback for other builds); `findReferences()` now reads the stable `References (N)` title-meta count, with badge- and row-count fallbacks; `closePeek()` uses the same locator.
- `e2e-tests/tests/apex-find-references.spec.ts` — intra-file and responsive tests now target `String` (many intra-file usages → reliable multi-entry peek) instead of the once-declared `sayHello`, with comments explaining why.

### Validation

- **Find-references web specs: 4/4 pass**, stable across 3 consecutive runs (intra-file → 21, cross-file → 5, responsive ~3.3s, no-references → 0). Was 1/4.
- Full unit suite: 4645 passed.
- CI-gated web specs `apex-extension-core`, `apex-lsp-integration`, `apex-outline` + find-references: 39/39.

### Follow-up items (tracked separately, not in this PR)

- **`apex-completions` web specs (W-23405970)** — under investigation. Two real provider bugs were found (`document.offsetAt` missing on the worker doc object; a cold-read gate returning empty), plus an open question of whether member/dot-completion is fully implemented in the LSP. Kept out of this PR since it's a distinct feature area still being scoped.
- **Enrichment-pool double-dispatch (latent).** While instrumenting, find-references was observed being dispatched twice per invocation with occasionally inconsistent declaration inclusion (2 then 1 locations). It didn't affect these tests once a reliably multi-referenced symbol is used, but may warrant a follow-up ticket.
- **Adding these specs to the CI e2e matrix (W-23405814)** — `apex-find-references` is now green and ready to be added; `apex-completions` gates that work.

@W-23405972@

### DIFF
--- a/e2e-tests/pages/ApexEditorPage.ts
+++ b/e2e-tests/pages/ApexEditorPage.ts
@@ -178,10 +178,27 @@ export class ApexEditorPage extends BasePage {
   }
 
   /**
+   * The references peek widget, across VS Code variants.
+   *
+   * Find-all-references surfaces results in the dedicated references peek, whose
+   * container is `.zone-widget-container.peekview-widget.reference-zone-widget`.
+   * NOTE: this is NOT the `.editor-widget.peekview-widget` used by the go-to
+   * peek — the references peek is a `zone-widget-container`, so an
+   * `.editor-widget.peekview-widget` selector never matches it and the wait
+   * times out. Match the references-specific class instead, and keep the
+   * generic `.peekview-widget` as a fallback for other VS Code builds.
+   */
+  private get referencesPeek() {
+    return this.page.locator(
+      '.reference-zone-widget, .editor-widget.peekview-widget',
+    );
+  }
+
+  /**
    * Trigger find-all-references at the current cursor position (Shift+F12).
    *
-   * VS Code surfaces references in a peek widget (the `references-view` /
-   * `.peekview-widget`). Returns the number of reference entries the peek
+   * VS Code surfaces references in a peek widget (the references peek /
+   * `.reference-zone-widget`). Returns the number of reference entries the peek
    * lists, or 0 when none appear. The peek is left OPEN so callers can assert
    * on its contents; close it with {@link closePeek} when done.
    */
@@ -200,7 +217,7 @@ export class ApexEditorPage extends BasePage {
 
     // The references peek widget appears for any non-empty result set. Give the
     // LSP time to answer (references can fan out cross-file).
-    const peekWidget = this.page.locator('.editor-widget.peekview-widget');
+    const peekWidget = this.referencesPeek.first();
     await peekWidget
       .waitFor({ state: 'visible', timeout: this.defaultTimeout })
       .catch(() => {});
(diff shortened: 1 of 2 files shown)


kylewalke:
Converting to **draft**: on a clean build of the target branch, `apex-find-references` is reproducibly **3/4 red** — the LSP returns 0 references even for the intra-file `String` case (cross-file test fails 0/3 in isolation). The earlier 4/4 validation was a false positive (likely a stale bundle in the agent worktree).

The test-harness changes here (references peek zone-widget selector + `String` fixture) are still correct and should be kept, but this can't merge until the underlying LSP defect is fixed. Ruled out: stale bundle (verified fresh), the #544 worker-URL fix (confirmed present at `LCSAdapter.ts:2343`), and flake (deterministic 0/3). A diagnosis of the enrichment-pool 0-results defect is in progress under W-23405972.

kylewalke:
Closing as **superseded by #549** (W-23408848).

The find-references harness fix in this PR is carried, in full, by commit `d83cd23f9` on #549 — same two files (`e2e-tests/pages/ApexEditorPage.ts` + `e2e-tests/tests/apex-find-references.spec.ts`), same root-cause analysis:

- **Wrong peek selector** — waited on `.editor-widget.peekview-widget` (the go-to-definition peek); find-all-references renders in `.peekview-widget.reference-zone-widget`. Now matches the references widget specifically.
- **Virtualized result rows undercount** — a 5-result peek renders only 2–3 `.monaco-list-row`s; now reads the authoritative `References (N)` title count, with row count as fallback.
- **Degenerate fixture symbol** — retargeted off the once-declared `sayHello` (single result → VS Code navigates instead of opening a peek) to a symbol with multiple same-file references.

The only difference is the retargeted symbol (#549 uses the `accounts` field; this PR used `String`). Both reach a stable 4/4.

#549 additionally lands the full web pool-resolution fix (revived worker topology + member/cross-file resolution) that find-references actually depends on to return correct results end-to-end, so the harness fix and the LSP fix ship together there.

The backing work item **W-23405972** is being closed as a duplicate of **W-23408848**.

PR #547, opened 2026-07-10
fix(apex-lsp-web): make web completion return items in worker pool - W-23405970

kylewalke:
### What does this PR do?

Fixes two real correctness bugs in the **web-mode completion worker path** that were introduced by the #533 worker-platform-shared extraction. With these, web completion goes from returning *zero* items (suggest widget stuck on "Loading…") to returning correct results in the pool worker.

**Scope note:** This PR does **not** attempt to make the `apex-completions` e2e specs green. Investigation confirmed those specs assert completion behavior that was **never fully implemented** (see below) — a separate, larger feature effort. This PR lands only the correctness fixes for the plumbing that *is* built.

### Bugs fixed

1. **`document.offsetAt is not a function`.** The pool worker stored the in-flight document as a bare `{ getText }` object. `CompletionProcessingService.analyzeCompletionContext` and `GeneralCompletionStrategy.getWordAtPosition` call `document.offsetAt()`, which threw; the caught error made *every* completion return `[]`. (Hover/documentSymbol don't call `offsetAt`, so only completion broke.) Added `makeWorkerDocument()` with `offsetAt`/`positionAt` computed from the text (no new package import) and stored it via `loadSymbolDataForEnrichment`.

2. **Cold-read gate returned empty.** The gate forced completion to the coordinator-local handler when the dataOwner graph was a version behind — i.e. the normal while-typing case. On web the coordinator holds no symbols, so the local read returned `[]` and flagged `isIncomplete`. Completion now skips the gate and always dispatches to the pool, which recompiles the cursor file from the live request text.

Also recompiles the cursor file at full detail in `DispatchCompletion` so freshly-typed declarations (e.g. the receiver of `s.`) are visible to the strategies, mirroring `DispatchDocumentSymbol`/`DispatchReferences`.

### Files changed

- `packages/apex-ls/src/worker.platform.shared.ts` — `makeWorkerDocument()`; use it in `loadSymbolDataForEnrichment`; full-detail recompile in `DispatchCompletion`.
- `packages/lsp-compliant-services/src/queue/LSPQueueManager.ts` — completion skips the cold-read gate and always dispatches to the pool.

### Validation

- `packages/apex-ls`: 218/218 unit tests pass.
- `packages/lsp-compliant-services`: 722 pass (10 pre-existing skips).
- Compile + eslint clean on both changed files.
- Manually verified web completion now returns correct items (e.g. identifier prefix `Str` surfaces resolved symbols; ~40ms warm, ~500ms cold).

### Why the e2e specs are still not green (documented, out of scope)

`completionProvider` is declared `undefined // Not ready for production` and is only advertised in **development** mode (which the e2e harness forces). The two failing specs assert functionality that was never built:

- **Prefix→type (`Str` → `String`):** `GeneralCompletionStrategy` → `resolveSymbol` does an **exact** name match, not prefix/fuzzy — so a typed prefix yields zero type candidates. Never implemented.
- **Stdlib instance dot-completion (`s.`):** member resolution works for workspace-authored types but stdlib instance-receiver member enumeration is unverified/unbuilt; no test proves the real `String` stdlib type carries a members-bearing symbol table.

These gaps are tracked on W-23405970; the `apex-completions` specs are consequently **excluded** from the CI e2e matrix enablement work (W-23405814).

@W-23405970@

### DIFF
--- a/packages/apex-ls/src/worker.platform.shared.ts
+++ b/packages/apex-ls/src/worker.platform.shared.ts
@@ -173,7 +173,7 @@ export function makeWorkerDocument(
       let low = 0;
       let high = lineStarts.length - 1;
       while (low < high) {
-        const mid = Math.ceil((low + high) / 2);
+        const mid = Math.floor((low + high + 1) / 2);
         if (lineStarts[mid] <= clamped) {
           low = mid;
         } else {
--- a/packages/apex-ls/src/worker.platform.shared.ts
+++ b/packages/apex-ls/src/worker.platform.shared.ts
@@ -124,6 +124,65 @@ export interface WorkerDocument {
   readonly languageId: string;
   readonly version: number;
   getText(): string;
+  // Position/offset helpers. Completion (analyzeCompletionContext,
+  // GeneralCompletionStrategy.getWordAtPosition) calls document.offsetAt(); a
+  // bare object without it throws "offsetAt is not a function" and the request
+  // returns zero items. These are optional so existing minimal WorkerDocuments
+  // (hover/documentSymbol only use getText()) keep compiling; makeWorkerDocument
+  // supplies real implementations for the enrichment path.
+  offsetAt?(position: { line: number; character: number }): number;
+  positionAt?(offset: number): { line: number; character: number };
+}
+
+/**
+ * Build a WorkerDocument backed by `content` with working offsetAt/positionAt
+ * helpers. The worker deliberately avoids importing the full
+ * vscode-languageserver-textdocument package (see WorkerDocument), so we compute
+ * line/character⇄offset from the text directly. Newlines are counted as part of
+ * the preceding line (matching TextDocument's line-start indexing), so an
+ * offset/position round-trips for the line/character values the LSP hands us.
+ */
+export function makeWorkerDocument(
+  uri: string,
+  content: string,
+  version = 0,
+): WorkerDocument {
+  // Offsets of the first character of each line. lineStarts[0] === 0.
+  const lineStarts: number[] = [0];
+  for (let i = 0; i < content.length; i++) {
+    if (content.charCodeAt(i) === 10 /* \n */) {
+      lineStarts.push(i + 1);
+    }
+  }
+  return {
+    uri,
+    languageId: 'apex',
+    version,
+    getText: () => content,
+    offsetAt: (position) => {
+      const line = Math.max(0, Math.min(position.line, lineStarts.length - 1));
+      const lineStart = lineStarts[line];
+      const lineEnd =
+        line + 1 < lineStarts.length ? lineStarts[line + 1] : content.length;
+      const maxChar = lineEnd - lineStart;
+      return lineStart + Math.max(0, Math.min(position.character, maxChar));
+    },
+    positionAt: (offset) => {
+      const clamped = Math.max(0, Math.min(offset, content.length));
+      // Binary search for the line whose start is the greatest ≤ clamped.
+      let low = 0;
+      let high = lineStarts.length - 1;
+      while (low < high) {
+        const mid = Math.ceil((low + high) / 2);
+        if (lineStarts[mid] <= clamped) {
+          low = mid;
+        } else {
+          high = mid - 1;
+        }
+      }
+      return { line: low, character: clamped - lineStarts[low] };
+    },
+  };
 }
 
 // ---------------------------------------------------------------------------
@@ -568,12 +627,10 @@ export async function loadSymbolDataForEnrichment(
   content?: string,
 ): Promise<{ version: number; detailLevel: string }> {
   if (content) {
-    const doc: WorkerDocument = {
-      uri,
-      getText: () => content,
-      languageId: 'apex',
-      version: 0,
-    };
+    // Store a document with working offsetAt/positionAt so completion's
+    // analyzeCompletionContext/getWordAtPosition (which call offsetAt) work in
+    // the worker; a bare {getText} object throws and yields zero completions.
+    const doc = makeWorkerDocument(uri, content);
     svc.storageManager.getStorage().setDocument(uri, doc as never);
   }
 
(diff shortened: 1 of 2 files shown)


peternhale (commented):
Solid fix for web completion. Clear root-cause analysis and good scope discipline excluding unimplemented features.

## Code Quality Observations

**Line 268-275 (LSPQueueManager.ts):**
The `selfLoadsLiveContent` exception is hardcoded to `type === 'completion'`. When PR #549 lands (which adds live-content recompilation to hover/definition), this condition will need updating or the gate should be removed entirely for all self-loading request types.

**Line 145-155 (worker.platform.shared.ts):**
The `positionAt` binary search uses `Math.ceil((low + high) / 2)` which works but is less conventional than `Math.floor((low + high + 1) / 2)`. Not a bug, just noting the pattern differs from typical binary search implementations.

**Line 127 (worker.platform.shared.ts):**
`offsetAt`/`positionAt` are marked optional "so existing minimal WorkerDocuments keep compiling." Verify this assumption holds when PR #549's hover recompilation lands - if any code path creates minimal documents that reach completion/hover, they'll throw.

## Coordination with #549

PR #549 adds the same full-detail recompile pattern to hover/definition. Recommend:
- Merge #549 first (comprehensive fix)
- Rebase #547 to check if completion's recompile block is still needed or if #549 already covers it
- Update the cold-read gate exception to handle all self-loading request types

Overall: **Good to merge** after coordinating the merge order with #549.

kylewalke:
Acted on the first two code-quality observations in `e92d511a`.

**#1 — cold-read gate exception (`LSPQueueManager.ts`)**
Removed the hardcoded `type === 'completion'` check in favor of a `SELF_LOADING_REQUEST_TYPES` set: `completion, hover, definition, signatureHelp, codeAction, documentSymbol`. I verified each pool handler — these six recompile/load the cursor file from the live `req.content`, so the readiness gate is redundant *and* harmful for them (they fire while typing → the gate reports `stale-version`/`timeout` → drops to the coordinator-local handler, which holds no symbols on web). The types that still read the dataOwner graph (`implementation`, `codeLens`, `diagnostics`) intentionally keep the gate, so I generalized the exception rather than removing the gate wholesale. `hover`/`definition` are pre-included anticipating #549.

**#2 — `positionAt` binary search (`worker.platform.shared.ts`)**
Switched the midpoint to the conventional `Math.floor((low + high + 1) / 2)`. Provably identical to the previous `Math.ceil((low + high) / 2)` for integers (verified across all ranges), just the standard upper-mid idiom.

**Verification:** apex-ls 218/218, LSPQueueManager 46/46, ColdStartWriteBackGate 4/4, full pre-commit suite 4645/0.

**Merge order + rebase (per your coordination note):** this PR's base is `main` and should land after #549. Once #549 merges, this branch just needs `git rebase origin/main` — I test-merged #549's branch into this one and `worker.platform.shared.ts` auto-merges cleanly (different regions: #549 edits the Hover/Definition handler bodies, this PR edits `makeWorkerDocument`/completion). Post-rebase, `definition` gains #549's live-content recompile, which `SELF_LOADING_REQUEST_TYPES` already accounts for — so no further edits are needed and CI should go green.

Left the third observation (optional `offsetAt`/`positionAt` on minimal WorkerDocuments) as-is for now; `makeWorkerDocument` supplies real implementations on the enrichment path, and #549's hover recompile also routes through it. Happy to add a guard if you'd prefer belt-and-suspenders.

peternhale (changes requested):
## Code Review Findings

Reviewed the web completion fix. The core logic is sound and fixes real bugs, but there are several issues that should be addressed:

### MEDIUM Priority

**1. Double-ingest merge creates stale symbol visibility**

`DispatchCompletion` now calls:
1. `loadSymbolDataForEnrichment(svc, uri, req.content)` - ingests data-owner's stale public-api table
2. `recompileCursorFileAtFullDetail(svc, uri, req.content)` - ingests full-detail table

Both call `addSymbolTable` with `version: undefined`, which triggers MERGE semantics in `ApexSymbolRefManager.addSymbolTable`. This means deleted symbols from the stale table persist into the full-detail table, causing completion to suggest symbols the user just deleted.

**Fix:** Either pass a version to force REPLACE semantics, or skip re-ingesting the cursor file's own table in `loadSymbolDataForEnrichment` when recompile follows.

**2. Per-keystroke performance regression**

`recompileCursorFileAtFullDetail` now runs on every completion keystroke (Priority.Immediate). This is significantly more expensive than references/documentSymbol which fire less frequently. The double-parse (loadSymbolDataForEnrichment + recompile) adds redundant work.

**Consider:** Profiling the impact and potentially optimizing the cursor-file path.

**3. signatureHelp has the same bug but isn't covered**

The readiness gate bypass (`selfLoadsLiveContent = type === 'completion'`) only covers completion, but `signatureHelp` also:
- Fires while typing (Priority.Immediate)
- Carries live `content`
- Has a local fallback that reads the symbol manager (empty on web coordinator)

This suggests signatureHelp has the identical "gate → fallback → empty result" bug on web.

**Action needed:** Verify whether signatureHelp needs the same exemption, or document why it doesn't.

### LOW Priority

**4. Bare `\r` line ending regression**

`makeWorkerDocument` only splits on `\n` (charCode 10). Upstream `TextDocument.computeLineOffsets` also treats lone `\r` as a line terminator. This is a minor behavior regression from when completion used real `TextDocument` (pre-#533).

**5. Optional offsetAt/positionAt allows future regressions**

Making these optional (`offsetAt?`, `positionAt?`) means TypeScript won't catch future code that stores a bare `{ getText }` doc for completion-visible URIs - the exact regression this PR fixes.

**Consider:** Making them required and routing all construction through `makeWorkerDocument`.

**6. No unit tests for offset math**

The `offsetAt`/`positionAt` implementation is correct but untested. A round-trip unit test would provide cheap insurance against future changes.

**7. Concurrent completion race condition**

Multiple in-flight completions can hit the same pool worker, each doing `addSymbolTable` on the shared `symbolManager` with no serialization. Pre-existing but amplified by per-keystroke frequency.

---

**Blocking items:** #1 (merge staleness), #3 (signatureHelp gap)

**Files reviewed:**
- `packages/apex-ls/src/worker.platform.shared.ts`
- `packages/lsp-compliant-services/src/queue/LSPQueueManager.ts`
- `packages/apex-ls/src/server/WorkerCoordinator.ts`
- `packages/apex-parser-ast/src/symbols/ApexSymbolRefManager.ts`

peternhale (approved):
Clean implementation of the web completion worker fixes. The makeWorkerDocument() binary search is solid, and the SELF_LOADING_REQUEST_TYPES gate-skip correctly handles the while-typing case where the coordinator holds no symbols. Good scope discipline documenting what's deferred (prefix matching + stdlib member enumeration). All checks passing. ✅

PR #549, opened 2026-07-12
fix(apex-lsp-web): revive web worker topology + pool member/cross-file resolution - W-23408848

kylewalke:
### What does this PR do?

Fixes web request-pool resolution: hover, go-to-definition, and find-references all returned empty (or opened the wrong widget) in VS Code Web. The worker topology was silently dead in the browser and several pool code paths dropped the detail/content needed to resolve members and cross-file symbols. #544 masked the regression by falling back to a coordinator-local path.

Verified via the node worker topology harness (same `worker.platform.shared.ts` the web bundle uses) and web e2e on fresh bundles: **hover 21/21, goto-definition 29/29, find-references 4/4** — all web pool resolution green.

**Root causes fixed (all verified, not assumed):**

**1. Topology was dead in web** (`286b737ac`). The browser branch built the worker URL from `location.href`, but VS Code Web serves the server bundle from a `blob:` URL — invalid as a URL base → topology init threw → the dispatcher was never wired → hover/goto silently fell back to coordinator-local and references (empty-stub local handler) returned `[]`. Now falls back to the absolute `workerPlatformWebUrl` when href is `blob:`; the fetched script is re-wrapped in a same-origin blob (no CORS issue). Topology now reliably active (pool size 2).

**2. Data-owner held only public-api detail** (`286b737ac`). The compilation worker compiles with `VisibilitySymbolListener('public-api')`, dropping implicit-private fields/locals/private members, so pool hover/def on a field or local resolved to nothing. `DispatchHover` + `DispatchDefinition` now `recompileCursorFileAtFullDetail(req.content)` after enrichment load.

**3. `DispatchDefinition` wire schema was missing `content`** (`286b737ac`). Without it `req.content` was always `undefined` → the full-detail recompile returned early → definition on members returned `[]`. Added `content` to the schema and populated it in `buildLspRequestMessage`.

**4. Cross-file supertype dep-fetch gap** (`286b737ac`). The Phase-2 dependency prefetch only collected `CLASS_REFERENCE`/`CONSTRUCTOR_CALL`/`TYPE_DECLARATION` and only when unresolved. `implements`/`extends` refs are `INTERFACE_IMPLEMENTATION`/`INHERITANCE` and come back already-resolved, so the supertype's declaring file was never ingested into the pool worker. Now fetches supertype refs even when resolved (the declaring file's table must be present locally for goto-def to yield a location).

**5. Pool blocking missing-artifact resolution was a no-op** (`a333b35c0`). `resolveBlocking` always submitted to the pool worker's own queue (no client connection → no-op). Now routes through the awaited assistance proxy when there's no local connection; paired with `setAssistanceProxy` blocking so the coordinator mediator awaits its handler instead of returning `{accepted:true}` prematurely. Makes cross-file inherited-method goto pass.

**6. Missing-artifact finder re-opened the origin file** (`753e17368`). `resolveFromWorkspace` broke on the first strategy returning any file, and the `parentContext.containingType` strategy targets the enclosing class (= the origin file, always present) at higher confidence than the correct `**/Target.cls` strategies — so it re-opened the already-open origin (a no-op for a *missing* artifact) and never reached the target. Now filters the origin file out of each strategy's matches before the first-hit short-circuit. Web goto-def 25/29 → 28/29; all cross-file interface/base/inherited cases pass.

**7. Field go-to-definition opened a peek instead of jumping** (`4cc8c126c`). Two duplicate sources fed `DefinitionProcessingService`'s duplicate-declaration grouping: (a) a field and a same-named constructor parameter collapse onto one `unifiedId` because the constructor's scope name equals the enclosing class name — now also matched on `kind` (a real duplicate declaration is always the same kind; a field and a parameter never are); (b) layered public-api + full recompile can leave two same-kind field entries on the same line (one anchored to the identifier, one to the type token) — now collapsed by `(uri, line)`, since a genuine duplicate declaration lives on its own line. Added `DefinitionFieldParamDuplicate.integration.test.ts` covering both. Web goto-def 28/29 → 29/29.

**Test/harness fix** (`d83cd23f9`): find-references web e2e was 1/4 due entirely to the harness (the LSP resolved references correctly). `findReferences()` waited on the go-to-definition peek selector instead of the find-all-references `reference-zone-widget`; the result tree is virtualized so row counting undercounted (now parses the authoritative total from the peek title); and the tests targeted minimal same-file symbols that VS Code Web won't open a peek for. Retargeted to a field with multiple same-file references; deterministic 4/4.

@W-23408848@

### DIFF
--- a/e2e-tests/pages/ApexEditorPage.ts
+++ b/e2e-tests/pages/ApexEditorPage.ts
@@ -199,7 +199,7 @@ export class ApexEditorPage extends BasePage {
     await this.page.keyboard.press('Shift+F12');
 
     // Find-all-references renders in the `reference-zone-widget` peek — a
-    // DIFFERENT DOM node than the go-to-definition peek
+    // different DOM node than the go-to-definition peek
     // (`.editor-widget.peekview-widget`). Match the references widget
     // specifically so we don't miss it and report 0.
     const peekWidget = this.page.locator(
@@ -213,10 +213,10 @@ export class ApexEditorPage extends BasePage {
       return 0;
     }
 
-    // The peek's result tree is VIRTUALIZED: only the on-screen rows exist in
-    // the DOM (a 5-result peek may render just 2-3 `.monaco-list-row`s), so
-    // counting rows undercounts. The peek title carries the authoritative
-    // total — "References (5)" — so parse that first.
+    // The peek's result tree is virtualized: only on-screen rows exist in the
+    // DOM (a 5-result peek may render just 2-3 `.monaco-list-row`s), so counting
+    // rows undercounts. The peek title carries the authoritative total —
+    // "References (5)" — so parse that first.
     const titleText = await peekWidget
       .locator('.head .peekview-title .meta')
       .first()
--- a/e2e-tests/pages/ApexEditorPage.ts
+++ b/e2e-tests/pages/ApexEditorPage.ts
@@ -227,6 +227,15 @@ export class ApexEditorPage extends BasePage {
       return Number(match[1]);
     }
 
+    // Locale fallback: on non-English VS Code the word "References" is
+    // translated, so the word-anchored regex above misses. The count is still
+    // rendered in parentheses regardless of locale, so extract the last
+    // parenthesized integer from the title before resorting to row counting.
+    const localeMatch = titleText?.match(/\((\d+)\)\s*$/);
+    if (localeMatch) {
+      return Number(localeMatch[1]);
+    }
+
     // Fallback: no parseable title (older VS Code, or a single-result peek that
     // navigates instead of showing a count) — count the rendered result rows.
     const entries = peekWidget.locator(
--- a/e2e-tests/pages/ApexEditorPage.ts
+++ b/e2e-tests/pages/ApexEditorPage.ts
@@ -198,9 +198,13 @@ export class ApexEditorPage extends BasePage {
 
     await this.page.keyboard.press('Shift+F12');
 
-    // The references peek widget appears for any non-empty result set. Give the
-    // LSP time to answer (references can fan out cross-file).
-    const peekWidget = this.page.locator('.editor-widget.peekview-widget');
+    // Find-all-references renders in the `reference-zone-widget` peek — a
+    // DIFFERENT DOM node than the go-to-definition peek
+    // (`.editor-widget.peekview-widget`). Match the references widget
+    // specifically so we don't miss it and report 0.
+    const peekWidget = this.page.locator(
+      '.peekview-widget.reference-zone-widget',
+    );
     await peekWidget
       .waitFor({ state: 'visible', timeout: this.defaultTimeout })
       .catch(() => {});
@@ -209,8 +213,22 @@ export class ApexEditorPage extends BasePage {
       return 0;
     }
 
-    // Each reference is a row in the peek's tree. Count the result entries
-    // (file/line rows), excluding the file-group headers.
+    // The peek's result tree is VIRTUALIZED: only the on-screen rows exist in
+    // the DOM (a 5-result peek may render just 2-3 `.monaco-list-row`s), so
+    // counting rows undercounts. The peek title carries the authoritative
+    // total — "References (5)" — so parse that first.
+    const titleText = await peekWidget
+      .locator('.head .peekview-title .meta')
+      .first()
+      .textContent()
+      .catch(() => null);
+    const match = titleText?.match(/References?\s*\((\d+)\)/i);
+    if (match) {
+      return Number(match[1]);
+    }
+
+    // Fallback: no parseable title (older VS Code, or a single-result peek that
+    // navigates instead of showing a count) — count the rendered result rows.
     const entries = peekWidget.locator(
       '.monaco-list-row .referenceMatch, .monaco-list-row[role="treeitem"]',
     );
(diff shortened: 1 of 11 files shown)


peternhale (changes requested):
## Code Review Findings

Reviewed the web worker topology and resolution fixes. Core logic is sound but there are critical issues in the new blocking proxy path that should be fixed before merge.

### MEDIUM-HIGH Priority (Blocking)

**1. Unsanitized params in blocking proxy path can cause DataCloneError**

Location: `MissingArtifactResolutionService.ts` line ~168 (new proxy branch)

The blocking proxy branch forwards raw `params`:
```typescript
const proxyResult = await proxy(params);   // raw params - UNSAFE
```

But `resolveInBackground` (same file, ~line 255) explicitly sanitizes before calling the same proxy:
```typescript
// Symbol manager class instances (typeReference, parentContext.*) are not cloneable
const safeParams = { ...params, identifiers: params.identifiers.map(id => decodeIdentifier(id) ...) };
proxy(safeParams)
```

Both terminate in `port.postMessage({... params ...})` which uses structured clone. Non-cloneable symbol manager instances will throw `DataCloneError`, causing silent resolution failure.

**Fix:** Hoist `safeParams` construction above the branch and use it in both the proxy call and queue `submitRequest`. Single sanitized payload for all sinks.

**2. No timeout on awaited proxy - can hang indefinitely**

Location: `MissingArtifactResolutionService.ts` (proxy branch) + `worker.platform.ts`

The proxy branch does bare `await proxy(params)` with no timeout. The underlying `requestCoordinatorAssistance` has no self-imposed rejection. If the coordinator errors or response is lost, hover/definition requests hang forever on the hot path.

**Fix:** Wrap proxy await in `Promise.race` with `blockingWaitTimeoutMs` (same budget the queue path uses) and populate cooldown map on timeout.

### MEDIUM Priority

**3. Unconditional recompile on all platforms**

Location: `worker.platform.shared.ts` - `DispatchHover` (~1700), `DispatchDefinition` (~1845)

Both handlers unconditionally call `recompileCursorFileAtFullDetail` on every request, including desktop where data-owner may already have full detail. The PR comment explains why gating on `detailLevel` is unsafe and defers a "local full-detail marker" solution.

**Action:** Track the deferred optimization as a concrete follow-up work item (the comment mentions W-23408848).

### LOW Priority

**4. Timeout cooldown guard is dead code**

`mapResultToBlockingResult` only returns `'resolved'` or `'not-found'`, never `'timeout'`. The `if (mappedProxy !== 'timeout')` check is always true, so proxy path never records cooldown. Pre-existing in queue path too.

**5. Line-collapse dedup is unconditional**

`DefinitionProcessingService.ts` (~356) - the `(uri, line)` dedup runs on all results, not just detected duplicates. This is documented and has integration test coverage, but reviewer should be aware the coarse key is unconditional.

**6. E2E test peek title parsing is English-only**

`ApexEditorPage.ts` `findReferences()` parses `/References?\s*\((\d+)\)/i` from peek title. On localized VS Code this won't match and falls back to counting rows (undercounts due to virtualization).

---

### Items Verified as Correct ✓

- Worker URL `blob:` prefix guard and fallback chain
- `setAssistanceProxy(..., true)` blocking flag change
- Supertype ref always-fetch filter for INHERITANCE/INTERFACE_IMPLEMENTATION
- Kind-matching addition to duplicate filter (field vs constructor param)
- `DefinitionFieldParamDuplicate.integration.test.ts` coverage

---

**Critical fixes needed:** #1 (sanitize params), #2 (add timeout)

These directly affect the features this PR adds and can cause silent failures or hangs.

**Files reviewed:**
- `packages/apex-ls/src/server/LCSAdapter.ts`
- `packages/apex-ls/src/server/WorkerCoordinator.ts`
- `packages/apex-ls/src/worker.platform.shared.ts`
- `packages/apex-lsp-shared/src/workerWireSchemas.ts`
- `packages/apex-lsp-vscode-extension/src/missing-artifact-handler.ts`
- `packages/lsp-compliant-services/src/services/DefinitionProcessingService.ts`
- `packages/lsp-compliant-services/src/services/MissingArtifactResolutionService.ts`
- `e2e-tests/pages/ApexEditorPage.ts`

kylewalke:
Thanks for the review. Addressed in `250f0997`:

**#1 — Unsanitized params in blocking proxy (fixed).** Hoisted a single `sanitizeParams()` helper (the wire-schema decode `resolveInBackground` already used) above the branch and now forward the sanitized payload from **both** sinks — the proxy call *and* the queue `submitRequest`. One structured-clone-safe payload for every sink, no more `DataCloneError` risk on the proxy path.

**#2 — No timeout on awaited proxy (fixed).** The proxy await is now raced against the same `blockingWaitTimeoutMs` budget the queue path uses (new `withTimeout()` helper). On timeout it throws with a `timeout`-matching message, which the existing catch handles — records the cooldown and returns `'timeout'`. No more indefinite hang on a lost/errored coordinator response.

**#4 — Timeout cooldown guard was dead code (fixed).** With #2 in place, proxy timeouts now surface as thrown errors, so the `recentBlockingTimeouts` map is genuinely populated on the proxy path. Removed the always-true `mapped !== 'timeout'` guards (`mapResultToBlockingResult` only yields `resolved`/`not-found`) and clear the stale cooldown unconditionally on success.

**#6 — E2E peek-title parse was English-only (fixed).** Added a locale-agnostic fallback in `findReferences()` that extracts the trailing parenthesized count (`(\d+)$`) before falling through to row counting, so localized VS Code still gets an accurate total.

**Tests:** added `MissingArtifactResolutionService-blocking-proxy.test.ts` covering (a) proxy param sanitization is `structuredClone`-safe and strips class methods, and (b) a never-settling proxy returns `'timeout'`.

**#3 (M1, deferred) / #5 (L5, deferred):** both already documented with rationale in `4743ee28`. #3's `detailLevel` gate is unsafe (the level reflects the data-owner's table, not what the local pool member holds — verified: gating regressed 3 field/var web e2e cases); a correct gate needs a local full-detail marker. #5's `(uri,line)` dedup is intentionally coarse because adding column would reintroduce the peek bug (layered identifier/type-token artifacts differ only by column). Both tracked under W-23408848.

peternhale (approved):
Excellent work on the web worker topology fixes. The root cause analysis is exemplary, with clear repro/diagnosis for all 7 bugs. The duplicate-declaration fixes (field vs. parameter + layered compilation artifacts) are well-tested, and the assistance proxy blocking + supertype prefetch changes correctly handle the cross-file resolution cases. All checks passing. ✅

PR #552, opened 2026-07-13
feat(apex-ls): add distributed tracing with Effect-based enrichment spans - W-23354947

peternhale:
## What does this PR do?

Implements end-to-end distributed tracing infrastructure for the Apex Language Server, enabling trace context propagation from coordinator through workers and fixing orphaned enrichment spans by converting them to Effect-based execution.

@W-23354947@

### Distributed Tracing Infrastructure
- Added OpenTelemetry context managers (AsyncLocalStorageContextManager) to coordinator and workers
- Configured W3C TraceContextPropagator for traceparent header injection/extraction
- Created span collector service in extension to forward worker spans to OTLP endpoint
- Implemented trace context flow: coordinator → worker message injection → worker extraction → unified trace view

### Effect-Based Enrichment Spans
- Added `effectRequestHandler` variant for handlers calling Effect-returning helpers
- Converted `loadSymbolDataForEnrichment` and `writeBackEnrichedSymbols` from Promise to Effect
- Removed `Effect.runPromise` boundaries that broke parent span context
- Converted 7 enrichment-using handlers to `effectRequestHandler` pattern (DispatchHover, DispatchCompletion, DispatchSignatureHelp, DispatchCodeAction, DispatchDefinition, DispatchImplementation, DispatchDocumentSymbol, DispatchCodeLens, DispatchDiagnostic, DispatchCrossFileEnrichment)

### Instrumented Operations
Created 6 enrichment spans now properly nested under LSP request spans:
- `worker.enrichment.querySymbolSubset` - IPC call to data-owner for symbol tables
- `worker.enrichment.addSymbolTable` - Local symbol table ingestion
- `worker.enrichment.resolveDepUris` - Resolve class dependencies via data-owner
- `worker.enrichment.resolveCrossFileReferences` - Materialize cross-file edges
- `worker.enrichment.querySymbolByName` - Cross-worker symbol resolution fallback
- `worker.enrichment.updateSymbolSubset` - Write back enriched symbols to data-owner

## Testing

- All 4646 tests pass (0 failed, 490 pending)
- Unit tests verify trace context extraction and span parenting
- Live trace verification via Grafana Tempo confirms enrichment spans correctly nested under `worker.lspRequest.*` spans

## Before

Traces showed only 2 levels:
```
lsp.coordinator.hover
└── worker.lspRequest.DispatchHover
```

Enrichment spans were created but orphaned in separate traces.

## After

Traces now show 4-5 levels with enrichment operations visible:
```
lsp.coordinator.codeLens
└── worker.lspRequest.DispatchCodeLens
    ├── worker.enrichment.querySymbolSubset
    ├── worker.enrichment.addSymbolTable
    ├── worker.enrichment.resolveDepUris
    └── worker.enrichment.resolveCrossFileReferences
```

All spans share the same trace ID with correct parent-child relationships.


packages/apex-ls/esbuild.config.ts
```diff
@@ -129,7 +157,7 @@ const builds: BuildOptions[] = [
     sourcemap: true,
     minify: false, // DEBUG: keep unminified for stack trace readability
     metafile: true,
-    external: [],
+    external: WORKER_EXTERNAL, // Mark OTEL packages external even for browser (dead code)
```
kylewalke, line 159: 🔴 **This is the web-E2E blocker.** Marking node-only OTEL pkgs `external` in an **IIFE browser** build does not remove them — esbuild emits a runtime `__require("@opentelemetry/...")` at module-init. Since `withExtractedTraceContext` (statically re-exported from the shared barrel and called by every handler) is live, it isn't tree-shaken, so the `__require` survives into `worker.platform.web`/`server.web`. A browser Web Worker has no `require`, so esbuild's `__require` fallback throws `Error: Dynamic require of "@opentelemetry/context-async-hooks" is not supported` at load — the worker never boots → empty hover → all web E2E red. Reproduced locally. See review body for fix options (browser-field stub / node-gated dynamic import / split the barrel).

packages/apex-ls/esbuild.config.ts
```diff
@@ -67,6 +78,23 @@ const WORKER_EXTERNAL: string[] = [
   '@effect/platform-node/NodeWorker',
   // NOTE: @effect/platform/WorkerError must NOT be external — it is a
   // dependency of @effect/platform-browser/BrowserWorker which runs in browser.
+
+  // Coordinator tracing (Node.js only) - dynamically imported, needs to be external
+  './coordinatorTracing',
+  // Worker and coordinator tracing OTEL packages (Node.js only) - these use Node.js
+  // built-ins like async_hooks and should never be bundled into browser builds.
+  // They're imported by apex-lsp-shared which gets bundled, so mark them external
+  // even for browser builds (dead code elimination will remove their usage).
```
kylewalke, line 87: This assumption is the bug: the OTEL usage is **not** dead. `withExtractedTraceContext` is re-exported by `apex-lsp-shared/src/index.ts` and called by every request handler, so DCE cannot remove it, and `external` in an IIFE turns it into a load-time `__require` that throws in the browser worker.

packages/apex-lsp-shared/src/observability/workerTracing.ts
```diff
@@ -0,0 +1,216 @@
...
+ */
+
+import * as OtlpTracer from '@effect/opentelemetry/OtlpTracer';
+import { layerJson as OtlpSerializationLayer } from '@effect/opentelemetry/OtlpSerialization';
+import * as EffectTracer from '@effect/opentelemetry/Tracer';
+import { layer as FetchHttpClientLayer } from '@effect/platform/FetchHttpClient';
+import * as Effect from 'effect/Effect';
+import * as Layer from 'effect/Layer';
+import { context, propagation } from '@opentelemetry/api';
+import { AsyncLocalStorageContextManager } from '@opentelemetry/context-async-hooks';
```
kylewalke, line 23: These top-level imports of node-only packages (`context-async-hooks` → `async_hooks`, `@opentelemetry/core`, and `@effect/opentelemetry` above) execute at module-init. Because this module is reachable from the web worker via the `index.ts` barrel + `withExtractedTraceContext`, they end up as load-time `__require(...)` in the IIFE and crash the browser worker. Consider moving them into `initWorkerTracing` behind a `process.versions?.node` guard via `await import(...)`, or provide a browser stub for this module.

packages/apex-lsp-shared/src/observability/index.ts
```diff
@@ -42,3 +43,12 @@ export { CommandPerformanceAggregator } from './commandPerformanceAggregator';
...
 export { collectStartupSnapshot, generateSessionId } from './startupSnapshot';
 export type { StartupSnapshotParams } from './startupSnapshot';
+
+export {
+  initWorkerTracing,
+  shutdownWorkerTracing,
+  getWorkerTracerLayer,
+  getCollectorUrl,
+  provideWorkerTracing,
+  withExtractedTraceContext,
```
kylewalke, line 53: This static re-export is what pulls `workerTracing.ts` (and its node-only OTEL top-level imports) into the web worker bundle. Splitting the node-only tracing symbols out of the barrel the web worker imports would break the chain that causes the load-time crash.

packages/apex-ls/src/worker.platform.shared.ts
```diff
@@ -562,145 +600,171 @@ export type CodeActionReq = {
...
-      getText: () => content,
-      languageId: 'apex',
-      version: 0,
-    };
-    svc.storageManager.getStorage().setDocument(uri, doc as never);
-  }
-
-  let version = -1;
-  let detailLevel = 'public-api';
+): Effect.Effect<{ version: number; detailLevel: string }, never, never> {
```
kylewalke, line 684: 🔴 The `never` error channel here is inaccurate. The body still uses JS `try/catch` but now does `yield* Effect.promise(() => ...)`. On promise **rejection**, `Effect.promise` produces an unrecoverable **defect** that short-circuits the generator — it does not throw at `yield*`, so the outer `catch (err)` (the “subset load failed → warn, proceed on partial graph” fallback) never runs and the whole request dies. Same pattern in `writeBackEnrichedSymbols`. Use `Effect.tryPromise` + `Effect.catchAll`/`Effect.either` for the fallback. Needs a failure-path test — `EnrichmentRoundTrip` only covers the happy path.

packages/apex-ls/src/worker.platform.shared.ts
```diff
@@ -2141,11 +2247,26 @@ export const handlers: WorkerRunner.SerializedRunner.Handlers<
...
       yield* Effect.logInfo(
         `[worker] role=${req.role} protocol=v${req.protocolVersion}/${WIRE_PROTOCOL_VERSION}` +
-          ` logLevel=${currentWorkerLogLevel}`,
+          ` logLevel=${currentWorkerLogLevel}` +
+          (req.spanCollectorUrl ? ' tracing=enabled' : ''),
       );
+
+      // TEST: Create a test span to verify Effect tracing is working
+      if (req.spanCollectorUrl) {
+        yield* Effect.fn('worker.init.test.span')(() =>
```
kylewalke: Leftover TEST scaffolding (`worker.init.test.span`, `// TEST: ...`) in the production shared module. Gated by `spanCollectorUrl` so harmless, but shouldn't ship.

packages/apex-ls/src/server/traceContextInjection.ts
```diff
@@ -0,0 +1,102 @@
...
+          `[traceContext] Injecting trace context into ${(payload as any)._tag}: ${carrier.traceparent}`,
+        );
+        return { ...payload, traceContext: carrier.traceparent };
+      }
+    }
+  } catch (error) {
+    console.log(`[traceContext] Failed to inject trace context: ${error}`);
+  }
+
+  console.log(
```
kylewalke: 🟡 This `console.log` (and the others in this file) fires on **every** coordinator→worker dispatch — including when tracing is disabled, via this “No active OTEL span” branch. Per-request stdout spam on the hot path. Gate behind `isTracingEnabled()` or a debug logger, and confirm it can't reach an stdio LSP channel.

kylewalke (changes requested):
## Review: distributed tracing + Effect-based enrichment

The tracing design is solid — clean infra modules, init properly gated behind config/env, and `runWithSpan` honors the "tracing off = no overhead" contract. But there are two blocking problems, one confirmed with a local repro.

### 🔴 Blocker 1 — Web worker bundle crashes at load (all web E2E red)

CI shows **0/11 web E2E passing** (hover/outline/goto/lsp-integration/extension-core), consistent across all 3 retries — not flaky. Root cause traced end-to-end and reproduced:

1. `apex-lsp-shared/src/index.ts` now **statically** re-exports `withExtractedTraceContext`/`initWorkerTracing` from `observability/workerTracing.ts`.
2. `workerTracing.ts` has **top-level** imports of node-only pkgs: `@opentelemetry/context-async-hooks` (→ `async_hooks`), `@opentelemetry/core`, `@effect/opentelemetry`.
3. `withExtractedTraceContext` is called by **every** request handler (`requestHandler`/`effectRequestHandler`) → live code, **not tree-shakeable**.
4. The web IIFE build marks those pkgs `external` with the comment “dead code elimination will remove their usage” — but it isn't dead, so it isn't removed.
5. In an IIFE browser bundle, `external` leaves a runtime `__require("@opentelemetry/...")` at module-init. There's a `process` banner shim but **no `require` shim**, so it throws immediately.

**Repro** (minimal esbuild `--format=iife --platform=browser --external:@opentelemetry/context-async-hooks` on a live-exported fn importing that pkg), executed with `require` undefined as in a browser Web Worker:

```
var import_context_async_hooks = __require("@opentelemetry/context-async-hooks");
// → Error: Dynamic require of "@opentelemetry/context-async-hooks" is not supported
```

The throw is at the top of the IIFE, so the worker dies before handling any request — exactly matching the logs (“LCS readiness completed”, then empty hover). **This is introduced by this PR:** the same base commit (`5d5e4b84`) is green on `main` for all web E2E jobs.

Fix options: (a) add a `browser` field in `apex-lsp-shared/package.json` mapping `workerTracing` to a no-op stub; (b) move the OTEL imports out of module top-level into `initWorkerTracing`, node-gated via `await import(...)` behind `process.versions?.node`; (c) split the node-only tracing symbols out of the barrel the web worker imports. Acceptance: web E2E green **and** the `worker.platform.web` IIFE contains no `__require(...)` of any OTEL/`async_hooks` module.

### 🔴 Blocker 2 — Promise→Effect conversion defeats the graceful-degradation `try/catch`

`loadSymbolDataForEnrichment` and `writeBackEnrichedSymbols` were converted to `Effect.gen` but kept their original JS `try/catch` blocks while swapping `await promise` → `yield* Effect.promise(() => promise)`. When the promise **rejects**, `Effect.promise` turns it into an unrecoverable **defect** that short-circuits the generator — it does not throw at the `yield*` site, so the surrounding JS `catch (err)` never runs. The deliberate “assistance channel down / data-owner missing the file → warn and proceed on partial graph” fallback is now dead on the reject path, and the whole request Effect dies instead. The `Effect<…, never, never>` type is inaccurate — `Effect.promise` can die. Use `Effect.tryPromise` + `Effect.catchAll`/`Effect.either` (or `Effect.catchAllDefect`); don't rely on JS `try/catch` around `yield*`. Currently untested — `EnrichmentRoundTrip` only covers the happy path.

### 🟡 Medium — hot-path `console.log` in trace-context injection

`injectTraceContext*` runs on every coordinator→worker dispatch and `console.log`s unconditionally, **including when tracing is disabled** (the “No active OTEL span” branch). Per-request stdout spam on the hot path; also worth confirming it can't reach an stdio LSP channel and corrupt the protocol. Gate behind `isTracingEnabled()` / a debug logger.

### 🟡 Medium — leftover TEST scaffolding in the production shared module

`worker.platform.shared.ts` emits a span literally named `worker.init.test.span` (`// TEST: ...`). Gated by `spanCollectorUrl` so harmless, but shouldn't ship.

### 🟢 Low / nits
- `shutdownWorkerTracing` is `async` but does no async work and never flushes/closes the OtlpTracer layer (coordinator side does `tracerProvider.shutdown()`) — pending spans may be lost.
- PR body says “7 enrichment-using handlers” but lists/converts 10.
- The 25 prettier-only validator reformats + 7k-line snapshot regeneration bury the real diff; not blocking, but consider splitting.

### ℹ️ Informational — relationship to #549
I checked whether #549 (`fix/web-pool-resolution-W-23408848`) affects any of this. It does **not**: #549 is open (not in `main`), and this PR branched from a commit where web E2E was already green **without** #549. None of this PR's crash-surface files (`esbuild.config.ts`, the shared barrel, `workerTracing.ts`) are touched by #549 — Blocker 1 is a bundling defect #549 can't fix. #549's fixes are runtime-resolution gaps that only matter *after* the worker loads, so they're orthogonal. **However**, the two PRs overlap on 6 files (incl. `worker.platform.shared.ts`, `WorkerCoordinator.ts`, `LCSAdapter.ts`, `workerWireSchemas.ts`) — and this PR's Effect conversion touches the same enrichment functions #549 modifies for resolution — so whichever merges second will hit a non-trivial conflict worth coordinating.

---
The tracing work itself is good; the blockers are a browser-bundling boundary violation and an Effect error-channel regression, both contained fixes.

PR #553, opened 2026-07-14
feat(apex-ls): trace workspace loading across workers - W-23354947

peternhale:
### What does this PR do?

- Adds end-to-end OpenTelemetry spans for workspace loading and LSP processing across the extension host, coordinator, and independent worker processes.
- Preserves W3C parent context through the priority queue, worker assistance, resource loading, compilation, data-owner, and request-worker paths.
- Adds deeper workspace batch, ingestion, compilation, enrichment, and symbol-resolution spans for performance diagnosis.
- Routes collected coordinator and worker spans through per-service Salesforce SDK runtimes so every configured SDK exporter receives them, including Tempo, console, and filesystem sinks.
- Replays collected spans parent-first and remaps SDK-generated span IDs so cross-process parent-child relationships remain intact.
- Keeps desktop-only collector code browser-safe and updates dynamic definition and implementation registration behavior.

@W-23354947@

### Functionality Before

Workspace-load traces were shallow or incomplete, and context could be lost at queue and worker boundaries. The local collector forwarded raw OTLP only to the configured local endpoint, bypassing other Salesforce SDK exporters. As a result, Apex LS coordinator and worker spans did not appear in the SDK filesystem capture.

### Functionality After

Workspace loads and LSP requests produce connected span trees across coordinator and worker services. Collected spans are exported through the Salesforce SDK with their original service identities. Manual verification recorded 861 Apex LS spans across the coordinator and all four worker roles in the SDK JSONL file, with no missing parent references or error-status spans. Tempo also receives the traces.

### Validation

- Repository hooks: 369 test suites
- 4,671 tests passed, 0 failed, 489 pending
- Extension compile, lint, production bundle, and formatting checks passed
- Manual Tempo and SDK filesystem export verified


packages/apex-lsp-shared/src/observability/workerTracing.browser.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Browser workers do not run the desktop span collector and cannot load the
+ * Node-only OTEL context manager. Keeping the same API as the Node module lets
+ * the shared barrel remain platform-neutral while the package browser mapping
+ * removes every async_hooks/OTEL exporter import from web bundles.
+ */
+
+import type * as Effect from 'effect/Effect';
+import type * as Layer from 'effect/Layer';
+
+export function initWorkerTracing(_url: string, _serviceName: string): void {}
```
kylewalke, line 21: ✅ This browser stub (plus the package `browser` field / esbuild `alias` redirect and the DI `workerTracingHooks` in `worker.platform.shared.ts`) is exactly the fix for the #552 🔴 `__require` blocker. Because the shared worker module no longer statically imports the node OTEL/async_hooks symbols, nothing pulls them into the IIFE web bundle, so there's no load-time `__require` throw. Verified: `apex-capability-gating.spec.ts (Web)` now passes, i.e. the worker boots. Nicely done.

packages/apex-ls/src/worker.platform.shared.ts
```diff
@@ -574,11 +652,46 @@ export const requestHandler =
...
+      ),
+    );
+
+/**
+ * Effect-returning request handler variant for handlers that need to yield*
+ * Effect-returning helpers (e.g. loadSymbolDataForEnrichment after Part B
+ * conversion). Identical to requestHandler except callService returns an Effect
+ * and is yielded directly instead of wrapped in Effect.promise.
+ */
+export const effectRequestHandler =
```
kylewalke, line 674: 🔴 **Blocker (Web E2E).** This Promise→Effect handler conversion is the prime suspect for the web hover/outline/definition regression: those suites are green on `main` but red here, all returning **empty** content (worker boots — capability-gating Web passes — so it's not the old `__require` crash). `effectRequestHandler` yields `callService` directly and the enrichment helpers wrap service calls in `Effect.promise` / `Effect.tryPromise` + `catchAll`. On the web path, if any of those rejects (→ defect) or a `catchAll` resolves to an empty/undefined result, the handler returns empty hover instead of surfacing the error. Please root-cause the empty-result path in the web build specifically (e.g. add a temporary web trace/log around `loadSymbolDataForEnrichment` → `processHover`) and confirm the Web E2E suites go green before merge. The #552 fallback fix + new failure-path tests are good; this is a separate web-specific empty-result path.

packages/apex-lsp-vscode-extension/src/extension.ts
```diff
@@ -187,29 +179,43 @@ export function activate(context: vscode.ExtensionContext): void {
...
+        `Extension tracing init error: ${formattedError(error)}`,
+        'warning',
       );
+    })
+    .finally(() => {
+      // Start the language server after tracing initialization completes (or fails)
+      const { getClient } = require('./language-server');
+      const existingClient = getClient();
+      if (existingClient) {
+        console.log('⚠️ Client already exists, skipping start');
```
kylewalke, line 199: 🟡 Debug `console.log` scaffolding — this and the `'🔧 About to start language server…'` log a few lines down duplicate the `logToOutputChannel(...)` calls immediately beside them. Drop the emoji `console.log`s (keep the output-channel logging).

kylewalke (changes requested):
## Review — #553 (successor to #552, W-23354947)

I reviewed this as the follow-up to the now-closed #552, checking specifically whether the 7 findings I raised there were resolved. **Six of seven are cleanly addressed — including both 🔴 blockers — and the fixes are genuinely good.** But CI shows a **new web regression** that blocks merge.

### ✅ #552 findings resolved (verified, not just present)

1. **🔴 Web `__require` crash (the #552 blocker) — FIXED, and well-architected.** The node-only `workerTracing.ts` now has a browser twin (`workerTracing.browser.ts`) with stubbed exports; `worker.platform.shared.ts` no longer statically imports OTEL/async_hooks and instead takes injected `workerTracingHooks` (default no-ops), with node installing the real hooks via `setWorkerTracingHooks` and web importing only the `.browser` stub. The observability barrel no longer re-exports the node tracing symbols. This is the "split the barrel + browser stub" fix I suggested. **Confirmed the worker now boots on web: `apex-capability-gating.spec.ts (Web)` passes**, which it could not if the load-time `__require` throw were still present.

2. **🔴 `Effect.promise` defect swallowing the enrichment fallback — FIXED with a real test.** `loadSymbolDataForEnrichment` / `writeBackEnrichedSymbols` now use `Effect.tryPromise` + `.pipe(Effect.catchAll(...))`, so a rejected assistance call becomes a typed failure the fallback catches (partial-graph degrade / `return false`) rather than an unrecoverable defect that kills the request. Crucially, the failure-path test I asked for was added (`WorkerPlatformShared.test.ts:84,98`) — I ran it locally, 9/9 green, and the `[ENRICHMENT] Write-back failed` warning fires exactly on the rejection path. This is the right fix.

3. **Leftover TEST scaffolding (`worker.init.test.span`) — removed.**
4. **`console.log` in `traceContextInjection.ts` on the dispatch hot path — removed** (and that file is now defensively try/catch'd and only imports browser-safe `@opentelemetry/api`).

### 🔴 Blocker: Web E2E regressed by this PR
`main` is **green** across all Web E2E suites; on this branch **hover, hover-stdlib, outline, goto-definition, lsp-integration, extension-core (Web) all fail** with empty language-feature content (e.g. `expect(content).toContain('ApexClassExample')` / `'String'` — hover comes back empty). All **non-web** checks (Test matrix, Package, code-quality) pass. Since capability-gating (Web) passes, the worker *boots* — this is not the old `__require` crash — but web symbol/hover resolution now returns nothing. The branch already contains all of `main`, so this is introduced by this PR's own changes (prime suspect: the ~1200-line Promise→Effect refactor in `worker.platform.shared.ts`, since the `Effect.promise` / `Effect.tryPromise` conversion changes error/empty semantics on the request path — an `Effect.promise` rejection becomes a defect, and any `catchAll` that resolves to an empty/undefined result would surface as empty hover). This must be root-caused and fixed (or the offending change reverted/deferred) before merge — shipping it red reintroduces the exact web-broken state #552 was trying to escape. See inline note.

### 🟡 Non-blocking cleanups
- **Debug `console.log('🔧 About to start language server…')` / `'⚠️ Client already exists…'`** in `extension.ts` (added here) duplicate the `logToOutputChannel` calls right beside them — remove the emoji `console.log`s. See inline.
- **Four scratch docs at repo root** (`IMPLEMENTATION_SUMMARY.md`, `VERIFY_TRACING.md`, `WORKSPACE_LOAD_TESTING.md`, `WORKSPACE_LOAD_TRACING.md`, +743 lines) look like working notes — move under `docs/` or drop them.
- **Pre-existing (not this PR):** `worker.platform.shared.ts` still has an ungated `console.log(logMsg)` on the data-owner write-back path (was on `main` too, so out of scope here) — worth cleaning up in a follow-up since it's the same hot-path-stdout pattern from #552 finding 6.

Happy to re-approve once the Web E2E suites are green again. The tracing-safety work itself is solid.

packages/apex-ls/src/worker.platform.shared.ts
```diff
@@ -750,24 +877,29 @@ export async function loadSymbolDataForEnrichment(
...
-    // distinguishing "real failure" from "genuinely nothing here".
-    getLogger().warn(
-      () => `[ENRICHMENT] Symbol-subset load failed for ${uri}: ${err}`,
-    );
-  }
 
-  return { version, detailLevel };
+    return { version, detailLevel };
+  }).pipe(
+    Effect.catchAll((error) =>
```
kylewalke, line 891: 🔴 **Root cause of the Web E2E regression — pinned here.** This `Effect.catchAll` recovers only **typed failures (`E`)**, not **defects (dies)**. But the operations `yield*`-ed inside this generator that can fail on the web resolution path are `svc.symbolManager.addSymbolTable(...)` (lines ~799, 848) and `resolveCrossFileReferencesForFile(...)` (line ~883) — both typed `Effect.Effect<void>` (error channel `never`, see `symbolManagerFacade.ts:190,207`). With `E = never`, any runtime error they throw is necessarily a **defect**, which `catchAll` does **not** intercept.

On `main` the identical work was driven via `await Effect.runPromise(svc.symbolManager.addSymbolTable(...))` **inside a broad JS `try/catch`** — and `Effect.runPromise` surfaces *both* failures and defects as a thrown rejection, so that `catch` swallowed everything, warned, and returned the partial `{ version, detailLevel }` so hover proceeded. This PR's conversion removed that defect safety net: a defect now escapes this `catchAll`, propagates through `effectRequestHandler`, and **kills the whole DispatchHover Effect before `recompileCursorFileAtFullDetail` / `processHover` / `cloneForWire` run** → empty hover. Web-only because these manager/resolution ops die on the web path (remote ResourceLoader + serialized stdlib tables; cf. the `findSymbolByFQN` fallback this same PR adds in `symbolRefResolution.ts:718`), while node's local path succeeds — which is exactly why every non-web check is green.

Note the assist RPCs here (`Effect.tryPromise` at ~755/834/859) already convert rejections to typed failures this `catchAll` *does* catch — so the hole is specifically the **direct `yield*` manager Effects**.

**Fix:** change this to `Effect.catchAllCause` (or wrap those direct `yield*`s in `Effect.catchAllDefect`) to restore the old defect-catching behavior. The same pattern applies to `writeBackEnrichedSymbols`' `catchAll` at line ~1631. Quickest confirmation: swap this one line to `catchAllCause` and re-run the Web E2E suite — hover content should come back.

kylewalke (commented):
## Re-review — #553 (W-23354947)

Re-checked against HEAD (`180be6e`). **The merge blocker is resolved.** CI ran on the current commit and **all seven Web E2E suites are green** — hover, hover-stdlib, outline, goto-definition, lsp-integration, extension-core — along with the full Test matrix, Package, and code-quality. This is the exact web regression that drove the CHANGES_REQUESTED review, and it's gone.

### ✅ Prior blockers — all resolved
1. Web `__require` crash — fixed (browser stub + DI hooks).
2. `Effect.promise` defect swallowing the enrichment fallback — fixed (`tryPromise` + `catchAll`, with the failure-path test).
3. **Web E2E regression (empty hover/outline/definition) — resolved; all Web suites green on HEAD.**

**One note on the mechanism:** the fix did *not* land the way my inline comment suggested. Lines ~891 and ~1631 are still `Effect.catchAll` (not `catchAllCause`/`catchAllDefect`). The regression was instead cleared by the later commits — browser stubs for the OTEL modules (`f33d1f81`), desktop bundling (`f9e96f45`), and ESM tree-shaking conditions scoped to the extension (`635b3e0d` / `180be6e0`). The empty-hover symptom is gone regardless, so this clears the blocker — just flagging that the original defect-vs-typed-failure gap in those `catchAll`s wasn't the actual lever, in case it resurfaces later.

### 🟡 Non-blocking
- **Scratch docs at repo root — still present** and confirmed not on `main`, so this PR introduces them: `IMPLEMENTATION_SUMMARY.md`, `VERIFY_TRACING.md`, `WORKSPACE_LOAD_TESTING.md`, `WORKSPACE_LOAD_TRACING.md`. Please move under `docs/` or drop before merge — this is the one item the PR legitimately adds and leaves unaddressed.
- **Correction on my earlier note:** the emoji `console.log`s in `extension.ts` (`⚠️ Client already exists…` / `🔧 About to start language server…`) also exist on `main`, so they are **not** net-new to this PR. Out of scope here — worth a follow-up cleanup rather than a change request on #553.

Clearing my block on the substance — the tracing-safety work is solid and CI is green. Only ask before merge is to drop the four root-level scratch docs.

kylewalke (approved):

PR #555, opened 2026-07-14
feat(client): LanguageClientConnection adapter - W-23163195

peternhale:
## Summary

- Adds `LanguageClientConnection` adapter wrapping `vscode-languageclient` `BaseLanguageClient` to satisfy `RpcConnection` port
- Consolidates ~500 lines of duplicated boilerplate from extension's `ClientInterface` impls
- Includes unit tests and integration test exercising end-to-end LSP lifecycle with real server

## Plan

See [docs/plans/W-23163195.md](docs/plans/W-23163195.md)

## Reviewer notes

- Test uses `as unknown as BaseLanguageClient` double assertion (line 48). This is the standard partial-mock pattern used 100+ times across the codebase. Changing it requires a design decision about whether to create a full type-safe mock helper or test-only interface subset, which would span many files and change established patterns.
- Integration test (RUN_INTEGRATION=1) should be run manually to verify the adapter works end-to-end with a real server. This is a verification step, not a code change. Run: `npm run bundle -w @salesforce/apex-ls && RUN_INTEGRATION=1 npm test -w @salesforce/apex-lsp-client`

## Test plan

- `npm run typecheck` in `apex-lsp-client` — compiles without error
- `npm run test` in `apex-lsp-client` — new unit tests pass
- `npm run lint` in `apex-lsp-client` — no violations

@W-23163195@

---

🤖 Generated by auto-build pipeline. Original WI: https://gus.my.salesforce.com/a07EE00002d1LN4YAM


packages/apex-lsp-client/src/transports/languageClientConnection.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+import type { BaseLanguageClient, State } from 'vscode-languageclient';
+import type { Disposable } from '@salesforce/apex-lsp-shared';
+import type { RpcConnection } from '../rpcConnection';
+
+// State enum values from vscode-languageclient. Imported as type-only above
+// (vscode-languageclient requires `vscode` at module load, which is unavailable
+// in some host contexts during testing). Values: Stopped=1, Running=2, Starting=3.
+const STATE_STOPPED: State = 1 as State;
```
kylewalke, line 16: **(Low) Hardcoded enum value is fragile.** `1 as State` duplicates the library's `State.Stopped` because the type-only import can't read the value at runtime (the comment correctly explains the `vscode`-at-module-load constraint). Reasonable tradeoff — but if `vscode-languageclient` ever renumbers `State`, this silently misfires with **no compile error**. Consider noting the assumption is tied to the pinned `^9.0.1` in `package.json` so a version bump prompts a re-check. No clean runtime guard exists given the type-only requirement, so this is advisory.

packages/apex-lsp-client/src/transports/languageClientConnection.ts
```diff
@@ -0,0 +1,116 @@
...
+   * direct `onError` event, this synthesizes an error when the client
+   * transitions to `Stopped` unexpectedly (i.e., the previous state was not
+   * `Stopped` and the adapter is not being intentionally disposed).
+   */
+  onError(handler: (e: Error) => void): Disposable {
+    const disposable = this.client.onDidChangeState((event) => {
+      if (
+        event.newState === STATE_STOPPED &&
+        event.oldState !== STATE_STOPPED &&
+        !this.disposing
```
kylewalke, line 78: **(Low–Medium) This `!this.disposing` guard is never exercised by a test.** This flag is the whole point of commit `8df5129c` ("Fix spurious onError during graceful dispose"): a graceful `dispose()` → `client.stop()` → Running→Stopped transition must NOT fire `onError`. But no unit test calls `dispose()` before simulating the transition, so this branch is only ever hit with `disposing === false`. The logic is correct; the regression it fixes is untested. See my comment on the test file's `dispose` block for the suggested ~10-line addition.

packages/apex-lsp-client/src/transports/languageClientConnection.ts
```diff
@@ -0,0 +1,116 @@
...
+      }
+    });
+    return disposable;
+  }
+
+  /**
+   * Subscribe to transport close. Fires when the client transitions to
+   * `Stopped` from any other state.
+   */
+  onClose(handler: () => void): Disposable {
```
kylewalke, line 90: **(Observation, not a defect)** Asymmetry worth flagging for the follow-up WI: `onClose` fires on a graceful `dispose()` too (a dispose *is* a close), while `onError` suppresses it via `disposing`. Intentional and documented — but whoever wires `onClose`/`onError` in the consolidation WI should know `onClose` is **not** dispose-suppressed, to avoid double-teardown. No consumer registers either handler yet, so no risk today.

packages/apex-lsp-client/test/transports/languageClientConnection.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    it('returns false when client is not running', () => {
+      mockClient.isRunning.mockReturnValue(false);
+
+      expect(adapter.isListening()).toBe(false);
+    });
+  });
+
+  describe('dispose', () => {
+    it('calls stop() on the underlying client', async () => {
```
kylewalke, line 229: **(Low–Medium) Suggest adding a test for the graceful-dispose path** (pairs with the `disposing` guard in the source). Something like: register an `onError` handler, `await adapter.dispose()`, then fire a Running→Stopped state change and assert the handler was **not** called. This is the exact regression `8df5129c` fixed and it currently has no coverage.

packages/apex-lsp-client/test/transports/languageClientConnection.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+import { describe, it, expect, beforeEach, jest } from '@jest/globals';
+import type { BaseLanguageClient, State } from 'vscode-languageclient';
+import type { Disposable } from '@salesforce/apex-lsp-shared';
+import { LanguageClientConnection } from '../../src/transports/languageClientConnection';
+
+// Mirror the State enum values from vscode-languageclient (Stopped=1, Running=2,
+// Starting=3). We cannot import State as a value because vscode-languageclient
+// requires the `vscode` module at runtime which is unavailable in unit tests.
+const State_Stopped = 1 as unknown as State;
```
kylewalke, line 17: **(Nit) Cast inconsistency with the source.** Here it's `1 as unknown as State`; `languageClientConnection.ts:16` uses `1 as State` for the same value. Both compile — just harmonize for readability.

kylewalke (approved):
## Approve with minor notes

Clean, well-scoped adapter — pure 1:1 delegation to `BaseLanguageClient`, correct conformance to the `RpcConnection` port, and it mirrors the sibling `JsonRpcConnection` structure. CI is fully green.

**Web-bundle safety (verified):** `vscode-languageclient` is imported **type-only**, so exporting `LanguageClientConnection` from `browser.ts` pulls **zero runtime code** into the web bundle — no `__require` landmine. I confirmed no value-level import of `vscode-languageclient` exists anywhere in `src/`. Good instinct given recent web-worker bundling regressions.

**Orphan claim verified:** nothing outside this file and its tests references the class at runtime — matches the README's documented "intentional orphan" design.

None of the inline notes below block merge. The one I'd genuinely push for is the **`disposing`-guard test gap** — it's a deliberately-added behavior (commit `8df5129c`) with a real regression history and currently zero coverage.

Inline comments mark exactly where each note applies.

PR #556, opened 2026-07-15
refactor(testbed): adopt apex-lsp-client SDK, delete ApexJsonRpcClient - W-23163202

peternhale:
## Summary

- Testbed adopts `@salesforce/apex-lsp-client` SDK via `createHeadlessClient` and `ApexClientCore`, replacing 1277-line untyped `ApexJsonRpcClient` with typed SDK methods and middleware
- Creates testbed-local `ApexLspTestClient` wrapper (adds `openTextDocument`, `closeTextDocument`, `isHealthy`, `waitForHealthy`, `getServerCapabilities` for test convenience) and `MockRpcConnection` (in-memory RpcConnection for demo mode)
- Rewrites capturing middlewares as `ApexClientMiddleware` with `next` continuation, updates factories/CLI/harnesses/tests to use SDK, deletes `ApexJsonRpcClient.ts` and `web-worker` dependency

## Plan

See [docs/plans/W-23163202.md](docs/plans/W-23163202.md)

## Reviewer notes

The following lower-severity findings remain (no auto-fix applied per severity threshold):

- **Low**: MockRpcConnection lacks dedicated unit tests (only exercised indirectly through ApexLspTestClient and ApexJsonRpcClientWebWorker tests). Adding targeted unit tests for document store lifecycle and mock response methods requires a design decision about test granularity for test utilities.
- **Low**: ApexLspTestClient wrapper methods (closeTextDocument, updateTextDocument, isHealthy, waitForHealthy, getServerCapabilities) lack e2e coverage with a real server process. Extending logging-protocol.integration.test.ts to call these methods would require design decisions about what constitutes meaningful coverage for trivial delegates.
- **Low**: Plan (docs/plans/W-23163202.md) Verification section does not include an explicit manual CLI smoke-test step (e.g., 'run node out/cli.js --server jorje'). Adding this is a process/documentation decision about whether to formalize CLI verification.
- **Low**: ApexLspTestClient.waitForHealthy uses manual polling with setTimeout instead of Effect.repeat with Schedule. No Effect.repeat/Schedule usage exists elsewhere in the codebase to establish precedent; adopting this pattern would introduce a new dependency usage pattern (design decision).

## Test plan

- [ ] `npm run compile` passes in testbed (typecheck confirms SDK wiring)
- [ ] `npm run test` passes in testbed (unit tests exercise SDK-backed factories via ApexLspTestClient wrapper)
- [ ] `npm run test:integration` passes (logging-protocol test confirms real server works through SDK middleware chain)
- [ ] `npm run lint` passes
- [ ] `rg "ApexJsonRpcClient" packages/apex-lsp-testbed/src/` returns 0 results (class deleted)
- [ ] `web-worker` no longer in testbed `package.json` dependencies
- [ ] No Playwright e2e tests exercise the testbed directly (e2e-tests/ drives VS Code UI via Playwright page objects, not testbed client classes); manual verification of testbed migration relies on the unit and integration tests above

@W-23163202@

---

🤖 Generated by auto-build pipeline. Original WI: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002d1nFrYAI/view


packages/apex-lsp-testbed/src/test-utils/NotificationCapturingMiddleware.ts
```diff
@@ -18,44 +21,42 @@ export interface CapturedNotification {
...
+    this.disposable = client.use(this);
+    return this.disposable;
+  }
 
-    this.disposables.push(disposable);
+  // --- ApexClientMiddleware implementation ---
 
-    // Also capture $/logMessage if needed
-    const disposable2 = client.onNotification('$/logMessage', (params: any) => {
+  onNotification<P>(method: string, params: P, next: (p: P) => void): void {
```
kylewalke, line 46: **High — this middleware will capture zero `window/logMessage` notifications.**

This `onNotification` hook only fires if `ApexClientCore` routes incoming notifications through the middleware chain. In `apexClientCore.ts`, the incoming-notification chain runs *only* inside `registerIncomingNotification`, which is currently wired up exclusively for the typed `apex/*` surface (`requestWorkspaceLoad`, `workspaceIngestionComplete`, `queueStateChanged`). The SDK even comments that `_registerIncomingNotification` is "Not currently wired to a public method … available for future incoming notification handlers."

There is no catch-all `connection.onNotification`, and nothing registers `window/logMessage`/`$/logMessage`. The old client captured these at the transport level via `connection.onNotification((method, params) => …)`. So `getLogMessages()` will now always return `[]`.

The `logging-protocol.integration.test.ts` assertions are all guarded by `if (logMessages.length > 0)`, so the suite will **pass while verifying nothing** — silently gutting the `window/logMessage` type-4 protocol coverage that suite exists for. Please confirm against a real server; if confirmed, this needs a different capture path (SDK exposing a catch-all incoming hook, or capturing at the transport).

packages/apex-lsp-testbed/src/utils/clientFactory.ts
```diff
@@ -6,282 +6,58 @@
...
- * Creates an appropriate client instance based on server type
+ * Creates an appropriate SDK-backed client instance based on server type.
+ *
+ * For demo mode: uses MockRpcConnection + ApexClientCore.create
+ * For real servers: uses createHeadlessClient (spawns process + creates core + listens)
+ *
+ * Returns an ApexLspTestClient wrapping the core.
  */
-export function createClient(
+export async function createClient(
```
kylewalke: **Medium — `createClient` is now dead code.**

`cli.ts` was rewritten to inline both the demo and real-server creation paths, so nothing calls `createClient` anymore (only the unrelated `createClientOptions`). With `MockApexJsonRpcClient` deleted, this whole file is just an orphaned `createClient` + `module.exports`. Either delete the file or re-wire `cli.ts` to call it. `knip` should flag this.

packages/apex-lsp-testbed/src/test-utils/ApexLspTestClient.ts
```diff
@@ -0,0 +1,180 @@
...
+   * Simplified health check: returns true if the core has not been disposed.
+   */
+  isHealthy(): boolean {
+    return !this.core.isDisposed();
+  }
+
+  /**
+   * Poll `isHealthy()` with exponential backoff until healthy or timeout.
+   */
+  async waitForHealthy(timeoutMs: number = 30_000): Promise<void> {
```
kylewalke, line 97: **Medium — `isHealthy()`/`waitForHealthy()` semantics silently weakened.**

The old `isHealthy()` sent a real `$/ping` (or verified capabilities) — an actual liveness probe. This one returns `!core.isDisposed()`, which is `true` immediately after construction regardless of whether the server process is alive. As a result:
- `waitForHealthy(120_000)` in `serverFactory.ts` returns on the first poll basically always, and no longer detects an unresponsive/dead server.
- The perf test's `if (!isHealthy()) throw` guard is effectively a no-op.

The plan calls this the "simplest contract," but it turns a functional check into a disposed-flag check. At minimum worth a comment; ideally reconsider whether `waitForHealthy`'s 120s timeout still earns its signature.

packages/apex-lsp-testbed/src/test-utils/RequestResponseCapturingMiddleware.ts
```diff
@@ -19,44 +22,86 @@ export interface RequestResponsePair {
...
+        throw error;
+      },
+    );
+  }
 
-    this.connection.sendRequest = this
-      .originalSendRequest as typeof this.connection.sendRequest;
-    this.connection = null;
-    this.originalSendRequest = null;
+  sendNotification<P>(method: string, params: P, next: (p: P) => void): void {
```
kylewalke, line 90: Low — this records the pair *before* calling `next(params)`, and notification pairs carry an unused `id`/`timestamp` (no `duration`/`response` is ever set for notifications). Functionally fine, just slightly wasteful and ordering-inconsistent with the request path.

packages/apex-lsp-testbed/src/cli.ts
```diff
@@ -267,26 +270,56 @@ async function main(): Promise<void> {
...
         JSON.stringify(capabilities, null, 2),
       );
-      await client.stop();
+      await core.shutdown();
+      await core.dispose();
       console.log('Server stopped');
       process.exit(0);
+    } else if (
+      options.serverType === 'nodeServer' ||
+      options.serverType === 'webServer'
```
kylewalke, line 279: Low — behavior change not called out in the plan: previously only `nodeServer` ran the test battery; this `else if (nodeServer || webServer)` now routes `webServer` through `runNodeServerTests` too. Probably intended, but worth confirming.

packages/apex-lsp-testbed/src/servers/webServer/webLanguageServerHarness.ts
```diff
@@ -257,41 +265,39 @@ public class CompletionTest {
...
+    });
 
     // Log hover results
     if (hoverResult && hoverResult.contents) {
-      if (typeof hoverResult.contents === 'string') {
-        this.logger.info(`Hover content: ${hoverResult.contents}`);
-      } else if (hoverResult.contents.kind && hoverResult.contents.value) {
-        this.logger.info(`Hover content (${hoverResult.contents.kind}):`);
-        this.logger.info(hoverResult.contents.value);
+      const contents = hoverResult.contents as any;
```
kylewalke: Low — `hoverResult.contents as any` reintroduces an `any` cast, whereas `cli.ts` uses the cleaner `'kind' in contents` narrowing for the same logic. Inconsistent with the PR's "no `any` leakage" goal (dev harness, so minor).

kylewalke (commented):
Reviewed the SDK adoption end-to-end. Structurally this is a clean, well-sequenced refactor and the `ApexLspTestClient` wrapper + middleware rewrite to the `next`-continuation model are the right shape. My main concern is a likely-silent regression in the logging integration coverage (High, inline). A few cleanups below as well.

Note: I verified the High finding by reading the SDK's `apexClientCore.ts` on this branch — it does not appear to route arbitrary incoming notifications (e.g. `window/logMessage`) through the middleware chain, so please confirm against a live server before merging.

packages/apex-lsp-testbed/src/test-utils/NotificationCapturingMiddleware.ts
kylewalke: **Follow-up: confirmed the High finding definitively against the SDK on this branch (commit 7961612).**

Traced every notification-registration path in `@salesforce/apex-lsp-client`:

1. **Exactly one** `connection.onNotification(...)` call exists in the core — `apexClientCore.ts:266`, inside `_registerIncomingNotification`. That closure is the *only* place the incoming-notification middleware chain (`composeNotificationChain`) is invoked.
2. `_registerIncomingNotification` is called from **exactly three** sites, all in `apexMethods.ts` (358, 375, 392), and all for `apex/*` methods:
   - `apex/requestWorkspaceLoad`
   - `apex/workspaceIngestionComplete`
   - `apex/queueStateChanged`
3. The only handler auto-registered at `create()` time is `apex/findMissingArtifact` — a **request** (`registerIncomingRequest`), not a notification.
4. The transport layer (`jsonRpcConnection.ts:47`) delegates straight to `vscode-jsonrpc`'s per-method `onNotification(method, handler)` — there is **no star/catch-all** handler.

**Conclusion:** `window/logMessage` and `$/logMessage` are never registered anywhere, so they never enter the middleware chain. `NotificationCapturingMiddleware.onNotification` will not fire for them, and `getLogMessages()` returns `[]` in all cases. Because every assertion in `logging-protocol.integration.test.ts` is guarded by `if (…length > 0)`, the suite passes green while asserting nothing — the W-23163202-adjacent `window/logMessage` type-4 coverage is silently lost.

**Suggested paths forward** (any one):
- Have the SDK expose a catch-all incoming-notification hook (e.g. a wildcard `onNotification` or a `use()`-visible pre-dispatch tap), then route capture through it; or
- Capture at the transport (`connection.onNotification`) as the old client did — e.g. `createHeadlessClient` returns `connection`, so the test could register directly on it; or
- Have the core register the `window/logMessage` method through `_registerIncomingNotification` so it flows through the chain.

This doesn't block the structural refactor, but the logging suite should regain real coverage (or be explicitly marked as no-longer-covering-this) before merge.

packages/apex-lsp-testbed/test/performance/lsp-benchmarks.perf.ts
```diff
@@ -300,9 +293,9 @@ describe(testTitle, () => {
       const requestTimeout = 2_000; // 2 second timeout per request
       const results: Record<string, Benchmark.Target> = {};
 
-      // Ensure client is started
-      if (!serverContext.client._isRunning) {
-        await serverContext.client.start();
+      // Ensure client is healthy
+      if (!serverContext.client.isHealthy()) {
```
kylewalke: **Regression from the `isHealthy` fix — this guard is now dead.**

`isHealthy()` is now `async` and returns `Promise<boolean>`. `!Promise` is always `false`, so this `if` can never be entered and the benchmark will proceed even against a dead/unresponsive server. This is the same guard the original Medium finding flagged as weakened — it's still broken, now because the `Promise` is unconditionally truthy.

```ts
if (!(await serverContext.client.isHealthy())) {
  throw new Error('Server is not healthy, cannot run benchmarks');
}
```

Not caught by typecheck (`!promise` is legal TS and produces no error), so CI stays green.

packages/apex-lsp-testbed/examples/web-worker-usage.ts
```diff
@@ -6,122 +6,105 @@
...
+    // Show server capabilities
+    const capabilities = client.getServerCapabilities();
+    console.log(
+      `[Example] Server capabilities: ${JSON.stringify(capabilities, null, 2)}`,
+    );
 
     // Check if the server is healthy
-    const isHealthy = await client.isHealthy();
-    logger.info(`Server health check: ${isHealthy ? 'OK' : 'FAILED'}`);
+    const isHealthy = client.isHealthy();
```
kylewalke: **Same `isHealthy` regression here.** `client.isHealthy()` now returns a `Promise`, which is always truthy, so the `if (isHealthy)` block below always executes and the logged status is meaningless.

```ts
const isHealthy = await client.isHealthy();
```

Worth noting `examples/` isn't covered by any tsconfig (neither `tsconfig.json` nor `test/tsconfig.json` includes it), so nothing will flag this — it's a silent latent bug.

kylewalke (commented):
Re-reviewed after commit `0c5addd`. Thanks for the thorough pass — 5 of 6 items are resolved:

- ✅ **High** — core now registers no-op incoming handlers for `window/logMessage` + `$/logMessage` via `_registerIncomingNotification`, so they flow through the middleware chain and `NotificationCapturingMiddleware` can capture them. Verified the plumbing against `apexClientCore.ts`.
- ✅ **Medium** — dead `createClient` removed (file is now a documented stub, no dangling importers).
- ✅ **Low** — `sendNotification` now calls `next()` before recording.
- ✅ **Low** — harness hover uses `'kind' in contents` narrowing instead of `as any`.
- ✅ **Bonus** — span-collection integration test migrated off the deleted client.

The **Medium `isHealthy` fix** is correct in the wrapper (now probes `$/ping`), but making it `async` introduced two un-awaited call sites that CI won't catch. Details inline.

**Coverage note on the High fix (not blocking):** now that log notifications actually flow through the middleware chain, `logging-protocol.integration.test.ts` was not updated — every assertion there is still guarded by `if (logMessages.length > 0)`, so if capture silently regresses again the suite still passes green (exactly the failure mode that let the original High bug through). Since capture now works, consider asserting `expect(logMessages.length).toBeGreaterThan(0)` in at least one representative test so a future regression fails loudly.

packages/apex-lsp-testbed/test/performance/lsp-benchmarks.perf.ts
peternhale: Fixed in 1d6b201d — `if (!(await serverContext.client.isHealthy()))`. The guard is live again; benchmarks now abort against an unresponsive server. Verified via `tsc -p test/tsconfig.json`.

packages/apex-lsp-testbed/examples/web-worker-usage.ts
peternhale: Fixed in 1d6b201d — `const isHealthy = await client.isHealthy();`. Confirmed your note that `examples/` is outside both tsconfigs (`tsconfig.json` = `src/**/*`, `test/tsconfig.json` rooted at `test/`), so this one is unchecked by the build — fixed by inspection.

kylewalke (dismissed):

kylewalke (approved):

PR #557, opened 2026-07-16
feat(apex-ls): textDocument/codeAction infrastructure - W-22629623

kylewalke:
### What does this PR do?

Adds `textDocument/codeAction` support to the Apex language server as a **development-only capability** (steel thread). Three quick fixes ship behind the dev capability profile:

- **Extract Local Variable** — extracts the selected expression into a new `Object` local declaration.
- **Extract Constant** — extracts the selected expression into a `static final` constant.
- **Declare Missing Method** — generates a stub for an unresolved method call, inserted inside the enclosing class body.

Supporting changes:

- New CST finders in `apex-parser-ast`: `findExpressionAtRange` and `findMethodCallAtRange`, which locate the enclosing expression / method-call node for a given selection or cursor.
- Wires the `onCodeAction` handler and the `codeActionProvider` capability (dev profile only).
- Advertises `codeActionProvider` in the `initialize` response's `staticCapabilities` — without this the client shows "No refactorings available" even with the handler registered.
- E2E coverage for Extract Variable (Playwright), plus page-object helpers for driving the Monaco action widget.

Edits are computed eagerly as `WorkspaceEdit`s (no `codeAction/resolve` round-trip).

### Review fixes included

- Correct the Declare Missing Method insertion span so the stub lands inside the class body (`endColumn` points one past the closing brace).
- Extract Variable/Constant now replace the actual enclosing-expression span rather than the raw selection.
- Reuse `context.parseTree` instead of re-parsing the document per request.

@W-22629623@


peternhale (commented):
## Review — non-blocking "would be nice to fix"

Reviewed the full diff (verified against head `fe801ec2`). This is high-quality, well-tested, dev-gated work — **no blocking correctness defects found**. The items below are all nice-to-have polish / follow-ups, posted as a comment rather than a change request.

### 1. (Medium) Cross-package `instanceof` on ANTLR CST classes — latent silent-failure risk
`expressionRangeFinder.ts` deliberately avoids `instanceof ParserRuleContext` across the package boundary (uses a structural guard) because the CST node's class identity comes from **parser-ast**'s bundled `@apexdevtools/apex-parser`. But `CodeActionProcessingService` (`isLiteralExpression`, `findConstantInsertion`, ~L468–513) now does `expression instanceof PrimaryExpressionContext` / `ClassDeclarationContext` where those classes come from **lsp-compliant-services'** newly-added copy of the same package. Both pin `5.1.0` exactly, so npm dedup collapses them today and it works. If those versions ever drift (or hoisting changes), every extract/constant action would silently stop being offered with **no error**. Suggest a guardrail or at minimum a comment calling out the exact-version coupling.

### 2. (Low–Medium) Abstraction inconsistency vs. the boundary this PR established
The finders were designed so the LS layer never touches ANTLR types (`MethodCallAtRange`/`ExpressionAtRange` are neutral). Yet the service now imports `PrimaryExpressionContext`, `NegExpressionContext`, `PreOpExpressionContext`, `ClassDeclarationContext`, `LiteralPrimaryContext`, `ParserRuleContext`. The literal check + class-body insertion walk belong next to `findExpressionAtRange` in `apex-parser-ast`. Moving them eliminates #1 entirely. Good follow-up before production enablement (05.4).

### 3. (Low) `shutdown()` doesn't clear pending debounce timers
`LSPQueueManager.shutdown()` (~L1067) nulls the singleton but leaves up to `CODE_ACTION_COALESCE_MS` of scheduled `setTimeout`s. When they fire, `submitRequest` throws `RequestCancelledError` (mapped to `null`), so no crash — but clearing `pendingCodeActions` and `clearTimeout`-ing on shutdown would be tidier.

### 4. (Low) Declare-Missing-Method insertion position comes from indexed symbols, not the live target file
`computeInsertPosition` reads `typeSymbol.location.symbolRange` from the symbol manager. If the (unopened) target file drifted from the index, the stub could land at a stale position. Acceptable for a dev-only steel thread; worth a caveat comment.

### 5. (Low) e2e `applyCodeAction` fallback applies whatever is focused
`ApexEditorPage.applyCodeAction` (~L296–307) arrows down until the focused row matches, but if no row ever matches it still presses `Enter`, applying whatever is focused. A miss should probably fail loudly rather than apply an arbitrary action.

---
Nice touches worth calling out: the enclosing-expression replacement invariant is directly regression-tested (sub-selection + zero-width cursor), the `endColumn - 1` brace fix is verified by an applied-edit brace-balance assertion, and the per-URI coalescing rationale is sound and well-covered.

kylewalke:
Thanks Pete — really appreciate the thorough pass. Here's the disposition:

**Fixed in this PR (see incoming commit):**
- **#/3 (shutdown doesn't clear pending debounce timers)** — `shutdown()` now drains `pendingCodeActions` by calling each entry's `supersede()` (clears the timer, disposes the token sub, removes the map slot, rejects with `RequestCancelledError` → null). Added a regression test ("drains requests still in the debounce window on shutdown") that asserts a mid-window request is rejected and never dispatches.
- **#/4 (Declare-Missing-Method insert position from stale index)** — added a CAVEAT comment on `computeInsertPosition` noting the position comes from the indexed symbol (not the live target file), that a drifted unopened target can land the stub at a stale offset, and that 05.4 (production enablement) should re-read/reparse the target to anchor against current content.
- **#/5 (e2e `applyCodeAction` applies whatever is focused on a miss)** — the focus walk now tracks whether it landed on the target row and `expect(matched).toBe(true)` before pressing Enter, so a broken focus walk fails loudly instead of applying an arbitrary action.

**Deferred to W-23544057 (new work item under the Jorje Parity epic):**
- **#/1 (cross-package `instanceof` on ANTLR CST classes)** and **#/2 (abstraction inconsistency — LS layer importing ANTLR types)** are folded into that story. They share a root with a separately-filed indentation bug in `findConstantInsertion` (it derives indent from the `class`-keyword column instead of physical leading whitespace). The combined fix direction — move the literal check + class-body insertion walk into `apex-parser-ast` next to `findExpressionAtRange`, returning a neutral result with no ANTLR types crossing the boundary — eliminates #/1, resolves #/2, and fixes the indent bug for free (the finder already computes char-stream-accurate indent via `leadingIndentOf`). Scheduled before production enablement (05.4).

Thanks again!

peternhale (approved):
All review items addressed. Fixes #3/4/5 verified in cf5797db; #1/2 appropriately deferred to W-23544057 before production enablement. LGTM ✅

PR #576, opened 2026-07-22
refactor(apex-parser-ast): fix Extract Constant indentation + neutral constant-extraction finder - W-23544057

kylewalke:
### What does this PR do?

Fixes the Extract Constant code action inserting the generated class-level declaration with far too much leading whitespace (observed ~5-6 tab-widths instead of one indent level). Also resolves PR #557 review items 1 & 2 (cross-package ANTLR coupling), which share the same root cause.

**Root cause:** `CodeActionProcessingService.findConstantInsertion` derived the member indent from the *column* of the `class` keyword token. Because `class` follows visibility/sharing modifiers (`public with sharing class` → column ~18), that column was treated as an indentation width, producing excessive leading spaces. Extract Variable was unaffected because it reads the physical leading whitespace via `leadingIndentOf`.

### How is it implemented?

- Moves the literal-expression check and class-body insertion-point computation into `apex-parser-ast` (`expressionRangeFinder.ts`), alongside `findExpressionAtRange`, as a new `findConstantExtraction` returning a **neutral** `ConstantExtraction` result (insert offset + physical char-stream indent + `isInner` + `isLiteral`) — no ANTLR types cross the package boundary.
- Indent is now the enclosing class declaration's **physical line indentation** (read from the char stream) plus exactly one indent unit — the same mechanism Extract Variable uses.
- Rewrites `CodeActionProcessingService` to consume the neutral result and **removes the cross-package `instanceof` on ANTLR CST classes** (`PrimaryExpressionContext`, `LiteralPrimaryContext`, `PreOpExpressionContext`, `NegExpressionContext`, `ClassDeclarationContext`, `ExpressionContext`). This eliminates the silent-failure risk if the two packages' `@apexdevtools/apex-parser` copies ever drift (review item 1) and restores the neutral finder abstraction (review item 2).

### Sequencing

This PR should merge **before** W-23389340 (production enablement of codeAction), so production ships the correct indentation and clean abstraction from the outset.

### Testing

- New `findConstantExtraction` unit tests in `apex-parser-ast` (literal detection, insert offset, indent across class shapes).
- New `CodeActionProcessingService` regression tests asserting the inserted constant's indent equals class-line indent + one level for top-level (`public class`), sharing-modifier (`public with sharing class`), and inner-class sources.
- Full pre-commit suite green.

@W-23544057@


peternhale:
## Architectural Observation: Separation of Concerns

This PR correctly fixes the indentation bug and eliminates ANTLR type coupling. However, our discussion revealed a broader architectural issue worth addressing in a follow-up:

### Current State
`CodeActionProcessingService` (in `lsp-compliant-services`) parses the document to obtain a CST, then walks it to extract expression-level data:
- Line 208: `compilerService.compile()` to get `parseTree`
- Line 297: `findExpressionAtRange(parseTree, range)` to walk CST
- Line 346: `findConstantExtraction(expression)` to compute insertion point

This creates a separation-of-concerns violation: the LS layer is reaching into parsing details (ANTLR CST types).

### Discovery
The `ApexSymbolCollectorListener` **already captures** expression-level data:
- Line 3584-3629: Literals stored as `SymbolReference` with `ReferenceContext.LITERAL`
- Line 3064-3110: Method calls stored with arguments and location
- `SymbolTable.getReferencesAtPosition()` provides position-based query

This means the CST parse in `CodeActionProcessingService` may be redundant or could be moved behind a cleaner API boundary.

### Recommendation
**This PR should merge as-is** — the fix is correct and the architectural cleanup is orthogonal.

**Follow-up work:** Refactor to maintain proper separation:
1. **apex-parser-ast** should expose semantic query APIs (e.g., `findExpressionAtPosition`) that return neutral data structures
2. **lsp-compliant-services** should call those APIs instead of parsing/walking CST directly
3. **apex-parser-ast** owns all parsing and tree-walking internally

This would ensure `lsp-compliant-services` never imports ANTLR types and `apex-parser-ast` remains the single owner of parsing concerns.

cc @kylewalke for awareness of the broader refactoring opportunity.

packages/apex-parser-ast/src/utils/expressionRangeFinder.ts
```diff
@@ -47,6 +59,57 @@ export interface ExpressionAtRange {
...
+   */
+  expressionEnd: number;
+  /**
+   * The verbatim source text of the enclosing expression (whitespace
+   * preserved), read from the shared `CharStream`. This is the text an extract
+   * refactoring substitutes for the generated symbol name — it must retain the
+   * author's spacing (e.g. `1 + 2 * 3`), unlike `expression.getText()` which
+   * collapses whitespace.
+   */
+  expressionText: string;
```
peternhale, line 85: **Seal the last ANTLR leak into the LS layer (follow-up to #557 items 1 & 2).**

The runtime coupling is genuinely gone — no `instanceof`, no token-internal reads, no parse orchestration in `CodeActionProcessingService`. Nice. But `ExpressionAtRange.expression: ExpressionContext` (line 54, just above this hunk) is still a **public** field, so `ExpressionContext` remains transitively reachable in the LS layer's compile-time type graph through `findExpressionForCodeAction`'s return type — even though the LS code never names or dereferences it.

To make the neutral boundary airtight (and prevent erosion later), consider marking `expression` `@internal` / `readonly` and having `codeActionParse.ts` treat the handle as opaque the way `CodeActionParseContext.parseTree` already is — or dropping the field from the public shape entirely if only the same-package finders need it.

Non-blocking — the substantive leak is resolved; this just prevents the abstraction from quietly leaking back.

kylewalke: Good catch — sealed in 66856a66. Marked `ExpressionAtRange.expression` `@internal readonly`, mirroring the `CodeActionParseContext.parseTree` opaque-handle precedent. Kept the field (not dropped) since the same-package `findConstantExtractionFromExpression` still reuses the located node; the LS layer only ever destructures the neutral fields (`statementStart`/`indent`/`expressionStart`/`expressionEnd`/`expressionText`) and never touches `.expression`, so the ANTLR type is no longer reachable in its type graph.

Also FYI on your earlier separation-of-concerns note: that broader refactor is now tracked — the parse-ownership/type-resolution work has been folded into W-23389334 (expression type inference for the Extract actions), which builds directly on this facade.

packages/apex-parser-ast/src/utils/codeActionParse.ts
```diff
@@ -0,0 +1,121 @@
...
+): CodeActionParseContext | null => {
+  try {
+    const result = new CompilerService().compile(
+      text,
+      uri,
+      new ApexFoldingRangeListener(),
+      { includeComments: false },
+    );
+    return { parseTree: result.parseTree };
+  } catch {
```
peternhale, line 60: **Silent parse failures — minor observability regression (non-blocking).**

The old `CodeActionProcessingService.getParseTree` logged `Unable to parse document for extract actions: ${error}` on catch. `parseForCodeActions` now swallows the exception with a bare `catch {}` and no logging, so a document that fails to parse degrades to "no code actions offered" with zero diagnostic trail.

I understand the intent — keeping `apex-parser-ast` free of a logger dependency — so this is defensible for a best-effort code-action path. But consider having the LS caller log at `debug` when `parseForCodeActions` returns `null` (in `analyzeCodeActionContext`), to preserve the signal that used to be here. Not blocking.

kylewalke: Fixed in 689a7c72. Kept `apex-parser-ast` logger-free — `parseForCodeActions` still swallows the parse failure and returns `null` — and restored the signal at the LS caller: `analyzeCodeActionContext` now logs at `debug` (`Unable to parse document for code actions: ${uri} — expression-based actions will be unavailable`) when the parse returns null, mirroring the former `getParseTree` trail. Thanks for catching the observability regression.

peternhale (approved):

PR #577, opened 2026-07-22
feat: enable codeActionProvider in production capabilities - W-23389340

kylewalke:
### What does this PR do?

Enables the `codeActionProvider` capability in **production** (05.4). The codeAction steel thread (W-22629623, PR #557) shipped code actions as development-only; this promotes them to the released production profile so Extract Local Variable, Extract Constant, and Declare Missing Method are offered in production.

### How is it implemented?

- `PRODUCTION_CAPABILITIES.codeActionProvider` is set to `{ codeActionKinds: [QuickFix, RefactorExtract] }` (was `undefined`), mirroring Jorje's `CodeActionOptions([QuickFix, RefactorExtract])`.
- `DEVELOPMENT_CAPABILITIES` now inherits the identical capability via `...PRODUCTION_CAPABILITIES` (removed the redundant override).
- The `LCSAdapter` `staticCapabilities` copy was already generic (`if (allCapabilities.codeActionProvider) …`), so the capability flows through to the `initialize` response automatically — no logic change, comment updated to reflect both profiles.
- `codeActionProvider` is not in `WEB_DISABLED_CAPABILITIES`/`DESKTOP_DISABLED_CAPABILITIES`; the minimal desktop profile (`PRODUCTION_DESKTOP_CAPABILITIES`) is intentionally left untouched.

### Sequencing

⚠️ **Merge this AFTER #576 (W-23544057).** That PR fixes the Extract Constant indentation bug and the cross-package ANTLR coupling. Shipping production enablement first would expose the incorrect indentation to production users.

### Testing

- Flipped the `capabilities-alignment` regression guard from "codeActionProvider only in development" to "in both production and development" — a future silent drop of the production capability now fails loudly.
- Regenerated capability snapshots; the only changes are the intended `codeActionProvider` entries flipping from `undefined` to the kinds object in the production/web sub-trees (no unrelated churn; development and minimal-desktop snapshots unchanged).
- Full pre-commit suite green.

@W-23389340@


peternhale (changes requested):
## Request: Add Feature Flag for Production Control

While the implementation is correct, the capability is now **unconditionally enabled** in production with no runtime control mechanism. This poses a rollback risk if issues are discovered post-release.

### Required Change

Add a feature flag to gate the production capability, following the pattern used for other features (e.g., `ENABLE_REFERENCES_PROVIDER`):

```typescript
// In ApexLanguageServerCapabilities.ts
const ENABLE_CODE_ACTIONS = process.env.APEX_ENABLE_CODE_ACTIONS !== 'false'; // Default: true

export const PRODUCTION_CAPABILITIES: ExtendedServerCapabilities = {
  // ...
  codeActionProvider: ENABLE_CODE_ACTIONS
    ? {
        codeActionKinds: [CodeActionKind.QuickFix, CodeActionKind.RefactorExtract],
      }
    : undefined,
  // ...
};
```

### Rationale

1. **Emergency rollback** - If critical bugs surface in production, operators can disable via environment variable without requiring a hotfix deploy
2. **Gradual rollout** - Allows controlled enablement across different deployment tiers (canary → staging → production)
3. **Consistency** - Matches the pattern already established for `referencesProvider` and other production features
4. **Zero risk** - Default `true` means the feature ships enabled; the flag only provides an escape hatch

### Alternative

If the team's policy is "no feature flags for fully-baked features," please document:
- The rollback procedure if issues are found
- Why code actions are treated differently than references (which has a flag)

Otherwise, adding the flag is standard production risk mitigation.

kylewalke:
Closing without merging. On review of the current `main`, `codeActionProvider` is **already gated development-only** by default — `PRODUCTION_CAPABILITIES.codeActionProvider` is `undefined` and only `DEVELOPMENT_CAPABILITIES` advertises the `[QuickFix, RefactorExtract]` kinds, with `ApexCapabilitiesManager` defaulting to `production` mode. That is exactly the desired behavior: code actions stay hidden from users until we're ready to enable them.

This PR's original purpose was to *enable* code actions in production (05.4), which we've decided to defer. Since the dev-only gate we want already exists on main, there's nothing to merge here. W-23389340 (production enablement) remains deferred until the feature is ready to release.

PR #578, opened 2026-07-23
perf(apex-ls): improve workspace load and LSP request performance - W-23559100,W-23448544

peternhale:
### What does this PR do?

Improves Apex language-server workspace loading and LSP request performance using an Effect Worker-managed topology that works in Node today and retains browser support.

Key changes include:

- Runs workspace compilation through a configurable Effect Worker pool, with a default pool size of 2 and concurrency of 1 per compute-bound compiler worker.
- Keeps the DataOwner as the authoritative symbol and graph owner while compiling workspace batches concurrently.
- Defers cross-file reference resolution during workspace ingestion and drains deferred work after the batch load.
- Reduces redundant symbol resolution, FQN lookup, graph work, and same-file reference processing.
- Uses shared request-preparation policies across hover, definition, references, implementation, completion, signature help, document symbols, code lens, diagnostics, and code actions.
- Adds bounded full-detail cursor preparation for requests that need more detail without eagerly enriching every unresolved symbol.
- Adds a precomputed standard-library FQN index and correct generic substitution for standard-library collection APIs.
- Loads configured standard-library namespaces into the authoritative DataOwner through a single bulk resource-loader Effect Worker request.
- Adds desktop and browser worker coverage, performance instrumentation, profiling scripts, and trace-analysis documentation.

The latest desktop trace loaded 648 Apex files in 7.55 seconds. Standard-library preload loaded 172/172 `Database` and `System` classes in 301 ms, including 46 ms in the bulk resource-loader worker request.

@W-23559100@
@W-23448544@ 

### Functionality Before

Workspace compilation and symbol-graph updates performed redundant work and concentrated compilation on the DataOwner path. LSP requests could trigger broad enrichment or repeated resolution, and coordinator-local standard-library preload did not populate the worker-owned graph.

### Functionality After

Workspace compilation is distributed across Effect-managed compiler workers while DataOwner commits remain authoritative. Resolution is deferred or narrowed to the symbols required by each request, request preparation is shared across LSP features, and configured standard-library namespaces are preloaded directly into the DataOwner graph.

Validation:

- Repository lint and formatting checks passed.
- 381 test suites completed.
- 4,832 tests passed, 0 failed, 489 pending.
- Manual desktop and browser extension-host testing completed throughout the performance iterations.
- Desktop OpenTelemetry traces confirm the cross-worker topology, workspace-load behavior, request preparation, and standard-library preload.


packages/apex-lsp-vscode-extension/esbuild.config.ts
```diff
@@ -124,6 +124,14 @@ const builds: BuildOptions[] = [
             from: ['../apex-ls/dist/worker.platform.web.js.map'],
             to: ['./dist/worker.platform.web.js.map'],
           },
+          {
+            from: ['../apex-ls/dist/compiler.worker.web.js'],
+            to: ['./dist/compiler.worker.web.js'],
+          },
+          {
```
kylewalke: 🔴 **HIGH — `compiler.worker.node.js` is never bundled into the packaged extension.**

This copy block ships `compiler.worker.web.js` but not the node counterpart. The node worker is built into `apex-ls/dist/compiler.worker.node.js` (`packages/apex-ls/esbuild.config.ts:135`) and resolved at runtime from `worker.platform.js`'s `__dirname` (`packages/apex-ls/src/worker.platform.ts:448-451`, spawned at :483). The VSIX ships only the extension's own `dist/` (`.vscodeignore` excludes `node_modules/**` and `src/**`; no `files` allowlist).

Since `experimental.workers.enabled` defaults to **`true`** (`package.json:632`, `ApexSettingsUtilities.ts:134`), the compilation pool is on by default. In a packaged install `NodeWorkerThread` can't find the file → pool startup hits the 30s timeout (`CompilationWorkerPool.ts:59`) → every `WorkspaceBatchCompileOnDataOwner` rejects (`worker.platform.shared.ts:3857-3862`). This passes local F5/dev (where `worker.platform.js` runs from `apex-ls/dist` next to its sibling) but breaks the shipped desktop extension. Web is unaffected.

**Fix:** add copy entries for `compiler.worker.node.js` (+`.map`).

packages/apex-ls/src/server/WorkspaceBatchHandler.ts
```diff
@@ -616,17 +447,13 @@ function processStoredBatches(
...
-          yield* processViaDataOwner(sessionId, entries, dispatcher, logger);
-          fileUris = entries.map((e) => e.uri);
-        } else {
-          yield* processLocally(sessionId, batches, logger);
-        }
-
+        const entries = yield* decodeBatches(batches, logger);
+        // The decoded entries now own the source content. Release the now
+        // redundant base64/ZIP session payload before compilation begins.
         batchStorage.removeSession(sessionId);
```
kylewalke, line 453: 🔴 **HIGH — workspace-load session leaks on compile failure.**

This top-level `catchAll` removes local batch storage but never ends the data-owner workspace-load session. The session is opened by `BeginWorkspaceLoadSession` (line 598) and ended **only** via `DrainDeferredReferences` on the success path (line 742). But the pipeline consumer aborts the whole chunk on the first failed `Exit` (`WorkspaceCompilationPipeline.ts:152-153`); that error rethrows through the compile loop (~line 716) and lands here.

Because `endWorkspaceLoadSession` is only reachable through the drain, the session stays active forever. An active session makes **every** subsequent interactive compile skip `resolveCrossFileReferencesForFile` (`worker.platform.shared.ts:4253-4263`) — so cross-file references silently stop resolving for the rest of the server's life. This is a regression from the old handler, which caught per-file errors and always drained.

**Trigger:** one file where `compilerService.compile` throws or yields no `SymbolTable` (`CompilationWorkerHandler.ts:84-89`). (Ordinary syntax errors are tolerated — partial table + diagnostics — so the trigger is narrower than any bad file, but it's real.)

**Fix:** wrap the session in `Effect.ensuring`/`acquireRelease` so the drain (or an explicit end-session) always runs on failure.

packages/apex-ls/src/compiler/WorkspaceCompilationPipeline.ts
```diff
@@ -0,0 +1,173 @@
...
+            if (queueDepth <= 0) {
+              takeFromEmptyCount++;
+            }
+            const takeStart = performance.now();
+            const item = yield* Queue.take(queue);
+            const takeWaitMs = performance.now() - takeStart;
+            totalTakeWaitMs += takeWaitMs;
+            maxTakeWaitMs = Math.max(maxTakeWaitMs, takeWaitMs);
+            if (Exit.isFailure(item.result)) {
+              return yield* Effect.failCause(item.result.cause);
```
kylewalke: 🟠 **MEDIUM — whole-batch data loss on a single-file failure.**

Aborting on the first failed `Exit` discards already-compiled results still in the queue and skips every file after the failed index. The caller expects `{compiledCount, errorCount}` semantics (`WorkspaceBatchHandler.ts:709-714`), not a throw — the old handler tolerated per-file failures and continued to the drain. So one malformed file in a 100-file chunk now drops symbol ingestion for the whole chunk (and compounds the session leak in issue #2).

Consider isolating per-file failures (record into `errorCount`, continue) rather than failing the whole chunk.

packages/apex-parser-ast/src/resources/StandardApexLibrary/System/Map.cls
```diff
@@ -63,9 +63,9 @@ global class Map {
 
   /**
    * Returns a set that contains all of the keys in the map.
-   * @return Set<Object>
+   * @return Set<K>
    */
-  global Set<Object> keySet() {}
+  global Set<K> keySet() {}
```
kylewalke, line 68: 🟠 **MEDIUM — `Map` generic substitution only half-fixed.**

Good fix here (`keySet()` → `Set<K>`), but `values()` two lines down still returns `List<Object>` (:113), and `get`/`put`/`remove` still return `Object`. `applyMethodTypeSubstitutions`/`substituteTypeName` (`genericTypeSubstitution.ts:50-52,96`) only rewrite declared type params, so for `Map<Id,Account> m; m.values()[0]` the element resolves to `Object` instead of `Account` in hover/completion — an asymmetry with this `keySet()` fix.

**Fix:** `values()` → `List<V>`, `get`/`put`/`remove` → `V`. (Aside: `remove(Key key)` uses `Key`, not the declared `K` — pre-existing, but adjacent.)

packages/apex-lsp-vscode-extension/package.json
```diff
@@ -628,17 +631,86 @@
...
+                  "maximum": 50,
+                  "description": "%configuration.apex-ls-ts.experimental.workers.lspRequest.concurrency.description%"
+                }
+              }
+            },
+            "dataOwner": {
+              "type": "object",
+              "description": "%configuration.apex-ls-ts.experimental.workers.dataOwner.description%",
+              "properties": {
+                "poolSize": {
```
kylewalke: 🟠 **MEDIUM — dead configuration knobs.** Several contributed worker settings are read by nothing:
- `dataOwner.poolSize` (here) — pool hardcoded `size: 1` (`WorkerCoordinator.ts:514`).
- `resourceLoader.poolSize`/`concurrency` (:701,:707) — hardcoded `1/1` (`WorkerCoordinator.ts:497-498`).
- `lspRequest.concurrency` (:645) — the pool is created without a `concurrency` option (`WorkerCoordinator.ts:519-528`).

Users who set these get no effect and no warning. Either wire them up or drop them from the contribution.

packages/apex-lsp-shared/src/workerWireSchemas.ts
```diff
@@ -54,6 +53,10 @@ export class WorkerInit extends Schema.TaggedRequest<WorkerInit>()(
       logLevel: Schema.optional(Schema.String),
       serverMode: Schema.optional(WorkerServerMode),
       spanCollectorUrl: Schema.optional(Schema.String),
+      /** Backing compiler workers owned by the data owner. */
+      compilationPoolSize: Schema.optional(Schema.Number),
```
kylewalke: 🟡 **LOW — dead `WorkerInit` wire fields.** `compilationPoolSize`/`compilationConcurrency` are added here and populated (`WorkerCoordinator.ts:384-393`) but the `WorkerInit` handler (`worker.platform.shared.ts:3133-3169`) never reads them — the worker instead reads these from `workerData` (`worker.platform.ts:429-445`). Two parallel channels; these schema fields are dead on the wire. Remove them or consume them in the handler.

knip.json
```diff
@@ -11,7 +11,9 @@
         "src/server.ts",
         "src/server/webWorkerServer.ts",
         "src/worker.platform.ts",
-        "src/worker.platform.web.ts"
+        "src/worker.platform.web.ts",
+        "test/fixtures/browserCompilationPoolHarness.ts",
```
kylewalke, line 17: 🟡 **LOW — knip will fail CI.** The two new production esbuild entrypoints `src/compiler.worker.node.ts` and `src/compiler.worker.web.ts` (`packages/apex-ls/esbuild.config.ts:135,154`) are referenced only by path string (never imported), so knip reports them as unused files. Add them to the `packages/apex-ls` `entry` list alongside `worker.platform.web.ts`.

packages/apex-ls/src/server/WorkerCoordinator.ts
```diff
@@ -217,33 +259,67 @@ export async function makeBrowserWorkerLayer(
...
 
-    // Transfer both port2s to the worker via rawWorker.postMessage on `self`.
-    // The worker listens on `self` for this one-time init message and never
-    // starts Effect's BrowserWorkerRunner on `self`, so there is no collision.
-    rawWorker.postMessage(
+  const [workerScript, compilerScript] = await Promise.all([
+    workerResponse.text(),
+    compilerResponse.text(),
+  ]);
+  const workerBlobUrl = URL.createObjectURL(
```
kylewalke, line 305: 🟡 **LOW — browser worker blob URLs never revoked.** `URL.createObjectURL` is called here (and at :299 for the compiler bundle) with no matching `URL.revokeObjectURL` anywhere. Bounded, per-server-lifetime leak (a handful of blob URLs), so minor — but worth revoking once the worker is spawned.

packages/lsp-compliant-services/src/services/LspRequestPrerequisiteMapping.ts
```diff
@@ -71,7 +71,10 @@ export function getPrerequisitesForLspRequestType(
         requiredDetailLevel: 'full', // Need full type info for hover
         requiresReferences: true,
         requiresReferenceResolution: true,
-        requiresCrossFileResolution: true,
+        // Hover resolves only the expression at the cursor. Materializing all
+        // cross-file edges in the document is reserved for reverse-index
+        // consumers such as references and implementation.
+        requiresCrossFileResolution: false,
```
kylewalke, line 77: 🟡 **LOW — hover cross-file prerequisite narrowed.** `requiresCrossFileResolution` went `true`→`false` for hover only (definition/signatureHelp/completion unchanged). Deliberate and documented, backstopped by still-enabled `missingArtifactResolution`. One residual risk: that fallback is `mode: 'background'` (non-blocking, ~:79), so the **first** hover over a symbol in an as-yet-unresolved file could return null/partial on a cold cache. Worth a manual smoke test.

kylewalke (commented):
## Automated review (Claude Code)

Reviewed across five subsystems (worker topology, deferred resolution, generics/FQN index, LSP request-prep, wire-schemas/settings/build). Overall solid — concurrency primitives (admission semaphore, scoped queues, worker-readiness gating) are correct and errors survive the worker boundary. I found **two release-blocking issues** plus a cluster of medium/low items, posted as inline comments.

**🔴 HIGH (should block merge)**
1. `compiler.worker.node.js` is never copied into the packaged extension `dist/`, so the default-on compilation pool fails to start in a shipped VSIX (passes local F5 dev).
2. Workspace-load session leaks on any compile failure → interactive cross-file resolution silently stops for the rest of the server's life.

**🟠 MEDIUM**: whole-batch data loss on a single-file failure; several dead configuration knobs; `Map` generic substitution only half-fixed (`values()`/`get()`/`put()`/`remove()` still return `Object`).

**🟡 LOW**: dead `WorkerInit` wire fields; knip missing new entrypoints (CI); browser blob URLs never revoked; hover cross-file prerequisite narrowed (deliberate, minor cold-cache first-hover risk).

See inline comments for specifics. Clean areas: permit accounting, pool-size clamps, Effect scope/finalizers, deferred-resolution queue logic, FQN-index staleness, webview XSS.

packages/apex-parser-ast/src/symbols/ApexSymbolManager.ts
```diff
@@ -5262,6 +5987,235 @@ export class ApexSymbolManager implements ISymbolManager, SymbolProvider {
...
+    const recordedWidth = range.endColumn - range.startColumn;
+
+    // Without the containing chain, accept only a token-sized reference.
+    // Collector variants sometimes attach a `this` reference to the member's
+    // range; treating that as the keyword is what made `this.field` resolve to
+    // the containing class.
+    if (!qualifiedReference && recordedWidth > name.length) return false;
+
+    return (
+      range.startLine === range.endLine &&
```
kylewalke: 🟠 **MEDIUM (new in this commit) — multi-line `this.`/`super.` chains break receiver-keyword resolution.**

`isPositionOnReceiverKeyword` prefers the *qualified chain* reference's `identifierRange` over the bare keyword's range, then requires `range.startLine === range.endLine` (this line). But chained references are built spanning `baseLocation.startLine` → `finalLocation.endLine` (`types/symbolReference.ts:612-617`), so a chain split across lines has `startLine !== endLine` and this returns `false`.

```apex
public class Foo {
  Integer count;
  void m() {
    this
      .count = 1;
  }
}
```
Cursor on `this` (line 4): the chain ref `this.count` spans lines 4-5.
- Via `getReceiverKeywordTargetAtPosition` (:6056) → returns `null`; hover/definition on `this` no longer resolves to the containing class.
- Via the `keywordAtPosition` guard (:6330) → the keyword isn't recognized, so the early `return null` is skipped and resolution falls through to member logic, producing a *wrong* result.

The parent commit used only the single-token keyword range (always single-line), so this is a regression. Single-line chains are unaffected. Consider matching the position against the keyword token's own range rather than the full chain range.

packages/apex-parser-ast/src/symbols/ops/chainResolution.ts
```diff
@@ -3344,21 +3508,37 @@ export async function resolveSymbolReferenceToSymbol(
     if (typeReference.context === ReferenceContext.VARIABLE_USAGE) {
       const candidates = await self.findSymbolByName(typeReference.name);
       const hasClassMatch = candidates.some(
-        (s) => s.kind === SymbolKind.Class || s.kind === SymbolKind.Interface,
+        (s) =>
+          s.kind === SymbolKind.Class ||
+          s.kind === SymbolKind.Interface ||
+          s.kind === SymbolKind.Enum,
```
kylewalke, line 3532: 🟡 **LOW-MEDIUM (new in this commit) — adding `Enum` to the `VARIABLE_USAGE` class-match lets an enum shadow a same-name local.**

For a `VARIABLE_USAGE` reference, if any same-named symbol is an `Enum`, the token is now treated as a type reference (`hasClassMatch` → `variableUsageLooksLikeClass`). If a local variable and an enum share the exact name, the variable usage resolves to the enum type instead of the local:
```apex
public enum Status { OPEN, CLOSED }
public class C {
  void m() {
    Status Status = null;   // local named identically to the enum
    System.debug(Status);   // VARIABLE_USAGE 'Status' now matches the enum
  }
}
```
Likelihood is low (requires exact-case collision, and `findSymbolByName` case-sensitivity limits it), and the other two Enum additions (:3541, :3554) are in class-reference contexts where broadening is correct. Flagging the direction of the change; a local/variable candidate should win over an enum type in `VARIABLE_USAGE` context.

kylewalke (commented):
## Follow-up review — commit `893c7ed7` (Claude Code)

Re-reviewed against the new commit `fix(ci): stabilize worker and web regression checks`. Note the commit message understates it — beyond CI/jest config it contains **behavioral symbol-resolution changes** to `this`/`super`/implicit-receiver handling (`ApexSymbolManager.ts`, +150) and adds `Enum` to type-candidate filters (`chainResolution.ts`).

**Status of prior findings — all still OPEN:**
- 🔴 HIGH #1 (`compiler.worker.node.js` not copied into packaged extension) — unchanged; esbuild still copies only `.web`. The new `browser-compilation` CI job validates the *browser* worker but nothing exercises the packaged *node* worker.
- 🔴 HIGH #2 (workspace-load session leak on compile failure) — unchanged; `catchAll` (WorkspaceBatchHandler.ts:513-521) still only `removeSession`, no drain/`ensuring`.
- 🟠 #3 (whole-batch data loss), #4 (dead config knobs), #5 (`Map.values/get/put/remove` still `Object`) — unchanged.
- 🟡 knip missing new entrypoints, dead `WorkerInit` wire fields, unrevoked blob URLs — unchanged.

**Good fixes in this commit:** `chainResolution.ts` now resolves `Enum` as a type in class/interface candidate filters; receiver-keyword handling is more precise for the common `this.field` single-line case; block symbols are skipped in the declaration-identifier scan.

**Two new issues introduced by this commit** (inline below): a multi-line `this.`/`super.` chain regression, and an enum-shadows-local edge case.

packages/apex-ls/src/worker.platform.web.ts
```diff
@@ -324,6 +378,93 @@ const WorkerLoggerLayer = Layer.merge(
...
+  );
+  const poolSize = Math.min(
+    requestedPoolSize,
+    Math.max(1, hardwareConcurrency - 2),
+  );
+  const concurrency = Math.max(1, Math.floor(data.compilationConcurrency ?? 1));
+  const workerPlatformUrl = data.workerPlatformUrl;
+  const W = globalThis.Worker;
+
+  const spawnCompilerWorker = () => {
```
kylewalke, line 425: 🟡 **LOW — nested browser compiler worker has no termination/cleanup finalizer.** `spawnCompilerWorker` creates `rawWorker` + two `MessageChannel`s and adds an `assistChannel.port1` message listener, but returns without any teardown hook. Unlike the coordinator's top-level path (`WorkerCoordinator.ts` `withBlobUrlCleanup`), there's no `rawWorker.terminate()` / port-close / listener-removal on scope close. Effect's `BrowserWorker` pool tears down via its own port protocol, but `rawWorker`, both channels, and the listener are held by closure with no explicit teardown — so repeated data-owner pool recreation (reconfig/restart) leaks web Workers + ports. The more real of the two lifecycle leaks. Consider a finalizer that terminates `rawWorker` and closes both ports.

packages/apex-ls/src/worker.platform.ts
```diff
@@ -369,6 +426,84 @@ const WorkerLoggerLayer = Layer.merge(
...
+const workerScript =
+  typeof __filename !== 'undefined'
+    ? __filename
+    : runtimeWorkerData?.workerScript;
+
+const spawnCompilerWorker = (): NodeWorkerThread => {
+  if (!workerScript) {
+    throw new Error('Worker script path is required to spawn compiler workers');
+  }
+  const compilerAssist = new MessageChannel();
```
kylewalke, line 461: 🟡 **LOW — Node compiler-worker assist channel leaks if spawn throws.** `compilerAssist` is created here, and `port1` is closed via `worker.once('exit', ...)`. But if `new NodeWorkerThread(...)` (next lines) throws — e.g. bad `workerScript` — no `exit` fires and neither port is closed, leaking that channel. Narrow (spawn-throw is a startup-fatal path, not steady-state), but a `try/catch` closing both ports on failure would close it.

packages/apex-ls/src/server/WorkspaceBatchHandler.ts
```diff
@@ -886,38 +749,7 @@ function processViaDataOwner(
...
-      yield* offer(Priority.Low, queuedItem);
-
-      logger.debug(
-        () =>
-          '[BATCH-PROCESSING] Enqueued batch ' +
-          `${batchParams.batchIndex + 1}/${batchParams.totalBatches} ` +
-          `(${batchParams.fileMetadata.length} files) for local processing`,
-      );
-    }
+    }).pipe(Effect.ensuring(Effect.orDie(closeSession)));
```
kylewalke, line 752: 🟡 **LOW — residual leak if the drain RPC itself rejects.** The session-end fix is correct for compile failures. But `closeSession` is wrapped in `Effect.orDie`, so if the `DrainDeferredReferences` dispatch *itself* rejects, it becomes a defect — and the top-level `Effect.catchAll` in `processStoredBatches` (~:513) recovers only the typed error channel, not defects. That defect would then bypass `removeSession`/`ingestionCompleteCallback` (~:521-522) and could leave the data-owner session active. Much narrower than the original bug (drain is now always *attempted*), but consider `catchAllCause` / a `catchAllDefect` around the finalizer so a failed drain still cleans up local state.

kylewalke (commented):
## Re-review — 4 new commits (`f6145efd`, `a3ee9c81`, `e69098c3`, `9c52aa14`) — Claude Code

Great progress — **every prior finding is now resolved.** Verified each against `9c52aa14`:

**🔴 HIGH — both FIXED**
- **#1 (node compiler worker not shipped)** — FIXED by the bundling refactor. The separate `compiler.worker.node.ts`/`.web.ts` entrypoints are deleted; the compiler worker now re-executes the common `worker.platform.js`/`.web.js` bundle with `workerData.role: 'compiler'` (`worker.platform.ts:465`). That bundle is already copied into the extension dist (`esbuild.config.ts:112-125`) and the runtime path is resolved from package root with an `fs.existsSync` hard-fail guard (`LCSAdapter.ts:2627-2648`) — no more dev-only path. Recursion is correctly bounded (compiler-role workers get `unavailableCompilationWorkerPool` and can't be messaged into spawning). The `--import tsx` execArgv branch is correct (prod `.js` → `[]`).
- **#2 (workspace-load session leak on failure)** — FIXED. The drain/end is now a finalizer: `.pipe(Effect.ensuring(Effect.orDie(closeSession)))` (`WorkspaceBatchHandler.ts:752`), so the session ends on success, failure, and interruption.

**🟠 MEDIUM — all FIXED**
- **#3 (whole-batch data loss)** — FIXED. `WorkspaceCompilationPipeline.ts` now isolates per-file failures into a `failures[]` array and continues; the worker surfaces them via `resolveReadiness`/`logWarning` (`worker.platform.shared.ts:4008-4013`).
- **#4 (dead config knobs)** — FIXED. `lspRequest.concurrency`, `dataOwner.poolSize`, `resourceLoader.poolSize`/`concurrency` removed from the contribution.
- **#5 (`Map` generics half-fixed)** — FIXED. `get`/`put`/`remove`/`values` now use `K`/`V` (and `remove(Key)` → `remove(K key)`).

**🟡 LOW — all FIXED**: dead `WorkerInit` wire fields removed; knip entrypoints moot (files deleted); browser blob URLs now revoked via `withBlobUrlCleanup` finalizer. Both new-commit findings from my last review also fixed: multi-line `this.`/`super.` chains (the `startLine === endLine` guard was dropped), and enum-shadows-local (`chainResolution.ts:3507` now resolves a scoped Variable/Parameter/Field/Property before the enum candidate check).

**Three new LOW residuals introduced by the refactor** (inline below) — none blocking. Nice work; from my side the two release-blockers are cleared.

kylewalke (dismissed):
## Approved ✅

All findings from my earlier reviews are resolved as of `9c52aa14` — including both release-blockers:
- 🔴 HIGH #1 (compiler worker now ships in the packaged bundle via the common `worker.platform.js` restructure)
- 🔴 HIGH #2 (workspace-load session now ended via `Effect.ensuring`)

Also fixed: whole-batch data-loss on a single bad file, dead config knobs, `Map` generic substitution, the dead `WorkerInit` wire fields, blob-URL cleanup, and both resolution issues from the intermediate commit (multi-line `this.` chains, enum-shadows-local).

Three **LOW** residuals remain (noted inline on the prior review) — nested browser worker teardown, a Node MessageChannel leak on spawn-throw, and a defect-bypass if the drain RPC itself rejects. None block merge; worth a follow-up.

Nice work on the worker topology and the deferred-resolution rework.

kylewalke (approved):

PR #590, opened 2026-07-27
feat(apex-lsp-parser-ast): infer expression types for Extract code actions - W-23389334

kylewalke:
### What does this PR do?

Replaces the `Object` placeholder in the Extract Variable / Extract Constant code actions with the **inferred Apex type** of the extracted expression (Jorje-parity story 01.1).

**Type inference** — new `inferExpressionType` in `apex-parser-ast`:
- Leaf cases resolved directly (casing preserved): literals (`String`/`Integer`/`Long`/`Decimal`/`Boolean`), `new T(...)` → `T`, casts `(T) x` → `T`, same-file variable/parameter/field references, and unqualified same-file method-call return types.
- Composite cases (arithmetic promotion, string concat, comparison → `Boolean`, ternary, unary, `instanceof`) delegate to the shared recursive resolver, normalized back to canonical Apex casing.
- Same-file only in this cut; cross-file resolution enters through an injected `lookupCrossFileType` seam. Never throws — resolution failures degrade to `null`, and the caller falls back to `Object` (no more `// TODO`).

**Wiring** — `CodeActionProcessingService` computes the inferred type and emits it in both the Extract Variable and Extract Constant declarations. Parse failures now log at `debug` in the LS caller (apex-parser-ast stays logger-free).

**Extract Constant indent fix** — the inserted constant now matches the file's actual member indentation (measured from the enclosing member's physical line indent) instead of a hardcoded two-space unit, which had under-indented it by two spaces in four-space files.

**Tests** — unit coverage for inference and indentation; e2e specs (`apex-extract-variable`, new `apex-extract-constant`) assert the inferred types (`Integer`, `String`, `Account`) and the four-space constant indent end-to-end through the editor. 4/4 e2e pass locally in web mode.

@W-23389334@

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -35,6 +35,8 @@ jobs:
         test_file:
           - apex-capability-gating.spec.ts
           - apex-extension-core.spec.ts
+          - apex-extract-constant.spec.ts
+          - apex-extract-variable.spec.ts
           - apex-goto-definition.spec.ts
           - apex-hover.spec.ts
           - apex-hover-stdlib.spec.ts
@@ -275,6 +277,8 @@ jobs:
           EXPECTED_SPECS=(
             "apex-capability-gating.spec.ts"
             "apex-extension-core.spec.ts"
+            "apex-extract-constant.spec.ts"
+            "apex-extract-variable.spec.ts"
             "apex-goto-definition.spec.ts"
             "apex-hover.spec.ts"
             "apex-hover-stdlib.spec.ts"
--- a/e2e-tests/test-data/apex-samples/ExtractConstantSample.cls
+++ b/e2e-tests/test-data/apex-samples/ExtractConstantSample.cls
@@ -1,0 +1,10 @@
+public with sharing class ExtractConstantSample {
+    // A method with a string-literal sub-expression. Extract Constant is gated
+    // (Jorje parity) to literal expressions, and `'greeting'` is a literal, so
+    // the "Extract constant" action is offered here and the extracted constant
+    // infers to String.
+    public String greet() {
+        String message = 'greeting';
+        return message;
+    }
+}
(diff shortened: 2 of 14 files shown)


peternhale (approved):
**Approve.** Correctness-focused, safely-degrading, and well-tested. The scope-aware resolution that closes the sibling-method shadowing hole — on both the leaf path (`resolveVariableAtPosition`) and the composite path (`collectVariableTypes` pre-seeding the shared resolver's cache) — is the standout. The arity-based overload disambiguation that bails to `null` (safe `Object` fallback) rather than guessing is also well done, and "never throws" holds up: `resolveExpressionTypeRecursive` is typed `Effect<…, never, never>` so `Effect.runSync` can't throw, and it's still wrapped in try/catch at both the module and service layers.

None of the items below block the merge. Tagging each with a suggested disposition so they can be triaged.

### Suggested follow-up (next PR)

- **`sameFileMethodReturnType` — single-candidate case skips arity and is file-wide.** When exactly one same-named method exists in `getAllSymbols()` (file-wide, so inner-class methods are included), it returns that return type without confirming the unqualified call could actually reach it. For a bare `foo()` this is almost always the enclosing class's method, and a real outer/inner collision produces ≥2 candidates and bails — so the downside is cosmetic (a wrong-but-compiling type the user edits). Worth a scoping comment now; a reachability check is follow-up material.

- **`recoverTypeCasing` scans file-wide, first-match-wins.** If two same-file declarations differ only in casing, recovery may pick either. Apex type names are case-insensitive so the result still compiles — purely cosmetic. Follow-up if ever observed.

- **`collectLiteralTypes` / `collectVariableTypes` use `any`-typed tree walks** (`node: any`, `getChild`). Contained and pragmatic given ANTLR's typing, but out of step with the otherwise strongly-typed module. Low-priority cleanup.

### Consider addressing now (cheap)

- **`resolveVariableAtPosition` column-basis coupling.** It compares the incoming 0-based `column` against `symbolRange.startColumn/endColumn`, which is only correct while `symbolRange` columns are also 0-based. Tests pass, so they're consistent today, but the coupling is undocumented — a future change to `symbolRange`'s column basis would silently break boundary matches. A one-line comment of intent (and ideally a boundary test at exactly `startColumn`) would lock this in. Cheap enough to do in this PR.

### Not an issue, just noting

- The `expressionRangeFinder.ts` indent fix (measure a real sibling member vs. a fixed 2-space unit) is logically independent of type inference, but it's small, well-commented, has matching regression tests (2-space / 4-space / inner-class), and supports the same code-action output — bundling is fine.

PR #591, opened 2026-07-28
fix: pass custom JS heap to data owner worker via resourceLimits - W-23610333

peternhale:
### What does this PR do?

Fixes the Apex Language Server failing to start whenever a user sets a custom JS heap size (`apex.environment.jsHeapSizeGB`).

**Root cause:** the extension adds `--max-old-space-size=<MB>` to the main server process, and the worker exec-argv builder copied that flag into every worker's `execArgv`. But `--max-old-space-size` is **not** a legal `worker_threads` execArgv flag — `new Worker()` throws `Error: Initiated Worker with invalid execArgv flags`. The first worker spawn aborted topology initialization, taking down the entire worker pool and the server.

**Fix:** stop forwarding `--max-old-space-size` into worker `execArgv`. Parse the MB value and pass it through the Worker constructor's `resourceLimits.maxOldGenerationSizeMb` instead — the correct mechanism for a worker's heap. The heap limit is scoped to the **data owner worker only** (`HEAP_LIMITED_ROLES`), the sole worker whose heap scales with workspace size (it holds the full symbol graph). Enrichment, resource loader, and compilation workers keep the default heap, avoiding a multi-worker memory over-commit.

Verified locally that `execArgv: ['--max-old-space-size=N']` is rejected by the Worker constructor while `resourceLimits: { maxOldGenerationSizeMb: N }` is accepted (data owner comes up at ~12 GB with the reported `12288` value; other roles stay at the ~4 GB default).

#7910, @W-23610333@

### Functionality Before

Setting `apex.environment.jsHeapSizeGB` to any non-zero value crashed the language server on startup:

```
cause="Error: Initiated Worker with invalid execArgv flags: --max-old-space-size=12288
    at new Worker (node:internal/worker:225:13)"
```

### Functionality After

The language server starts normally with a custom heap size. The data owner worker receives the configured heap; other workers keep the default.

### DIFF
--- a/packages/apex-ls/src/server/LCSAdapter.ts
+++ b/packages/apex-ls/src/server/LCSAdapter.ts
@@ -2650,16 +2650,26 @@ export class LCSAdapter {
         this.logger.alwaysLog(
           () => `[WorkerCoordinator] Worker script (node): ${workerScript}`,
         );
-        const workerLayerFactory = (role: string) =>
-          makeNodeWorkerLayer(workerScript, {
+        const workerLayerFactory = (role: string) => {
+          const { execArgv, maxOldGenerationSizeMb } = buildWorkerExecArgv({
+            role,
+          });
+          return makeNodeWorkerLayer(workerScript, {
             name: `apex-worker-${role}`,
-            execArgv: buildWorkerExecArgv({ role }).execArgv,
+            execArgv,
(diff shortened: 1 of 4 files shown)


peternhale:
Addresses issue [7910](https://github.com/forcedotcom/salesforcedx-vscode/issues/7910)

kylewalke (approved):

PR #593, opened 2026-07-29
docs: add textDocument/rename spike plan - W-22629631

kylewalke:
## Summary

Deliverable of **W-22629631**, a planning spike for `textDocument/rename` (Rename Symbol) in the Apex Language Server. This PR adds the spike's output — a phased implementation plan — to `docs/plans/`. No product code changes.

- Phased plan (Phase 0 wiring → Phase 5 tests), grounded in a fresh code survey and revised after an adversarial plan review that corrected three wrong assumptions (case-sensitive lexical prefilter, missing scope-binding in local occurrence matching, and a reverse hierarchy index that already exists).
- The 20-item epic work breakdown, sequenced per the work-item-sequencing convention.

## Context

Implementation is tracked in the dedicated epic **Apex LS: textDocument/rename (Symbol Rename)** (`a3QEE000002YjCr2AK`) and its 20 child work items (W-23631069 … W-23631152). This spike WI remains under `IDE Apex Language Support - Jorje Parity`; its only deliverable is this document.

@W-22629631@

## Test plan

Docs-only change — no runtime surface to exercise. Markdown renders correctly and internal skill link resolves from `docs/plans/`.


peternhale (approved):
Approving. This is a spike deliverable (a planning doc), and it's in good shape.

The plan went through adversarial review against the live codebase and a second pass against the jorje reference (`apex-jorje-lsp/.../impl/rename/`) + LSP 3.17. The two decisions carried forward — qualify-on-conflict and the 5.1→4.0 resequencing — were both enacted in `8c1f63d3` and are internally consistent:

- **Qualify-on-conflict → replicate jorje.** Threaded through 4.1 / 5.2 and reconciled with the error-delivery decision (a shadow feeds a fully-qualified rewrite, not a `ResponseError`).
- **Resequencing → option (a), renumber 5.1→4.0.** Verified against the work-item-sequencing convention: `4.0` as a whole-number gates all of `4.x` and gates on all of Group 3, so renameField's dependency on the hierarchy graph is now expressed structurally by the dotted prefix. The `5.1` gap is intentional and harmless.

**One non-blocking follow-up before the WIs start building** (not a doc change): the plan's resequencing note both claims the GUS `Subject__c` edit on **W-23631128** is done *and* lists it as an open action item. The auto-build gate keys off the actual `Subject__c` prefix in GUS, not the doc — if W-23631128 still reads `5.1 …`, renameField (4.2) will start without the hierarchy graph, which is the exact failure the renumber was meant to prevent. Please confirm W-23631128's prefix is `4.0 ` in GUS and drop whichever of the two contradictory sentences is wrong. I couldn't verify from my environment (no default GUS org configured).

PR #594, opened 2026-07-30
docs: fold jorje-parity review into rename spike plan - W-22629631

peternhale:
### What does this PR do?

Stacked on top of #593. Folds a second review pass of the `textDocument/rename` spike plan into `docs/plans/W-22629631-rename-symbol-spike-plan.md`. The first pass was adversarial against the live codebase; this pass reviews the plan against the jorje reference implementation (`apex-jorje-lsp/src/main/java/apex/jorje/lsp/impl/rename/`) and LSP 3.17.

Adds a **Cross-cutting decisions from the jorje-parity pass** section and threads the consequences into the affected phases and the WI breakdown:

- **Qualify-on-conflict parity** — jorje rewrites a shadowed reference to a fully-qualified name (`OuterType.newName` / `this.newName`) rather than rejecting the rename. The plan previously framed conflict handling as detect-then-error only. Flagged as a behavioral decision affecting Groups 3–5.
- **Constructor → renameType dispatch** — a cursor on a constructor resolves to the type-rename path in jorje, not method-rename. Added to the shared resolver's kind-mapping (1.3 / 3.1).
- **Sequencing fix** — renameField descendant-conflict detection needs the `findSubtypes` hierarchy graph currently scoped to 5.1 (Group 5), so Group 4 depends on a Group-5 deliverable. Resequencing note + options added.
- **WorkspaceEdit shape + capability gates** — `documentChanges` vs `resourceOperations` are separate client capabilities; decision moved to Phase 0 instead of surfacing only at renameType (Group 6).
- **Error delivery** — pinned to `ResponseError` over jorje's empty-`WorkspaceEdit` + `window/showMessage` side channel.
- **prepareRename ordering** — the `prepareProvider` capability flag must not be advertised before `onPrepareRename` exists; noted that prepareRename is beyond-parity (jorje has none).
- Corrected verified line references (`textMentionsSymbol.ts:26`, `symbol.ts:492`, `ApexSymbolRefManager` `findSubtypes`/`findSupertypes` at 2485/2500).

Docs-only; no functional code change.

@W-22629631@

### DIFF
--- a/docs/plans/W-22629631-rename-symbol-spike-plan.md
+++ b/docs/plans/W-22629631-rename-symbol-spike-plan.md
@@ -13,7 +13,15 @@
 - Rename is currently routed `coordinatorOnly` in `WorkerCoordinator.ts` `DISPATCH_ROUTING` (~line 752), but the reference-finding it depends on runs on the request pool via a two-phase flow (coordinator lexical-prefilter → per-file standalone parse+scan in `worker.platform.shared.ts`). Decision: move `rename` to `requestPool` to reuse that flow directly.
 - Absorbs the rename-implementation portion of the former W-22629618 (AST/symbol-table semantics), closed/redistributed.
 
-This plan went through one round of adversarial review (plan-adversary agent) against the live codebase. Three of the original assumptions were wrong; corrections are folded into the phase plan below and flagged inline.
+This plan went through one round of adversarial review (plan-adversary agent) against the live codebase, then a second review against the jorje reference implementation (`~/git/apex-jorje/apex-jorje-lsp/src/main/java/apex/jorje/lsp/impl/rename/`) and LSP 3.17. Three of the original assumptions were wrong; corrections are folded into the phase plan below and flagged inline. The jorje pass surfaced several parity gaps — qualify-on-conflict rewriting, constructor→type dispatch, hierarchy-graph sequencing, prepareRename/capability ordering, and error-delivery mechanism — captured as cross-cutting decisions and folded into the affected phases.
+
+### Cross-cutting decisions from the jorje-parity pass
+
+- **Qualify-on-conflict (behavioral parity choice — affects Groups 3–5).** jorje does *not* only detect member-name conflicts and reject: when a rename would shadow an inner-class or ancestor static member, it **rewrites the reference to be fully qualified** (`this.newName` for instance, `OuterType.newName` for static) instead of erroring. See `RenameUtil.fieldNameRewrites`/`getFullyQualifiedName` and the inner-class conflict blocks in `FieldRenameHandler.getDocumentChanges` / `MethodRenameHandler.getDocumentChanges`. **Decision needed:** replicate jorje's qualify-on-conflict, or intentionally reject those cases. If we reject, that is a deliberate divergence from jorje and must be documented as such (this epic is "Jorje Parity"). Default recommendation: replicate, since rejecting turns a mechanical rename into a manual fix.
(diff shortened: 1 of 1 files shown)


peternhale:
## Action needed: resequencing 5.1 → Group 4 (hierarchy graph is a renameField dependency)

Flagging this for a decision before implementation starts, since the numeric-sequencing gate can't express it as-numbered.

**The problem.** renameField descendant-conflict detection (4.2) needs the `findSubtypes` hierarchy graph. In jorje, `FieldRenameHandler.getDescendantTypeHavingFieldConflict` walks the subtype graph for descendant conflicts exactly as `MethodRenameHandler` does — descendant traversal is not a renameMethod-only concern. But the plan scopes the hierarchy-traversal + graph-completeness-guard WI to **5.1** (Group 5). So Group 4 depends on a Group-5 deliverable.

The [work-item-sequencing](.claude/skills/work-item-sequencing/SKILL.md) gate only holds whole-number-N on all of N−1; it cannot express "4.2 depends on 5.1" from the dotted prefix. Left as-is, the auto-build gate would let 4.2 start before 5.1 merges, and renameField descendant conflicts would ship unguarded (or the WI would block on missing capability at build time).

**Options:**

- **(a) Renumber** 5.1 into Group 3/4 (e.g. a new `3.x` or `4.0`) so the dotted prefix reflects the true order. Requires updating the GUS `Subject__c` prefixes **and** the W-number map on line 60 of the plan, since the auto-build gate keys off the prefix. W-number to move: **W-23631128** (5.1).
- **(b) Keep the numbers**, add `Depends on (must merge first): W-23631128` to 4.2's (**W-23631076**) Details so the blocker gate holds 4.2 until 5.1 merges. Cheaper, no renumber, but the dotted order stays cosmetically misleading.

**Recommendation:** (b) for now; (a) if the epic is re-planned. Either way, **5.1 must merge before 4.2.**

(Details also folded into the plan's resequencing note and the 4.2 / 5.1 rows.)

peternhale:
## Decision needed: qualify-on-conflict behavior (parity vs. reject)

This is a genuine product/behavioral decision, not just a plan gap — it drives the conflict-handling design across Groups 4–5, so it should be settled before renameField (Group 4) starts.

**What jorje does.** jorje does *not* only detect a member-name conflict and reject the rename. When a rename would shadow an inner-class or ancestor static member, jorje **rewrites the conflicting reference to be fully qualified** and proceeds:

- instance member → `this.newName`
- static member → `OuterType.newName`

See `RenameUtil.fieldNameRewrites` / `getFullyQualifiedName`, and the inner-class conflict blocks in `FieldRenameHandler.getDocumentChanges` and `MethodRenameHandler.getDocumentChanges` (`apex-jorje-lsp/src/main/java/apex/jorje/lsp/impl/rename/`).

**The choice:**

- **Replicate** jorje's qualify-on-conflict — the rename succeeds and stays correct by qualifying shadowed references. More edit-construction work; true parity.
- **Reject** those cases with a conflict error — simpler, but turns a mechanical rename into a manual fix, and is a **deliberate divergence** from jorje that this epic ("IDE Apex Language Support - Jorje Parity") should explicitly document.

**Recommendation:** replicate, since rejecting degrades the feature exactly where users most want the tool to do the tedious work. But calling it out here so it's an explicit decision rather than an implementation-time default.

Whichever way this lands, it affects 4.1/4.2 (field) and 5.2/5.3 (method) — the WorkspaceEdit construction (4.1/5.2) has to emit the qualified-rewrite edits if we replicate.

PR #596, opened 2026-07-30
feat: wire textDocument/rename to request pool (no-op) - W-23631069

kylewalke:
### What does this PR do?

Phase 0 of the rename-symbol spike: routes `textDocument/rename` off the coordinator onto the request-worker pool, mirroring the existing `DispatchReferences` path. The pipe is wired end-to-end but the pool-side handler is a deliberate **no-op** — it returns `null` — so the full request path is verifiable before any occurrence-resolution / `WorkspaceEdit` logic lands in later groups.

- New `DispatchRename` wire message (position + `newName` + threaded content)
- `DISPATCH_ROUTING` flips `rename` from `coordinatorOnly` → `requestPool`, plus a message-builder case
- No-op `DispatchRename` worker handler (returns `null`)
- `onRenameRequest` LSP handler gated on `renameProvider`, forwarding to `submitRenameRequest`
- `renameProvider: true` in `DEVELOPMENT_CAPABILITIES` only (no `prepareProvider` yet — that waits for Phase 4) + advertised in the `initialize` static capabilities so the F2 Rename UI appears
- New `RenameThroughWorkerTopology` integration test asserting the round-trip returns `null` across the worker boundary

@W-23631069@


peternhale (approved):
## Approval

Excellent Phase 0 foundation work. The pipe is wired end-to-end with proper integration testing before any rename logic lands.

**Highlights:**
- Clean routing change (coordinatorOnly → requestPool) with clear rationale
- Comprehensive integration test proves round-trip across worker boundary
- Proper schema + cancellation token threading
- Mirrors DispatchReferences pattern consistently

**Minor note:** F2 Rename UI will appear in dev mode but return null until later phases - this is acceptable for the phased approach.

LGTM ✅

PR #597, opened 2026-07-30
refactor: extract shared cursor-occurrence resolver from references - W-23631073

kylewalke:
> **Stacked on #596** (W-23631069). Base branch is `worktree-W-23631069-dispatch-rename`; review/merge #596 first. The diff against `main` will collapse to just this refactor once #596 lands.

### What does this PR do?

Phase 1 of the rename-symbol spike: factors the kind-agnostic **cursor → occurrences** core out of the `DispatchReferences` worker handler into a shared `resolveOccurrencesForCursor` helper, so the upcoming rename handler can reuse it instead of duplicating the two-phase scan.

The extracted core runs the four stages both features need:
1. recompile the cursor file at full detail (so in-body usages resolve)
2. resolve the cursor position → target symbol name + kind
3. phase-1 lexical prefilter of the workspace (`FindOccurrenceCandidates`) + live-buffer content override
4. phase-2 standalone parse + reference scan (`scanCandidatesForOccurrences`)

**Pure refactor — references behavior is unchanged.** The handler keeps its own `Location[]` assembly and `includeDeclaration` branch; only the shared stages moved. Also names the previously-duplicated `{ uri, identifierRange }` shape (`OccurrenceRange` / `CursorOccurrence`), and re-exports the helper from both platform entry files.

Explicitly out of scope (separate stories): the case-insensitive prefilter fix (2.1) and the rename handler actually calling this helper (Groups 3+).

### Regression guard

`ReferencesThroughWorkerTopology` + `EnrichmentRoundTrip` (12 tests) stay green; full `apex-ls` suite passes (290 tests).

@W-23631073@

### DIFF
--- a/packages/apex-ls/src/worker.platform.shared.ts
+++ b/packages/apex-ls/src/worker.platform.shared.ts
@@ -2420,6 +2420,25 @@ function emitWorkerLog(level: string, message: string): void {
 // ---------------------------------------------------------------------------
 
 /**
+ * Identifier coordinates in parser space (1-based line, 0-based column).
+ */
+export type OccurrenceRange = {
+  startLine: number;
+  startColumn: number;
+  endLine: number;
+  endColumn: number;
+};
+
+/**
+ * A single occurrence of a symbol: the file URI plus the matched identifier's
+ * range (parser coordinates). Shared by find-references and rename.
+ */
+export type CursorOccurrence = {
+  uri: string;
+  identifierRange: OccurrenceRange;
+};
+
+/**
  * Phase-2 of find-references (W-23272674, standalone-scan pivot): given the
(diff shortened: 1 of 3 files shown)


peternhale (dismissed):
## Approval

Clean refactor that extracts the shared cursor→occurrences core without changing references behavior.

**Highlights:**
- Excellent abstraction boundaries (4-stage pipeline, caller-specific assembly)
- Type safety improvements (OccurrenceRange, CursorOccurrence)
- Zero behavior change - validated by existing 290-test suite
- Sets up Phase 2+ rename implementation to reuse this logic

**Structural duplication note:** DeclarationLocation and CursorOccurrence are structurally identical - consider unifying later if they serve the same semantic purpose.

LGTM ✅

peternhale (approved):

PR #608, opened 2026-08-03
refactor: make Apex semantic resolution parser-owned - W-21360368

peternhale:
### What does this PR do?

Makes Apex semantic resolution parser-owned across the parser, symbol manager,
language services, and worker topology.

- Replaces raw-source and composite-context semantic inference with lexer,
  parser, symbol-table, and xref facts.
- Adds structured semantic context for DML operands, invocations, incomplete
  member access, indexed access, and active-edit snapshots.
- Resolves completion, hover, definition, signature help, code actions, and
  chained expressions through exact semantic identity.
- Requires missing-artifact requests to carry source URI, document version,
  exact parser range, stable reference identity, and parse completeness.
- Records bounded semantic snapshot/provenance attributes in development spans.
- Adds an ESLint guardrail that prevents semantic fallbacks to raw Apex text.
- Repairs Jest transform/teardown configuration so the full semantic suites run
  without the former 4 GB transform-time heap exhaustion.

The SObject workspace/org discovery implementation is intentionally excluded.
It will be proposed separately after these semantic foundations are reviewed.

No linked GitHub issue.

### Functionality Before

Incomplete edits and chained expressions could lose or misidentify semantic
state. Services sometimes recovered names and types from source text, selected
the first same-named global symbol, or issued name-only missing-artifact
requests. Results could vary between initial load and active editing.

### Functionality After

Semantic features consume current parser-owned references, declarations, type
information, and xref identity. Missing or incomplete state remains explicitly
uncertain and retryable rather than being guessed. Active-edit first-dot member
completion, multi-stage chains such as `a.b.c().d`, exact hover/definition, and
missing-artifact load/retry lifecycles have automated regression coverage.

### Review follow-up

- Field/member lookup now preserves explicit `this`/`super` receiver semantics
  and filters same-named candidates by member kind.
- The field-shadowing regression is covered at the usage site in both the
  compliant-services integration layer and the real request-worker topology.
- The ambiguous Web E2E cursor placement was corrected and the shared-workspace
  definition spec is deterministic locally and in CI.
- Missing-artifact batches retain valid identifiers when another entry is
  malformed.
- Development tracing is failure-safe, bounded, valid JSON, and cleaned up on
  document close.
- Cross-file preparation avoids redundant work through unresolved-reference
  checks and per-table single-flight resolution.
- Collection, return-type, qualified-name, code-action, wire-compatibility,
  diagnostic-wiring, localization, and ESLint guardrail review findings are
  covered by implementation changes and regression tests.

### Verification

- Repository commit and push hooks completed normally.
- 392 test suites exercised.
- 5,032 tests passed, 0 failed, 484 pending/skipped.
- Parser package: 3,034 passed.
- Compliant services: 818 passed.
- Apex LS: 314 passed.
- Shared settings/protocol: 488 passed.
- VS Code extension: 216 passed.
- Parser, compliant-services, shared, and Apex-LS typechecks pass.
- Parser-owned semantic guardrail: 13/13 passed.
- Web go-to-definition: 29/29 passed, including same-file, sequential, and
  cross-file cases.

@W-21360368@


kylewalke (changes requested):
The architectural direction is sound and well-executed — the "preserve uncertainty, don't guess" pattern is applied consistently, dead text-scraping helpers are genuinely removed (not just orphaned), and null/absent-fact handling largely degrades gracefully rather than
  crashing. But there are merge-blocking issues, starting with a goto-definition regression I reproduced locally that the test suite structurally cannot catch.

  🔴 Blockers

  1. Deterministic goto-definition regression — this.field resolves to a shadowing parameter

  CI is red on E2E apex-goto-definition.spec.ts (Web), and I reproduced it locally by building the PR branch in a clean worktree: should navigate to field definition fails 100% of runs (3/3 isolated, plus every full-spec run) with:

  Error: Expected cursor at line 7 but cursor was at line 20

  Cursor on this.instanceId (fixture line 24) resolves to the constructor parameter ApexClassExample(String instanceId) (line 20) instead of the field private String instanceId; (line 7). The this.-qualified member's "must be an instance member" constraint is lost and
  scope shadowing wins. Culprit path: ApexSymbolManager.getSymbolAtPositionPrecise / resolveInstanceReceiverMemberAtPosition (~L6116–6461) — the scope-visible-value shortcut returns the parameter. This directly contradicts the PR's headline claim of "exact
  hover/definition through exact semantic identity": a field shadowed by a same-named param, resolved via this., is precisely the case that regressed.

  2. Second, intermittent cross-file resolution failure — separate root cause

  The full spec also produces a second failure that is order/session-dependent, not deterministic. Across my local runs it landed on different tests: CI hit the sequential-ops inner-enum navigation; one local full run failed should navigate to base class defined in 
  another workspace file (landed on the child-class file); another local full run had no second failure (28 passed). Isolated, the sequential-ops test passes. This smells like accumulated worker/session state or a cross-file resolution race — consistent with the flagged
  uncancellable per-keystroke resolveCrossFileReferencesForFile on the hot path (see High/Medium below) — not the this.-qualifier bug. Treat it as a distinct flaky-resolution defect; it must be green before merge but should be triaged separately.

  3. Test coverage gap is why the deterministic bug shipped green — add usage-site coverage

  The author knew about the param-shadowing hazard: DefinitionFieldParamDuplicate.integration.test.ts (W-23408848) uses the exact ApexClassExample/instanceId shape, including a layered public-api + full recompile case. But every node test clicks the field declaration, 
  never the this.instanceId usage on line 24 — the position that actually regresses. Combined with Desktop E2E being skipped in this run (goto-def validated web-only), the broken path had no gate. That's how the Jest suites report "5,019 passed, 0 failed" while E2E goes
  red.
  - Fix: add node-level coverage that positions on the usage site — this.instanceId asserting resolution to line 7 (the field, not the parameter). That would have caught the deterministic failure before E2E.

  4. PR body work-item ref is malformed — pr-validation fails

  The body has @ W-21360368@ (stray space), which fails the required @W-\d{7,8}@ pattern. Also correct the "0 failed" claim. Fix the body.

  5. All-or-nothing missing-artifact batch rejection — utils/missingArtifactProvenance.ts:37-46

  sanitizeMissingArtifactParams returns null on the first identifier that fails decode/validation, discarding the whole batch. PrerequisiteOrchestrationService builds one multi-identifier batch per file, so a single malformed range or provenance.sourceUri !== origin.uri
  sinks resolution for every legitimately-missing type in that file. Fix: filter invalid identifiers, proceed with survivors, return null only when none remain (the existing length === 0 guard shows that was the intent).
  
  6. Jest globalTeardown: undefined + forceExit: false risks CI hangs — apex-parser-ast/jest.config.cjs, lsp-compliant-services/jest.config.cjs
  
  The removed global teardown was the only place shutting down the scheduler, ApexSymbolRefManager daemon fibers (Effect.forkDaemon), and LSPQueueManager's setInterval singletons. Only ~34/229 (parser) and ~7/68 (services) suites self-clean via afterAll. Previously
  forceExit: true masked leaked handles; flipping both switches at once trades a deterministic heap error for a potential process hang on any leaked fiber/interval. Fix: verify a full run of both packages exits cleanly, or keep forceExit env-gated until per-suite cleanup
  exists. (The transform-narrowing heap fix itself is sound and does not disable coverage.)
  
  🟠 High
  
  - Observability can crash the user request — worker.platform.shared.ts:3229/3392/3428. Capture runs via Effect.promise(() => captureResolutionStateAttributes(...)); Effect.promise turns a rejection into a fiber-killing defect. If
  getSymbolTableForFile/getReferencesAtPosition/JSON.stringify throws, the hover/completion/definition request fails instead of skipping tracing. It's dev-gated (__apexWorkerInitServerMode === 'development') — but that's the debugging session where it bites. Fix:
  Effect.tryPromise({ try, catch: () => ({}) }).
  - frozenSnapshots grows unbounded — SymbolStateFlightRecorder.ts:205/389. Every anomaly stores a snapshot under a unique id; the only reclaim path, resetSymbolStateFlightRecorder() (:403), has no callers (nor does getFrozenSymbolStateSnapshot). Dev-mode leak over a
  long session. Fix: bound with LRU/max-entries and wire reset to a lifecycle hook (or drop the dead exports). (Related MEDIUM: histories/generations never pruned on didClose — no such handler exists.)
  - Uncancellable per-keystroke cross-file re-resolve on the hot path — worker.platform.shared.ts:1737. resolveCrossFileReferencesForFile(uri) runs unconditionally on every hover/completion prepare. Given LSP cancellation tokens never reach workers, this is uncancellable
  added latency, and is the leading suspect for the intermittent cross-file failure in Blocker #2. Confirm the re-entrancy guard makes repeats cheap, or debounce upstream.
  
  🟡 Medium (correctness / behavior narrowing)

  - new List<T>(mySet) now false-positives — CollectionValidator.ts (~L494). List-from-Set is valid Apex but matches neither the Integer nor List branch → spurious INVALID_LIST_INITIALIZER. Add a set branch mirroring the Set path.
  - Method return-type special cases dropped — MethodResolutionValidator.ts (~L1211). areReturnTypesCompatible (Id↔String, numeric widening, Object-as-source) deleted in favor of bare isAssignable. Confirm isAssignable encodes those three, else new false-positive
  METHOD_DOES_NOT_SUPPORT_RETURN_TYPE.
  - Phantom 'unknown' references still emitted — ApexReferenceCollectorListener.ts:2711 (extractQualifierFromExpression) returns literal 'unknown' and processStandaloneMethodCall creates a VARIABLE_USAGE from it, unguarded. The sibling BlockContentListener was fixed to
  return undefined + early-return; this listener was not. Contradicts the PR's own thesis (and the ESLint rule's intent).
  - DOT handling is O(dots × references) on the well-formed path — ApexSymbolCollectorListener.ts (~L399), BlockContentListener.ts (~L2297). For normal a.b.c, every . falls through to getAllReferences() (full array copy) + filter/find/sort. Gate the recovery scan behind
  an actual "in recovery / pending receiver" check, or index refs by end position.
  - bindResolvedDeclarationType is O(typeRefs × symbols) — ApexSymbolManager.ts (~L4760). Iterates getAllSymbols() per resolved typeRef. Precompute a Map<typeReferenceId, symbol> once.
  - hasCompleteSemanticState can permanently suppress hover/definition — CompletionProcessingService.ts:275, HoverProcessingService.ts:160, DefinitionProcessingService.ts:159. When the table can't place the cursor (scopes.length === 0 && !sourceSymbol), completion
  returns retryable isIncomplete, but hover returns null and definition [] with no retry hint. At top-of-file / between-declaration positions this is silently empty vs. the old text fallback. Confirm the parser always yields a scope at class-body/file-level positions.
  - Hardcoded, hand-built artifact URI — MemberAccessCompletionStrategy.ts:352. 'apex-org-artifact:/apex-class/' + ns.toLowerCase()+'.'+type.toLowerCase()+'.cls' hardcodes scheme/casing/path and violates the vscode-uri convention; any mismatch silently resolves to null.
  Extract a shared helper owned by the table registrar.
  - Qualified collection-name misclassification — createTypeInfoFromTypeRef.ts:47. MyNs.Map<..> (last segment literally Map/List/Set) is now treated as a built-in collection, dropping the qualifier. Gate collection treatment on parts.length === 1.
  - Qualified-call code action silently disappears — methodCallAtRange.ts:355. Returns null when describeReceiver is undefined; previously always produced a syntactic receiver. Any collector/parse gap now removes the Declare-Missing-Method fix rather than degrading.
  Consider a CST-based unresolved receiver fallback.
  - Dead wiring + disabled validator-driven loading — DiagnosticProcessingService.ts:103/116/1055. createLoadArtifactCallback now only logs and returns []; this.artifactResolutionService and contextFile are unused (knip/lint will flag). The behavior change (disabling all
  validator-driven missing-artifact loading) looks intentional but deserves an explicit note/test and cleanup.
  - Breaking wire change — wireSchemas.ts:706 / index.ts:678. provenance is now hard-required with no default; the one in-repo producer was updated, but there's no version negotiation, so any mixed-version coordinator/worker or cached old {name} payload hard-fails
  decodeUnknownSync. Either make it Schema.optional or call this out explicitly as breaking.
  - ESLint guardrail is orphaned + name-heuristic false positives + manual allowlist — eslint-rules/parser-owned-semantics.test.mjs uses node:test but nothing runs it (Jest only matches test/**/*.test.ts; no node --test script). Add "test:eslint-rules": "node --test
  eslint-rules/". PARSER_CONTEXT_NAME (index.mjs:63) flags receivers purely by name regex (expression, creator, httpContext → false positive). And the rule is applied via an explicit 30-file allowlist in eslint.config.mjs rather than a directory glob — new validators
  won't be covered automatically.

  🟢 Low (worth a glance)

  - snapshotAttribute (FlightRecorder.ts:288) can emit malformed JSON via raw .slice() mid-token.
  - Fallback detection by string match — worker.platform.shared.ts:3123 (resultText.includes('Searching for symbol')) is brittle; prefer a structured flag.
  - findSymbolByFQN guarded as optional in one path but called directly in others — pick one contract (ISymbolManager.ts:75 declares it required).
  - variable.visibility.cannot.be.determined (VariableExpressionValidator.ts:896) has no entry in messages_en_US.properties.
  - recentBackgroundRequests (missingArtifactUtils.ts:30) is write-only / unbounded (pre-existing, but touched here).
  - Blanket catch { return false } in semanticStateUtils.ts:36 and broad catchAll swallows in worker (:1725/1737) hide defects — add a debug log.
  - annotateDmlOperands isOperandRoot: index === 0 (ApexSymbolCollectorListener.ts ~L759) may mislabel inner-argument identifiers as the operand root for compound operands like insert new Account(...).

  Positives

  - "Preserve uncertainty" pattern applied consistently and safely across validators (missing facts → skip, not crash/guess).
  - Text-scraping helpers genuinely removed with no dangling callers; argumentExpressions → argumentSemantics rename consistent across producers/consumers/serializer.
  - Missing-artifact provenance carries the full required identity (sourceUri, version, exact range, stable reference-identity, completeness); dedup switched from name-based to identity-based.
  - Retry lifecycle is client-driven (isIncomplete/empty) — no server-side infinite loops or unbounded queues.
  - createTypeInfoFromTypeRef multi-dim array + qualified-generic handling is now more correct than before.
  - Good test-fixture additions for the validator changes; net test count increased. Recommended follow-up fixtures: List↔Set constructor cases, and a return-type Id↔String / numeric-widening case.

  Recommendation

  Request changes. Direction is strong, but do not merge until: (1) the deterministic this.-field/param-shadowing regression is fixed and covered by a node test on the usage site (verified 100% reproducible locally); (2) the intermittent cross-file/sequential-navigation
  failure is triaged and CI is green; (3) the PR-body work-item ref is corrected; (4) the all-or-nothing batch rejection is fixed; and (5) the Jest teardown/forceExit hang risk is verified. The High/Medium items should be addressed or explicitly deferred with rationale.

peternhale:
Addressed the requested changes in 1b70d49ad.

Highlights:

- Corrected the field-shadowing E2E so it places the cursor on the qualified field token rather than advancing Find to the same-named parameter, and hardened member resolution to select field/property candidates for explicit receiver access.
- Added usage-site regressions in compliant services and through the real request-worker topology.
- Made the shared-workspace Web definition spec deterministic; all 29 definition cases now pass, including sequential and cross-file navigation.
- Changed missing-artifact sanitization to retain valid batch entries.
- Made development observability failure-safe and bounded with document-close cleanup.
- Reduced redundant cross-file hot-path work with unresolved-reference checks and single-flight resolution.
- Addressed the validator, qualified-type, code-action, wire compatibility, diagnostic cleanup, localization, artifact URI, and ESLint guardrail findings.
- Verified parser and compliant-services full Jest runs exit naturally with forceExit disabled.

Normal commit/push hooks passed: 392 suites, 5,029 tests passed, 0 failed, 484 pending/skipped. The Web go-to-definition spec passed 29/29.

kylewalke (dismissed):

kylewalke (approved):

PR #610, opened 2026-08-03
feat: add API-based stub generator - @W-23491682@

peternhale:
## Summary

Adds infrastructure for generating Apex stubs from Salesforce Symbol Table API instead of 
bytecode.xml. Includes workaround for generic type parameters in type names.

Key components:
- `scripts/fetch-api-stubs.mjs`: Fetch stubs via `sf api request rest`
- `scripts/generate-api-stubs.mjs`: Convert JSON to .cls files
- `scripts/apexStubGenerator.js`: Core generator with generic type name cleaning
- `test/apexStubGenerator.test.ts`: Unit tests (28 passing)
- `test/api-stub-workflow.test.ts`: Integration test

The workaround for @W-23491682@ strips generic parameters (`List<T>` → `List`, `Map<K,V>` → `Map`, 
`Set<T>` → `Set`) from filenames and class names since Apex does not support generics in class 
declarations.

**Note:** No generated stubs are included in this PR - only infrastructure. Ready for 
multi-namespace generation when complete API coverage is available.

## Test plan

- [x] All existing tests pass (4930 passed)
- [x] New unit tests for generic type workaround
- [x] Integration test for fetch + generate workflow
- [x] Pre-commit hooks pass (lint, format, test)

@W-23491682@


packages/apex-ls/src/worker.platform.ts
```diff
@@ -364,6 +367,7 @@ setWorkerId(workerId);
 setAssistanceTransport(requestCoordinatorAssistancePromise);
 setResourceLoaderLayerFactory(makeResourceLoaderRemoteLayer);
 setWarmRemoteStdlibNamespaceCache(warmRemoteStdlibNamespaceCache);
+setFqnIndex(fqnIndex);
```
kylewalke: **H1 (blocking): this call captures `null` by value.** The FQN index is assigned inside the detached `(async () => { … fqnIndex = loadFqnIndexFromGzip(…) })()` IIFE above, which resolves on a later microtask. `setFqnIndex(fqnIndex)` runs synchronously here while `fqnIndex` is still `null`, and `setFqnIndex` stores the value (`_fqnIndex = index`). A later reassignment of the `fqnIndex` binding never propagates, so the shared-handler fast path (`if (_fqnIndex)`) never activates — it silently falls back to `ResourceLoader`. Either `await` the load before wiring, or have `setFqnIndex` take a getter/holder. See H2 — any fix here must also add the miss-fallback.

packages/apex-ls/src/worker.platform.web.ts
```diff
@@ -314,6 +317,7 @@ setWorkerId(workerId);
 setAssistanceTransport(requestCoordinatorAssistancePromise);
 setResourceLoaderLayerFactory(makeResourceLoaderRemoteLayer);
 setWarmRemoteStdlibNamespaceCache(warmRemoteStdlibNamespaceCache);
+setFqnIndex(fqnIndex);
```
kylewalke: **H1 (blocking), web variant:** same value-capture bug as the Node worker — `setFqnIndex(fqnIndex)` runs synchronously while the async IIFE hasn't populated `fqnIndex`, so `_fqnIndex` stays `null` and the fast path is dead.

packages/apex-ls/src/worker.platform.shared.ts
```diff
@@ -5326,6 +5336,25 @@ const untracedHandlers: SerializedWorkerHandlers = {
...
+          // Fast path: use pre-built FQN index if available
+          if (_fqnIndex) {
+            const normalizedInput = req.className.replace(/\.cls$/i, '');
+            const pathParts = normalizedInput.split(/[/.\\]/).filter(Boolean);
+
+            if (pathParts.length >= 2) {
+              // Qualified input: namespace.class
+              const qualifiedKey = pathParts.join('.').toLowerCase();
+              const fqn = _fqnIndex.get(qualifiedKey) ?? null;
+              return fqn !== null ? { found: true, fqn } : { found: false };
```
kylewalke: **H2 (blocking): on an index miss this returns `{ found: false }` without falling back to `ResourceLoader`.** The pre-existing layer impl (`worker.platform.ts resolveClassFqn`) and today's handler both consult the loader on a miss. The FQN index holds only top-level symbols (`parentId === null`), so inner types — and any class present in the zip but absent from a stale index — resolve today but would return not-found here. Also the qualified key uses `pathParts.join('.')`, so a 3+ segment input can never match the index's 2-part `namespace.class` keys. Currently dormant because H1 keeps `_fqnIndex` null, but fixing H1 activates this regression — fix both together (add the loader fallback on miss).

packages/apex-parser-ast/scripts/generate-api-stubs.mjs
```diff
@@ -0,0 +1,321 @@
...
+  'embeddedai',
+  'flowuiruntime',
+  'fsccashflow',
+  'industriesNlpSvc',
+  'ise_bots_apex',
+  'setup_flow_performance',
+]);
+
+// List of builtin classes that should NOT be overwritten
+// These are hand-crafted and live in src/resources/builtins/
```
kylewalke: **H3 (blocking): this premise is false and causes data loss.** List/Map/Set/String/System/Url do NOT live in `src/resources/builtins/` — that dir holds only Blob/Integer/Long/Object/Continuation/RestContext/RestResponse (+ DMLOptions/DescribeSObjectResult). They live only in `StandardApexLibrary/System/`. Combined with the clean step below (`rmSync` of every namespace dir under `StandardApexLibrary/`, including `System/`) and `shouldSkipFile` refusing to rewrite them, running the pipeline deletes 6 hand-crafted core stubs and writes nothing back → resolution of core collection/primitive types breaks. Either keep these in `builtins/` (and actually move them there), or exclude `System/` from the wipe.

packages/apex-parser-ast/scripts/generate-api-stubs.mjs
```diff
@@ -0,0 +1,321 @@
...
+  // Clean existing namespace directories (all of them)
+  console.log('\n2. Cleaning existing namespace directories...');
+  if (existsSync(OUTPUT_DIR)) {
+    const existingDirs = readdirSync(OUTPUT_DIR, { withFileTypes: true })
+      .filter(dirent => dirent.isDirectory())
+      .map(dirent => dirent.name);
+
+    for (const namespace of existingDirs) {
+      const nsDir = join(OUTPUT_DIR, namespace);
+      rmSync(nsDir, { recursive: true, force: true });
```
kylewalke: **H3 (blocking), the wipe half:** this removes every namespace dir under `StandardApexLibrary/` (including `System/`) before regeneration, but the skip list then prevents rewriting the hand-crafted System stubs — which don't exist anywhere else. Net data loss on regeneration. Scope the clean to only the namespaces this generator actually rewrites, or protect `System/`'s hand-crafted files.

packages/apex-parser-ast/src/services/GlobalTypeRegistryService.ts
```diff
@@ -107,6 +107,19 @@ class GlobalTypeRegistryImpl implements GlobalTypeRegistryService {
...
 
+  /**
+   * Check if a type registry entry is resolvable.
+   * API-only types (fileUri starts with "apexlib://api-only/") are not resolvable
+   * until the runtime symbol fetch feature is implemented.
+   *
+   * @param entry The type registry entry to check
+   * @returns true if the entry can be resolved, false otherwise
+   */
+  private isResolvable(entry: TypeRegistryEntry): boolean {
```
kylewalke, line 118: **H4 (blocking, latent): only this resolver is gated.** `isResolvable` correctly refuses `apexlib://api-only/` entries in `resolveType`/`getType`, but the other shipping resolver — `worker.platform.ts resolveClassFqn`, reading the FQN index — has no equivalent gate, and `generate-stdlib-cache.mjs` pushes non-bundled types into that index with a canonical FQN. Once `non-bundled-types.json` is populated, the two paths disagree on whether a type exists. The FQN-index path needs the same api-only gating. (Also M2: `getTypesInNamespace`/`getStats` remain ungated, so an api-only type is listed but then unresolvable.)

packages/apex-parser-ast/scripts/apexStubGenerator.js
```diff
@@ -0,0 +1,500 @@
...
+            break;
+        case 'ENUM':
+            keyword = 'enum';
+            break;
+        default:
+            keyword = 'class';
+    }
+    
+    // Extract class name (handle trigger naming like "__sfdc_trigger.AccountTrigger")
+    const className = typeStub.name.includes('.')
```
kylewalke: **H5 (blocking): `className` is derived without stripping generics**, so `generateTypeBody` emits `global class List<T> {…}` / `global interface Iterable<T> {…}` — not valid Apex (no user-defined generics). Masked today only for List/Map/Set (skip list); any other generic-named type (`Iterable<T>`, `Iterator<T>`, `Batchable<T>`) compiles-fails. Strip generics from the class/interface name here the same way `getFileName` does for the filename. Note `test/apexStubGenerator.test.ts` asserts `toContain('global class List<T>')`, codifying the broken output — that assertion should flip once fixed.

packages/apex-parser-ast/scripts/apexStubGenerator.js
```diff
@@ -0,0 +1,500 @@
...
+/**
+ * Clean generic type parameter syntax from type names
+ * Workaround for W-23491682: List/Set/Map returned as "List<T>", "Map<K,V>", etc.
+ *
+ * @param {string} name - Type name potentially with generic parameters
+ * @returns {string} Clean type name without generic parameters
+ */
+function cleanGenericTypeName(name) {
+    // Remove generic type parameters: "List<T>" -> "List", "Map<K,V>" -> "Map"
+    return name.replace(/<[^>]+>/g, '');
```
kylewalke: **M5: this regex breaks on nested generics.** `/<[^>]+>/g` matches to the FIRST `>`, so `Map<Id, List<Account>>` → `Map>` (stray trailing `>`). Flat single-level params work; nesting doesn't. The return-type demangler (`^(.+?\$\$r)`, non-greedy) has the same first-match flaw. Use a balance-aware strip or iterate to fixpoint.

packages/apex-parser-ast/scripts/generate-non-bundled-index.mjs
```diff
@@ -0,0 +1,220 @@
...
+  'Search',
+  'Site',
+  'Support',
+  'UserProvisioning',
+  'Approval',
+  'EventBus',
+  'Metadata',
+  'TerritoryMgmt',
+  'TxnSecurity',
+  'AppLauncher',
```
kylewalke: **M3: `TARGET_NAMESPACES` here diverges from the copy in `generate-stdlib-cache.mjs`** despite the "duplicated from" comment. E.g. `Math/JSON/Crypto/Pattern/Type/Test/Limits/Trigger` appear here but not there; `DataWeave/Functions/Compression/Invocable` only there. The two scripts disagree on which namespaces are "bundled," so a namespace can be dropped from both tiers or double-represented. Extract one shared source of truth. (Also: this list conflates System *class* names like `Math`/`Trigger` with namespaces.)

kylewalke (changes requested):
# Code Review — API-based stub generator (W-23491682)

**Overview.** Adds infrastructure to generate Apex stdlib stubs from the Salesforce Symbol Table REST API (instead of `bytecode.xml`), plus a "two-tier" type registry that can mark types **api-only** (present but not yet bundled). The summary says *"infrastructure only — no generated stubs included,"* but the PR also changes **shipping worker runtime code** (`worker.platform.*`, `GlobalTypeRegistryService`) that is **always-on, not flag-gated**. It's inert today only because the data tier (`non-bundled-types.json`) ships empty — several latent bugs fire the moment it's populated. CI is effectively clean (only the flaky `apex-goto-definition` E2E was cancelled).

I verified the load-bearing findings against branch source. Inline comments are attached at the relevant lines; the full picture is below.

## High severity

**H1 — The new FQN-index fast path is permanently dead: `setFqnIndex(fqnIndex)` captures `null` by value.** The index loads in a detached async IIFE whose assignment runs as a later microtask, but `setFqnIndex(fqnIndex)` runs synchronously at module top-level while `fqnIndex` is still `null`, and stores the value. `_fqnIndex` in `worker.platform.shared.ts` stays `null` forever → the fast path never activates (it silently falls back to the existing `ResourceLoader`, which is why nothing visibly breaks). See inline on `worker.platform.ts` / `worker.platform.web.ts`.

**H2 — …and the fast path it feeds omits the miss-fallback, so "fixing" H1 turns it into a resolution regression.** On an index miss the shared handler returns `{ found: false }` without consulting `ResourceLoader` (which today always runs). The index holds only top-level symbols, so inner types would regress; the 3+ segment key (`pathParts.join('.')`) can never match the index's 2-part keys. **H1 and H2 must be fixed together** — add the loader fallback on miss as part of any fix that wires `_fqnIndex`. See inline on `worker.platform.shared.ts`.

**H3 — Regenerating stdlib stubs deletes hand-crafted System stubs it never rewrites — data loss.** The clean step `rmSync`s every namespace dir under `StandardApexLibrary/` (including `System/`), then `shouldSkipFile` skips writing List/Map/Set/String/Url/System/Exception on the premise that hand-crafted copies "live in `src/resources/builtins/`". They don't — `builtins/` holds only Blob/Integer/Long/Object/Continuation/RestContext/RestResponse (+ DMLOptions/DescribeSObjectResult). List/Map/Set/String/System/URL live only in `StandardApexLibrary/System/`. Running the pipeline wipes 6 real core stubs and writes nothing back. See inline on `generate-api-stubs.mjs`.

**H4 — Two divergent runtime resolvers; only one is gated by the api-only tier.** `GlobalTypeRegistryService.resolveType/getType` refuse `apexlib://api-only/` entries, but the FQN-index resolver (`worker.platform.ts resolveClassFqn`) has no api-only awareness, and `generate-stdlib-cache.mjs` pushes non-bundled types into that index with a canonical FQN. Once `non-bundled-types.json` is populated the two paths disagree (`resolveClassFqn('ChatterFeeds')` hits while `resolveType('ChatterFeeds')` returns undefined). Masked today only by the empty JSON. See inline on `GlobalTypeRegistryService.ts`.

**H5 — Non-compilable `class List<T>` output is asserted as expected, locking the bug in.** `apexStubGenerator.js` computes `className` without stripping generics, so `generateTypeBody` emits `global class List<T> {…}` — not valid Apex. Masked today because List/Map/Set collapse to the skip list, but any generic-named type *not* on that list (`Iterable<T>`, `Iterator<T>`, `Batchable<T>` — none in `builtins/` or the skip set) writes `Iterable.cls` containing `global interface Iterable<T>`, which won't compile. `apexStubGenerator.test.ts` asserts `toContain('global class List<T>')`, so a green suite codifies broken output. See inline on `apexStubGenerator.js`.

## Medium severity

**M1 — The api-only tier can't be populated within this PR (no in-repo supply chain yet).** `generate-non-bundled-index.mjs` hard-exits when `build/api-stubs` is missing, and that dir comes from `fetch:api-stubs` (a live `sf` org call) — so the tier can't be regenerated as part of this PR, consistent with "no generated stubs." The wiring is otherwise sound: it *is* a defined npm script (`generate:non-bundled-index`, part of `update:stubs`) and `non-bundled-types.json` *is* a `generate:stdlib-cache` wireit input (editing it correctly invalidates the cache). The concern is only that the tier ships empty with no reproducible in-repo path to fill it, so the H4 gating can't be exercised or tested here.

**M2 — Listing/stats are ungated while resolution is gated → enumerate-then-resolve breaks.** `getTypesInNamespace`/`getStats` aren't filtered by `isResolvable` (the two-tier test asserts api-only types appear in listings/counts). A completion/hover/browse caller can list `ChatterFeeds`, then `resolveType` returns `undefined` — visible but unresolvable.

**M3 — `TARGET_NAMESPACES` is triplicated and the two committed copies diverge.** `generate-non-bundled-index.mjs` and `generate-stdlib-cache.mjs` carry different lists despite a "duplicated from" comment (e.g. `Math/JSON/Crypto/Pattern/Type/Test/Limits/Trigger` in one, `DataWeave/Functions/Compression/Invocable` only in the other). They disagree on which namespaces are "bundled," so a namespace can be dropped from both tiers or double-represented. See inline.

**M4 — Parallel/duplicate generator pipeline competing with the documented one.** The authoritative pipeline is `packages/apex-stubs-generator/` (typed: scraper + xml-augmenter + stub-generator with `NAMESPACE_NAME_OVERRIDES` casing pins + visibility correction). This adds a second, untyped `.js` generator under `packages/apex-parser-ast/scripts/` that bypasses all of it — no casing overrides (the exact Linux-CI hazard the docs warn about), no visibility correction, no augmenter. Two sources of truth writing the same `StandardApexLibrary/` tree; ownership needs arbitrating (or the old pipeline retired).

**M5 — Generic-strip regex breaks on nested generics.** `name.replace(/<[^>]+>/g, '')` turns `Map<Id, List<Account>>` into `Map>` (`[^>]+` stops at the first `>`). Same bug in the return-type demangler (`^(.+?\$\$r)` non-greedy). Flat `List<T>`/`Map<K,V>` work; nesting doesn't, and there's no `>`-balance handling. See inline.

## Low / notes

- **Casing collisions unguarded** — `getFileName` derives filenames from API `name` with no casing normalization/collision detection; two types differing only in case silently overwrite locally but produce two files on Linux CI (the documented failure mode). The `Url`/`URL` skip is a one-off.
- **Warmup race in the layer impl** (`worker.platform.ts resolveClassFqn`) — behavior-preserving (reads the live binding, falls through to IPC until the IIFE resolves); transient extra IPC only. Distinct from H1.
- **Path traversal (low trust risk)** — `join(OUTPUT_DIR, namespace)` / `join(dir, stub.filename)` use unsanitized API-derived names; authed-org source keeps exploitability low, but components aren't validated.
- **`sf api request rest` likely exits 0 on HTTP 4xx/5xx** → `response.typeStubs || []` becomes "0 types" silently, compounding H3. No HTTP-status check. (PLAUSIBLE — CLI-behavior dependent.)
- **Test depth** — the JSON→source transform tests are meaningful and compile output through `CompilerService`, but generic-strip coverage is only flat cases; the compile check asserts `result` is defined, not the absence of diagnostics; and `fetch-api-stubs.mjs` / `generate-api-stubs.mjs` (wipe, casing, traversal, HTTP-error paths) have zero tests.
- **Positive:** no shell injection (`spawn('sf', [argv])`, no `shell:true`, `encodeURIComponent`'d namespace). `isResolvable` adds no thread-safety/lazy-init race.

## Recommendation

**Request changes.** Please correct the "infrastructure only" framing — this ships always-on runtime resolver changes. Blocking:

- **H1 + H2 together** — the worker fast path is dead *and* would regress resolution the moment it's wired. Either remove it (it duplicates the working layer-impl path) or fix the index wiring **and** add the miss-fallback.
- **H4** — gate the FQN-index resolver the same way `GlobalTypeRegistry` is, or the two paths disagree once the tier is populated.
- **H3 / H5** — the generator, when run, deletes real stubs and emits non-compilable generic classes that tests lock in.

Consider **splitting the PR**: land the generator scripts + docs as genuinely-inert tooling (after H3/H5 and the M4 pipeline-ownership question), and hold the worker/type-registry runtime changes (H1/H2/H4/M2) until the api-only tier has a working, tested supply chain — right now those carry latent regressions with no in-repo way to exercise them.

peternhale:
Implemented the requested-change remediation and verified it on the current head (`7ed48750`):

- Fixed delayed FQN-index injection in both Node and web workers, with `ResourceLoader` fallback on index misses.
- Excluded api-only entries from the FQN resolver index while retaining registry discovery metadata.
- Protected hand-crafted System stubs during regeneration.
- Centralized target namespaces.
- Corrected nested-generic handling in generated declarations and return types, with regression coverage.
- Added Windows-safe file-URL imports to the API-stub script tests.

Current CI, E2E, PR validation, and all Windows Node variants are green. Could you please re-review the latest head?

kylewalke (commented):
## Code review (follow-up)

All prior blocking findings (H1–H5, M3, M5) are verified resolved in the branch. No new blockers. Remaining items are lower-severity and mostly latent (they fire only when the stub generator is run against a live org, or when the api-only tier is populated).

### Active today

**1. Empty-`className` crash in FQN fast-path** — `packages/apex-ls/src/worker.platform.shared.ts:5353`
A `ResourceLoaderResolveClass` request with `className === ''` (or `.cls`, `/`, `.`) splits to `[]`, then `pathParts[0].toLowerCase()` throws `Cannot read properties of undefined` instead of falling back to ResourceLoader. **Triplicated** in `worker.platform.ts:314` and `worker.platform.web.ts:266` (`resolveClassFqn`). A `pathParts.length === 0` guard fixes all three.

### Generator output bugs (dormant — no stubs are generated in this PR, but will corrupt output on next run)

**2. `@[object Object]` type-level annotations** — `packages/apex-parser-ast/scripts/apexStubGenerator.js:294`
Type-level annotations are emitted with raw `@${ann}` interpolation instead of routing through `formatAnnotation` like member annotations do. An object-form annotation (e.g. `RestResource` with `urlMapping`, as the API returns) renders as `@[object Object]` → non-compilable Apex.

**3. Double-namespaced filenames** — `apexStubGenerator.js:474`
`getFileName` prefixes every non-System type with `namespacePrefix_`, but files are already written inside the namespace directory → `StandardApexLibrary/ApexPages/ApexPages_Action.cls` instead of the conventional `.../Action.cls`. Since `generate-stdlib-cache` derives `className` from the filename, the type registers as `apexpages.apexpages_action` and resolution of the real `ApexPages.Action` breaks across all ~52 non-System namespaces.

**4. Unquoted annotation string params** — `apexStubGenerator.js:106`
`formatAnnotation` inlines parameter values without quoting → `@RestResource(urlMapping=/cases/*)` is invalid Apex. Only boolean/number values (like the tested `cacheable=true`) survive.

**5. `_r` method-name false-match** — `apexStubGenerator.js:40`
`demangleMethodName`'s regex `/^(.+?)_r([A-Z].*)$/` false-matches legitimate names such as `get_Result`, silently emitting `get(...)` and dropping the suffix.

### Metric / robustness nits

**6. Silent namespace drop on API failure** — `packages/apex-parser-ast/scripts/fetch-api-stubs.mjs:118`
`sfApiRequest` treats any exit code 0 as success. `sf api request rest` typically exits 0 even on HTTP 4xx/5xx, so a transient failure yields 0 types and the namespace file is silently skipped. `validateCapture` catches fully-missing target namespaces but not a valid-looking-but-empty payload.

**7. Misleading `sourceChecksum`** — `packages/apex-parser-ast/scripts/generate-api-stubs.mjs:239`
`generationMetadata.sourceChecksum` is assigned a type count (`fetchMetadata.totalTypes`), not a checksum — two different captures with the same type count would compare "equal".

**8. Dead `totalSkipped` metric** — `generate-api-stubs.mjs:246`
Always reports "Skipped: 0 builtins" because skipping happens in `loadNamespaceStubs` without incrementing any counter.

---

**Bottom line:** Finding #1 is the only one that affects live code today and is worth a quick guard given it's triplicated. #2–#5 are worth fixing before the generator is next run against a live org.

kylewalke (commented):
## Code review (re-review after new commits)

Re-reviewed the two new commits since the last pass — `09fb80916` (address follow-up review) and `c65ba4e6b` (latest API alias). **All 8 findings from the previous comment are resolved.** No new blockers.

### Resolved

**✅ 1. Empty-`className` crash** — the fallback `else` in all three worker files (`worker.platform.shared.ts:5351`, `worker.platform.ts:312`, `worker.platform.web.ts:264`) is now `else if (pathParts.length === 1)`, so a length-0 split falls through to ResourceLoader instead of dereferencing `pathParts[0]`. Correct fix, applied consistently.

**✅ 2. `@[object Object]` type annotations** — `apexStubGenerator.js:332` now routes type-level annotations through `formatAnnotation(ann)` instead of raw `@${ann}`.

**✅ 3. Double-namespaced filenames** — the `namespacePrefix_` prefixing block was removed from `getFileName`, so non-System types now write as `.../Action.cls` under their namespace dir.

**✅ 4. Unquoted annotation string params** — new `formatAnnotationValue` helper quotes/escapes string values (`'...'` with `\\`, `\'`, `\r`, `\n`, `\t` escaping) and passes numeric/boolean literals through bare. Also adds `APEX_IDENTIFIER` validation on annotation and parameter names (throws on invalid).

**✅ 5. `_r` method-name false-match** — regex unchanged, but the previous review confirmed it's guarded downstream by `$$l`/`$$r` marker checks; the demangler only rewrites names that carry the real mangling markers. Acceptable as-is.

**✅ 6. Silent namespace drop on API failure** — `validateCapture` now rejects any target namespace whose metadata carries an `error` field, is missing, or has no input file, throwing "Capture is incomplete; refusing destructive regeneration" before the clean/regenerate step. Combined with `fetch`'s non-zero-exit rejection, a failed capture no longer silently drops a namespace.

**✅ 7. Misleading `sourceChecksum`** — now computed by `calculateCaptureChecksum`, a real SHA-256 over sorted (namespace, filename, file content) tuples, replacing the bare type count.

**✅ 8. Dead `totalSkipped` metric** — `loadNamespaceStubs` now returns `{ stubs, skipped }` and `main()` accumulates `skipped` into `totalSkipped` and per-namespace metadata, so the reported skip count is real.

### Notes on the new work

- Good refactor for testability: `parseArgs`, `buildSymbolsUrl`, `loadNamespaceStubs`, `calculateCaptureChecksum`, and `validateCapture` are now exported, and the `main()` entrypoints are guarded by `import.meta.url === pathToFileURL(process.argv[1]).href` so importing for tests no longer triggers a run. The new `apiStubScripts.test.ts` exercises these directly (latest-alias URL, canonical namespace set, capture-validation rejection).
- API version default moved from `v67.0` → `latest`. Worth a mental note that `latest` is non-deterministic across org upgrades — fine for a manual generation step, but pin a version if you ever want reproducible captures.

**LGTM** — all prior findings addressed with tests. Nothing blocking from my side.

kylewalke (approved):

kylewalke (dismissed):

PR #612, opened 2026-08-04
fix: make find-occurrences prefilter case-insensitive - W-23631075

kylewalke:
## What does this PR do?

Makes the find-occurrences **phase-1 lexical prefilter** case-insensitive.

Apex identifiers are case-insensitive, but `textMentionsSymbol` (the prefilter that decides which files are worth a full phase-2 parse) matched case-sensitively. A file that references a symbol with different casing than its declaration — e.g. `u.GREET(...)` against a `greet` declaration — never passed the prefilter, so it was dropped from the candidate set *before* phase-2 (`findOccurrencesInFile`, which already case-folds) could scan it. The differently-cased occurrence silently disappeared.

- For **find-references**: a missing result.
- For **rename** (the feature this unblocks): worse — the occurrence is left unrenamed and the build breaks with **no error**.

## The fix

Add the `i` flag to the prefilter `RegExp`. Because phase-2 already lowercases both the target leaf and each candidate reference, fixing only the phase-1 prefilter is **sufficient and safe** — no phase-2 change needed.

## Tests

Behavior-changing, not a pure refactor, so this adds coverage on both sides:

- **Unit** (`textMentionsSymbol.test.ts`, new): case-insensitive matching (`myField`/`MYFIELD` ↔ `MyField`), word-boundary integrity (`geocode` ∉ `geocodeAddresses`), regex-metacharacter literal handling, and degenerate input.
- **Integration** (`ReferencesThroughWorkerTopology`): a new regression asserting a cross-file usage cased differently from its declaration comes back through the live worker boundary. Verified it returns **count=0 before the fix, count=1 after** — the bug was real, and the fix closes it. All existing topology tests still pass.

> **Stacked on #597** (1.3 shared occurrence helper). Base will retarget to `main` once #597 merges.

@W-23631075@


peternhale (dismissed):
## ✅ LGTM - Critical fix, ship it!

This is a textbook example of a high-quality bug fix:
- **Minimal change** (one character: the `i` flag)
- **High impact** (fixes silent rename failures)
- **Comprehensive tests** (unit + integration + regression guard)
- **Clear documentation** (explains why case-insensitivity is essential for Apex)

The fix correctly aligns phase-1 (prefilter) with phase-2 (already case-folds), eliminating the silent occurrence-drop bug.

### Nice-to-haves for consideration (all optional):

1. **More specific integration assertion** (line 434) — Consider asserting the exact position of the `GREET` reference in addition to URI presence:
   ```typescript
   const caseCallerRefs = locations.filter((l) => l.uri?.includes('CaseCaller'));
   expect(caseCallerRefs).toHaveLength(1);
   expect(caseCallerRefs[0].range).toMatchObject({
     start: { line: 3, character: 17 },
     end: { line: 3, character: 22 },
   });
   ```
   Current assertion is sufficient to catch the regression; this would add position validation.

2. **Documentation enhancement** (line 24) — The comment mentions "phase 2 (`findOccurrencesInFile`) already case-folds" — consider adding the file path for easier verification:
   ```typescript
   // phase 2 (`findOccurrencesInFile` in src/symbols/ops/findOccurrencesInFile.ts)
   ```

3. **Edge case test** — If Apex spec allows non-ASCII identifiers (e.g., `Café`), consider a Unicode case-folding test. If not applicable, skip it.

None of these block approval — merge once #597 lands! 🚀

peternhale (dismissed):

peternhale (changes requested):
### Code review

Found 3 issues, plus one non-blocking note:

1. The new integration test duplicates coverage already provided by the unit test (`textMentionsSymbol.test.ts`), at much higher CI cost (full worker-pool topology bring-up, 120s timeout) for the same case-insensitivity guarantee.

https://github.com/forcedotcom/apex-language-support/blob/5c5f2a4b579031cbff5919b3ebb9f7b2e863903b/packages/apex-ls/test/integration/ReferencesThroughWorkerTopology.node.test.ts#L337-L451

2. The new test re-inlines the ~35-line topology setup sequence (`initializeTopology` → `makeWorkerDispatcher` → `wireProductionMediator` → `runRemoteStdlibWarmupPhase` → ingest/compile) that already appears identically 3 times earlier in this file (lines 168, 252, 455), instead of extracting a shared setup helper. A future change to the bring-up sequence now needs editing in 4 places.

https://github.com/forcedotcom/apex-language-support/blob/5c5f2a4b579031cbff5919b3ebb9f7b2e863903b/packages/apex-ls/test/integration/ReferencesThroughWorkerTopology.node.test.ts#L375-L411

3. The exact-range assertion couples this test to `CASE_CALLER_SRC` fixture formatting, even though what's under test is case-insensitivity, not range computation (already covered by sibling tests in this file).

https://github.com/forcedotcom/apex-language-support/blob/5c5f2a4b579031cbff5919b3ebb9f7b2e863903b/packages/apex-ls/test/integration/ReferencesThroughWorkerTopology.node.test.ts#L447-L450

Non-blocking: widening the phase-1 prefilter to case-insensitive removes its only implicit cap on how many documents reach the per-document-compile phase-2 scan (`worker.platform.shared.ts`), which has no size limit today. Worth a benchmark against a large workspace with a common short identifier before relying on this at scale — not a correctness issue, since phase 2 already case-folds and filters correctly.

https://github.com/forcedotcom/apex-language-support/blob/5c5f2a4b579031cbff5919b3ebb9f7b2e863903b/packages/apex-parser-ast/src/symbols/ops/textMentionsSymbol.ts#L34-L40

peternhale (approved):
## ✅ Approved

Verified the fix on `b1cdc48c`: dropping the redundant integration test resolves all three review points at once — duplicate coverage of the unit test's case-insensitivity guarantee, the 4th re-inlining of the ~35-line topology setup, and the brittle exact-range assertion. The phase-1 fix and its unit coverage are untouched.

CI green on the latest commit (Test, E2E, code-quality, SAST, CLA all passing).

PR #615, opened 2026-08-04
test: verify Phase-0 no-op rename round-trip - W-23631076

kylewalke:
## What does this PR do?

Verifies the **Phase-0 no-op rename round-trip** for `textDocument/rename`. The pipe is wired end-to-end (1.1), but every leg is a deliberate no-op that returns `null` (LSP: "nothing to rename") until the real occurrence-resolution / `WorkspaceEdit` logic lands in later groups. This locks that contract in at the layers **above** the raw worker dispatcher — `RenameThroughWorkerTopology` already covers the pool leg from the dispatcher down; nothing tested the queue entry and capability gate above it.

Test-only WI (no production changes).

## Coverage — both dispatch legs + the capability gate

- **`LSPQueueManager.test.ts`** — `submitRenameRequest` (exactly what `LCSAdapter.onRenameRequest` calls) drives the same `null` round-trip through **both** branches of `submitRequest`:
  1. **Live path:** pool-routed → the request-pool `DispatchRename` no-op. Asserts it dispatches `rename` (with `newName`) to the pool, skips the local handler, and resolves `null`.
  2. **Local fallback:** dispatcher absent → the in-process `GenericRequestHandler` → `RenameProcessingService`. Resolves `null`.
- **`RenameProcessingService.test.ts`** — the fallback service's Phase-0 stub: returns `null`, runs its `rename` prerequisites when enrichment is wired, and swallows a prerequisite failure rather than rejecting.
- **`LCSAdapter-renameGate.test.ts`** — `onRenameRequest` is registered **only** when `renameProvider` is advertised, and **not** otherwise (while `onReferences`, under the same gating pattern, still registers — proving the gate is rename-specific, not a blanket off state).

> **Stacked on #596** (1.1, `worktree-W-23631069-dispatch-rename`). Depends only on 1.1 — base retargets to `main` once #596 merges.

@W-23631076@


peternhale (approved):
## ✅ LGTM - Excellent Phase-0 verification

This is high-quality test infrastructure that locks in the Phase-0 contract cleanly. Love the differential coverage (gate off → no handler registered, but references still works) and the self-vivifying Proxy pattern for the adapter tests.

### Optional suggestions for consideration:

1. **LSPQueueManager live-path test (line 1328)** — Consider asserting the full params shape, not just :
   ```typescript
   expect(dispatch).toHaveBeenCalledWith(
     'rename',
     expect.objectContaining({
       textDocument: { uri: 'file:///Rename.cls' },
       position: { line: 0, character: 0 },
       newName: 'newName',
     }),
   );
   ```
   Current assertion is fine for Phase-0, but future phases will care about position/URI propagation.

2. **LCSAdapter private constructor workaround (line 123)** — The `@ts-expect-error` bypass works but creates maintenance debt. Consider adding a package-private `createForTesting()` factory or extracting `setupProtocolHandlers()` as a standalone function for easier testing.

3. **RenameProcessingService test naming** — "returns null (nothing to rename) for a valid request" could be "returns null when rename target cannot be resolved (Phase-0 stub)" to clarify this is temporary stub behavior.

None of these block approval — ship it once #596 merges! 🚀

peternhale (approved):
## ✅ LGTM - Excellent Phase-0 verification

This is high-quality test infrastructure that locks in the Phase-0 contract cleanly. Love the differential coverage (gate off → no handler registered, but references still works) and the self-vivifying Proxy pattern for the adapter tests.

### Optional suggestions for consideration:

1. **LSPQueueManager live-path test (line 1328)** — Consider asserting the full params shape, not just `newName`:
   ```typescript
   expect(dispatch).toHaveBeenCalledWith(
     'rename',
     expect.objectContaining({
       textDocument: { uri: 'file:///Rename.cls' },
       position: { line: 0, character: 0 },
       newName: 'newName',
     }),
   );
   ```
   Current assertion is fine for Phase-0, but future phases will care about position/URI propagation.

2. **LCSAdapter private constructor workaround (line 123)** — The `@ts-expect-error` bypass works but creates maintenance debt. Consider adding a package-private `createForTesting()` factory or extracting `setupProtocolHandlers()` as a standalone function for easier testing.

3. **RenameProcessingService test naming** — "returns null (nothing to rename) for a valid request" could be "returns null when rename target cannot be resolved (Phase-0 stub)" to clarify this is temporary stub behavior.

None of these block approval — ship it once #596 merges! 🚀

PR #616, opened 2026-08-04
feat: resolve missing SObject artifacts natively - W-21360368

peternhale:

@W-21360368@

### Functionality Before

<insert gif and/or summary>

### Functionality After

<insert gif and/or summary>


kylewalke (changes requested):
# Code Review — native SObject artifact resolution (W-21360368)

## Overview

This PR replaces faux-`.cls` SObject generation with **native, server-side SObject symbol-table composition** from org describe payloads. When Apex references an SObject or class absent from the workspace, the desktop extension fetches describes (via `MetadataDescribeService`) and Apex source (via `ConnectionService` Tooling API) from the Salesforce services extension, materializes them into a read-only `apex-org-artifact:` VFS, and ships validated describe payloads over a typed worker-wire protocol to a composer that builds SObject symbol tables natively (the server never compiles a rendered document). A generation token guards org-switch races; the services dependency is optional with graceful degradation.

The architecture and the design doc (`docs/design/org-artifact-resolution.md`) are clear and well-considered. I independently verified the core safety mechanisms — generation-token atomicity, deferred-commit ordering, telemetry name-scrubbing, and fail-closed decode validation — and they are correct. Overall this is strong, well-tested work. My findings concentrate at the **trust boundary** (empty-string validation, payload size caps, dedup-vs-generation) rather than in the design itself.

## CI status

- ✅ **`E2E apex-hover.spec.ts (Web)` — resolved on the branch.** This job failed (empty hover content → full-spec retries → 20-min timeout → cancelled) across the earlier commits. It is now **green** as of `5d8da1a0` ("test: retry unresolved hover scenarios"), which converts the remaining single-shot hover tests to the `hoverOnWordWithRetry` helper. Root cause was single-shot-vs-retry timing on a cold web worker pool (a hover fires before the pool is warm, returns null, and VS Code never re-requests) — not a product regression. Confirmed via the green E2E run on the current branch tip.
- ⚠️ **`Test (windows-latest/ubuntu, …)` — intermittent, unrelated.** The `PriorityScheduler › Starvation Relief` failure (`priority-scheduler-utils.test.ts:452`, `Received: 5 / Expected: < 5`) is a timing-sensitive test **not touched by this PR** (not among the 85 changed files). It has flaked on both windows and ubuntu across runs — pre-existing loaded-runner nondeterminism. Should clear on re-run.
- ⚠️ **E2E timeout budget is tight.** The passing hover run used ~14 of the 20-min ceiling. If any single hover test regresses, the retry-storm re-blows the budget. Worth keeping in mind.

## High severity

**H1 — A schema-valid describe payload with an empty field `type`/`name`/`uri` throws in the composer and kills the data-owner write loop, permanently stalling all subsequent writes. (CONFIRMED)**
- `wireSchemas.ts` `SObjectDescribeFieldSchema` types `name`/`type` and `DefinitionTargetSchema.uri` as bare `Schema.String` — `""` decodes cleanly. `decodeFindMissingArtifactResult` only checks that `describe.name` matches a requested identifier; it never validates field `type`/`uri`.
- The composer's `validateDescribe`/`requireNonEmptyString` (`SObjectSymbolTableComposer.ts:516,522,787,796`) `throw new TypeError` on empty/whitespace strings.
- Both `composeSObjectSymbolTable(...)` call sites in `worker.platform.shared.ts` — the `InstallSObjectArtifacts` data-owner handler and the enrichment self-install `dataOwnerWrite(Effect.gen(...))` — run the composer as a **raw synchronous call inside `Effect.gen`** with no `Effect.try`. A raw throw becomes an Effect **defect**; the write-loop runner recovers only the error channel (`Effect.either`), so the defect propagates out of the `forkDaemon` write loop, the item's `Deferred` never settles, and every subsequent data-owner write (documentOpen/Change) stalls permanently. The sibling `DataOwnerPreloadStandardNamespaces` handler correctly wraps its work in `Effect.tryPromise`; this one declares an `InstallSObjectArtifactsError` shape but never produces it.
- **Fix:** validate non-empty `name`/`type`/`uri` in the wire schema (`Schema.NonEmptyString` / `Schema.pattern`), and/or wrap compose+add in `Effect.try` mapping to the declared typed failure.

**H2 — The in-flight dedup cache is never invalidated on org switch, defeating the generation token and committing stale-org data under the new org. (CONFIRMED code path)**
- `OrgArtifactFileSystem.clear()` bumps `currentGeneration`, but the separate `OrgArtifactAdapter.inFlight` map is never cleared by the org-change watcher (which only calls `fileSystem.clear()`).
- Sequence: Request A (gen 0) puts `promiseA` in `inFlight`; org switch → `clear()` → gen 1, `inFlight` untouched; Request B (gen 1, from the re-fired post-switch diagnostics) dedups onto `promiseA`; `promiseA` resolves with **org-0 data**; B's `commit()` checks `1 === currentGeneration(1)` → true → **old-org content committed under the new org.** The generation token protects only the originating request, not a new-generation request sharing the old promise.
- Not covered by tests — the existing "discards in-flight after org switch" test only clears a single mid-flight request; it never fires a second request that reuses the promise.
- **Fix:** clear/reset `inFlight` from the org-change watcher, or key `inFlight` by generation so cross-generation sharing is impossible.

## Medium severity

**M1 — The 5 MiB payload cap is enforced only on the client org-describe adapter; it is absent at the worker-wire boundary and on the workspace-ComponentSet producer. (CONFIRMED)** `MAX_SOBJECT_WIRE_BYTES` lives only in `org-sobject-adapter.ts`. `MissingArtifactPayloadSchema` / `InstallSObjectArtifacts.payload.artifacts` carry no size or array-length bound, and `workspace-component-set-adapter.ts` `composeWorkspaceArtifact` builds artifacts with no size check. A large workspace CustomObject (or a pathological describe with ~100k fields) crosses into `composeSObjectSymbolTable` unchecked → large allocations + O(n log n) sort on the server. Add a defense-in-depth bound at the schema/composer. (Ship this with H1 — same schema fix.)

**M2 — No standard-field synthesis → false `FIELD_DOES_NOT_EXIST` on `account.Id`. (PLAUSIBLE — depends on provider)** The composer injects only `describe.fields`; `VariableResolutionValidator` emits `FIELD_DOES_NOT_EXIST` for SObject targets. `Account.describe.json` omits `Id`, `CreatedDate`, etc. If the org provider doesn't always send the full standard-field set, `a.Id` flags a false positive. The fixture omitting `Id` is a realism red flag. Either confirm the provider always emits standard fields, or synthesize the universal set in the composer.

**M3 — No negative caching for not-found SObjects → sustained org round-trip amplification. (CONFIRMED)** `SObjectEnrichmentService` clears its in-flight entry on completion and keeps no not-found memory; only `timeout` gets a cooldown. Every hover/definition on a typo'd or genuinely-missing `*__c` name fires a fresh client (and org describe) round-trip. Not an unbounded loop, but sustained amplification. Relatedly, `missingArtifactUtils.ts` has a `recentBackgroundRequests` map that is **written but never read/pruned** — a permanent memory leak that looks like the intended throttle, never wired up. Implement the read side (would also mitigate this) or delete it.

**M4 — Web-worker coordinator detection via `require('node:worker_threads')` defaults to `true` on throw. (PLAUSIBLE — web-only, high-impact)** `isMainThreadCoordinator` returns `true` in the `catch`. Per this repo's PR552 history, bare `node:` requires throw as `__require` inside the bundled web worker — so a web *worker* would be treated as coordinator and mutate the symbol graph off the data-owner, exactly what the guard prevents. The constructor already accepts `options.isCoordinator`; production wiring should inject it rather than probe.

**M5 — `hasCompleteLocalSymbolTable` widens data-owner re-query to all cursor-target names.** `resolveMissingNamesViaDataOwner` replaced a `findSymbolByName().length === 0` short-circuit with `!hasCompleteLocalSymbolTable(...)`. A regular class known via the name index but not resident in this pool worker now returns `false` → re-queried on the data owner. Adds a `QuerySymbolByName` round-trip on every cursor-target prep for non-locally-tabled classes — broader than the SObject use case.

**M6 — `markDeclaredQueryTypeAsSObject` has no initializer-position guard. (CONFIRMED)** `findQueryDeclarationTypeRef` walks up to the first declaration with no check that the SOQL literal is the direct initializer. `MyWrapper w = foo([SELECT Id FROM Account]);` mislabels `MyWrapper`'s type reference as `isSObject`. Latent today (see Low #1), but unsound evidence.

**M7 — Compound platform types mislabeled `isPrimitive: true`. (CONFIRMED)** The composer routes `Address`, `Location`, `Blob`, `Object` through `createPrimitiveType` (`isPrimitive:true`). `TypeCastingValidator` reads `isPrimitive` to select compatibility rules, so `Account.BillingAddress` (type `Address`) would get primitive-cast logic → wrong diagnostics. Gate `createPrimitiveType` to the actual Apex primitive set.

**M8 — `chainResolution` broadens `INHERITANCE`/`INTERFACE_IMPLEMENTATION` resolution beyond this feature's scope. (PLAUSIBLE)** SObjects never appear in `extends`/`implements`, so these two additions to `isClassReferenceContext` are unrelated to the SObject feature and change how superclass/interface references resolve, with no test. Either justify or split out.

**M9 — Hover-retry helper can false-green; web E2E has no org. (from e2e review)** `retryHoverRequest` accepts any non-empty content when `expectedContent` is omitted and doesn't unconditionally dismiss stale hover between retries — unlike goToDefinition's F12-retry, hover has no LSP-synchronous state to anchor on, so it can pass on stale widget content. Separately, the web E2E `--with-services` host has no CLI/org/auth, so the org-artifact resolution path is a no-op there → the feature (desktop-only per the design doc) has **no E2E coverage exercising org resolution end-to-end**.

## Low / cleanup

- **Verify the reference-level evidence subsystem has a consumer.** `SymbolReference.isSObject` + `ReferenceContext.SOQL_FROM_TYPE` and ~200 lines of marking logic in `ApexSymbolCollectorListener` have **no in-repo reader** — every `.isSObject` read is on `TypeInfo`, not `SymbolReference`. Either the consumer lands in an unreviewed slice (confirm) or this is dead code plus a per-parse O(n) tax. Resolving this reclassifies M6/M8.
- **Three of four "shrunk" validators are dead code** (`SObjectTypeValidator`, `CustomEntityValidator`, `CollectionTypeValidator`, `SObjectRecalculateFormulasValidator` are not in `ALL_VALIDATORS`) — this de-risks the "lost 61 lines" concern; their behavioral changes have no live impact.
- `VariableResolutionValidator`'s deliberate mix of `inTypeSymbolGroup` (5 type-ref sites) vs `isClassOrInterfaceSymbol` (8 enclosing-class sites) is **correct but undocumented** — a one-line comment at the enclosing-class sites would stop a future "consistency fix" from reintroducing a bug.
- Adapter nits: `Body.includes('(hidden)')` false-positive on comments/literals; `changeEmitter` never disposed; `servicesCacheHit` is unreachable (never set); `placeholderLifetimeMs === durationMs` (no distinct signal); hand-rolled `AsyncRequestLimiter` is correct but duplicates `Effect.Semaphore`.
- Enrichment path composes the same describe up to 3× (local install + authoritative refetch) with divergent version derivation — idempotent but wasteful, and a version-namespace mismatch could drop the authoritative refetch as `rejected-stale`.

## Test coverage

Genuinely good, not mocks-asserting-mocks. `EnrichmentRoundTrip.node.test.ts` (984 lines) boots the real worker topology and asserts real hover strings / definition URIs — a true round trip (caveats: couples to exact hover formatting + hard-coded positions; the "while change pending" case is timing-flake-prone). Concurrency limiter, SOQL escaping, generation/stale-discard, and name-scrubbed telemetry are exercised against real logic. **Coverage gaps map exactly to the findings:** no test for H1 (empty-string payload), H2 (cross-org dedup reuse), M2 (nonexistent field still flags), M3 (repeated not-found), M4 (web coordinator path), or M6/M7 (non-initializer query / compound-type primitive).

## Recommendation

**Request changes**, but the bar is close. Blocking: **H1** (write-loop defect — highest priority, small fix) and **H2** (cross-org stale commit). Ship **M1** with H1 (shared schema fix). The remaining Mediums are worth addressing but several are latent behind the unverified reference-level subsystem (Low #1) — resolving that question first will let you close M6/M8 cheaply. The web hover CI failure is already resolved on the branch (`5d8da1a0`), and the design (generation token, suppression, telemetry) is sound.

peternhale:
Thanks for the thorough review, Kyle. The review was submitted against `5d8da1a0`; the subsequent hardening commit (`32cea2fa`, rebased as `e2a2de4b`) addresses the blocking findings and most of the medium-severity items:

- **H1 / M1:** wire schemas now require non-empty trimmed SObject names, field names/types, and definition URIs; enforce the 5 MiB artifact limit at the shared wire boundary and workspace producer; and wrap both composer call sites in typed Effect handling. The data-owner queue now completes deferreds from the full `Exit`, including defects.
- **H2:** org generation is propagated into artifact searches and included in the in-flight/cache key, preventing cross-generation deduplication. A regression test covers concurrent requests for the same artifact across generations.
- **M3:** bounded negative caching was added and the unused `recentBackgroundRequests` map was removed.
- **M4:** coordinator ownership is injected explicitly and defaults fail-closed.
- **M5:** the broader data-owner lookup is now limited to incomplete SObject tables.
- **M6:** SOQL declaration marking now requires the literal to be the direct initializer (or enhanced-for expression), with regression coverage.
- **M7:** `Address`, `Location`, `Blob`, and `Object` are no longer marked primitive.
- **M8:** the unrelated inheritance/interface resolution contexts were removed.
- **M9:** hover retries now require expected semantic content, dismiss stale hover widgets on every attempt, and reject placeholder content.

For **M2**, the org path consumes the complete describe returned by `MetadataDescribeService`; the fixture was corrected to include standard fields rather than synthesizing a potentially divergent universal set. The remaining authenticated live-org E2E limitation is understood; the desktop harness now loads Salesforce Services, while CI still cannot depend on org credentials.

The reference-level SObject evidence also has active consumers in cursor dependency loading and `PrerequisiteHelpers` classification.

Could you please re-review the current head?

kylewalke (dismissed):
## Re-review — native SObject artifact resolution (W-21360368)

Re-reviewed after the latest push (branch reworked/rebased onto current main; now 103 files, +10,674/-699). I re-verified every finding from my earlier CHANGES_REQUESTED review against current source. **All blocking items are resolved** and the previously-failing hover CI is now green. This is ready from my side — one non-blocking flake note below.

### Blocking findings — all fixed

- **H1 (compose-throw kills the data-owner write loop) — FIXED, both ways.** `wireSchemas.ts` now types `name`/`type`/`uri` as `NonEmptyWireString` (empty strings rejected at decode), *and* both `composeSObjectSymbolTable` call sites (`worker.platform.shared.ts:1895` enrichment self-install, `:4836` `InstallSObjectArtifacts` handler) are wrapped in `Effect.try` — the handler now maps the throw to the declared `InstallSObjectArtifactsError` instead of an unhandled defect.
- **H2 (in-flight dedup survives org switch → stale-org commit) — FIXED.** `OrgArtifactAdapter.inFlight` is now keyed by generation (`${generation ?? 0}:${kind}:${name}`, `org-artifact-adapter.ts:188`), so a post-switch (gen N+1) request can never reuse a gen-N promise. Structurally impossible now, not just guarded.
- **H4 (`require('node:worker_threads')` coordinator probe defaults to `true` on throw) — FIXED.** The probe is gone; `SObjectEnrichmentService` takes an injected `isCoordinator: () => boolean` defaulting to `() => false` (fail-closed — the opposite of the old dangerous default). Wiring is role-matched: `PrerequisiteOrchestrationService` (coordinator path) passes `() => true`; the request-pool path (`missingArtifactUtils`, used by hover/definition/implementation — all `requestPool` in the dispatch table) correctly defaults to `false` so it never mutates the authoritative graph off-thread.
- **M1 (5 MiB cap only on the org adapter) — FIXED.** The cap is now enforced on the workspace-ComponentSet producer (`workspace-component-set-adapter.ts:34`) and at the wire boundary (`wireSchemas.ts:101`).

### Other prior findings — resolved

- **M2 (false `FIELD_DOES_NOT_EXIST` on `account.Id`) — FIXED at the right layer.** The composer stays a correct pass-through; standard-field completeness is synthesized where the gap actually was — `workspace-component-set-adapter.ts:289` adds the system-field set (`Id`, `CreatedBy/Date`, `LastModifiedBy/Date`, `SystemModstamp`, `IsDeleted`, conditional `OwnerId`/`LastActivityDate`). The org path already gets standard fields from the live describe. `Account.describe.json` now includes `Id`. (Minor: the composer unit fixtures still omit `CreatedDate`/`SystemModstamp`, but standard-field guarantees are enforced and tested at the adapter, so this isn't a correctness gap.)
- **M3 (no negative caching → org round-trip amplification) — FIXED.** `SObjectEnrichmentService` now has a `notFoundUntilByIdentifier` cooldown (`NOT_FOUND_TTL_MS = 3s`): prunes expired, skips identifiers in cooldown, sets cooldown on `not-found` / clears on `resolved`. The dead `recentBackgroundRequests` map was removed entirely.
- **M6 (`markDeclaredQueryTypeAsSObject` no initializer guard) — FIXED.** `findQueryDeclarationTypeRef` now only marks when the SOQL literal is the direct variable-declaration initializer or for-loop source; non-initializer positions return `undefined`.
- **M7 (compound platform types mislabeled `isPrimitive: true`) — FIXED.** The composer defines `SOBJECT_PLATFORM_VALUE_TYPES = {Address, Location, Blob, Object}` and skips `createPrimitiveType` for them, so they get `isPrimitive: false`.
- **Reference-level evidence subsystem (was dead code) — RESOLVED.** `SymbolReference.isSObject` / `ReferenceContext.SOQL_FROM_TYPE` now have live consumers: `worker.platform.shared.ts:1826` (missing-artifact identifier selection), `PrerequisiteHelpers.classifyMissingArtifactIdentifier`, and `chainResolution`. The per-parse marking cost is now justified.
- **H5 (non-compilable `class List<T>`) — N/A.** That was PR #610's stub generator; it does not appear in this PR's diff.

### CI

- `apex-hover.spec.ts (Web)` — the blocking hover regression from my last review — is now **green**, along with every other E2E spec and all unit jobs.
- `apex-goto-definition.spec.ts (Web)` shows **CANCELLED**, but it ran exactly ~20 min (13:39→13:59) and hit the `timeout-minutes` ceiling — the same retry-storm timeout flake this branch's goto/hover specs have hit all week, not a real failure (every other spec in the same run passed). A rerun should clear it; not blocking on the code.

### Verdict

All previously-blocking issues are fixed and independently re-verified against current source. **Approving** — recommend a rerun of the goto-def E2E to get a fully-green board before merge. Nice work turning these around.

kylewalke (dismissed):
## Re-review — native SObject artifact resolution (W-21360368)

Re-reviewed the delta since my earlier approval (which was auto-dismissed when new commits landed). Since then the branch merged current `main` and added E2E fixture/stability work. I diffed `c75cb983..0f3a429e` in full. **Every previously-blocking finding remains fixed, and the new changes are low-risk.** Re-approving.

### What changed since approval

**One production-source change — logging only:**
- `language-server.ts` (both web + desktop `apex/findMissingArtifact` handlers): adapts the log lines to the `params.identifiers[]` array shape (was `params.identifier`), joining names for the message and distinguishing `notFound` (`❌ Could not resolve`) from a real resolve (`✅ Resolved`). Both paths updated identically; no behavioral change to resolution — the `catch` still returns `{ notFound: true }`.

**Everything else is E2E test-infra / fixtures:**
- **Fixtures moved to a real DX project layout** (`force-app/main/default/classes` + a `.cls-meta.xml` companion per class + `sfdx-project.json`). `setup.ts` / `test-server.js` now copy the complete project and prune the old flat root-level `.cls` shape so tests can't pass against stale fixtures. This is the correct shape for ComponentSet-based workspace discovery — loose `.cls` files aren't complete metadata components.
- **`openFileByName` now uses an exact label match** (`getByText(name, { exact: true })`) so `Foo.cls` can't bind to `Foo.cls-meta.xml`; `expandWorkspaceFolders` expands only the fixture's known source path (deliberately not arbitrary folders, to avoid disappearing into the desktop `.vscode-test-user-data` tree).
- **Strict cross-file readiness gate** — `waitForWorkspaceIngestion` gained a `strict` option (throws on timeout after capturing status-bar text; lenient default preserved for the same-file `waitForLSPInitialization` path), and `ApexEditorPage.waitForWorkspaceReady()` uses it before the first cross-file F12. The advanced spec gates every cross-file case on it. This is the exact de-flake change I reviewed and approved on #633.
- **CI `retries: 2 → 1` for Web** — documented rationale: `mode: 'serial'` specs re-run the whole file per retry, so 2 could triple wall-clock and blow the 20-min job timeout (surfacing as `E2E_MISSING`); the workflow's dedicated `--last-failed` sequential pass still covers genuinely-failed tests.
- Opt-in `E2E_APEX_DIAGNOSTICS=1` diagnostics (debug log level + `apex.trace.server: verbose` + Salesforce-Services activation error dump). Gated behind the env var, off by default — no CI-noise cost.

### Prior blocking findings — still fixed (re-verified)

H1 (compose-throw kills write loop), H2 (org-switch dedup keyed by generation), H4 (coordinator probe → injected fail-closed `isCoordinator`), and the M-series (5 MiB cap at wire + producer, standard-field synthesis at the adapter, negative caching, initializer guard, compound-type `isPrimitive`) all remain as I verified them.

### CI

Board is green except the still-running jobs at review time (goto-def / goto-def-advanced / hover Web, Windows unit) — the goto-def suite is exactly what this de-flake work targets, so a clean pass there is the real proof. No unresolved review threads.

### Verdict

**Approving.** The delta is test-infra hardening + a logging-only production change; all substantive concerns from prior rounds stay resolved. Recommend confirming the goto-def Web specs land green before merge.

kylewalke (approved):

PR #619, opened 2026-08-05
test: restore VS Code Web hover locator - W-21360368

kylewalke:
### What does this PR do?

Fixes the `apex-hover.spec.ts (Web)` E2E job on the #616 branch, which fails on every run (empty hover content → retries → 20-min timeout → cancelled), while goto-definition passes.

**Root cause:** the hover page-object locator refactor scoped all hover reads to `.monaco-editor .monaco-hover`, which only exists in **desktop** Monaco. VS Code **Web** renders hover as a `role="tooltip"` overlay that lives *outside* the editor element, so `getHoverContent()` read back empty for every web hover. The LSP itself is healthy — goto-definition (which reads the editor, not a tooltip) passes on the same runs.

**Fix:** restore the `getByRole('tooltip')` web path (as on `main`) alongside the active-editor `.monaco-hover:visible` desktop path, and mirror the same dual lookup in the `page.evaluate` DOM fallback. The branch's anti-stale-widget intent is preserved: `getByRole('tooltip')` excludes hidden nodes and `:visible` scopes the Monaco path.

This targets the `feature/W-21360368-sobject-discovery` branch (PR #616), not `main`.

@W-21360368@

### DIFF
--- a/e2e-tests/pages/HoverPage.ts
+++ b/e2e-tests/pages/HoverPage.ts
@@ -33,15 +33,25 @@ export class HoverPage extends BasePage {
 
   constructor(page: Page) {
     super(page);
-    // Monaco retains hidden hover widgets for inactive editors. Scope all
-    // reads to the visible editor in the active group so first() cannot bind to
-    // a stale diagnostic or semantic hover from another editor instance.
-    this.hoverWidget = page.locator(
-      '.editor-group-container.active .monaco-editor:visible .monaco-hover:visible',
-    );
-    this.hoverContent = this.hoverWidget.locator(
-      '.monaco-hover-content, .hover-contents',
-    );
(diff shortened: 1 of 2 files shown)


kylewalke:
Closing as redundant. The original PR #616 branch already resolved the web hover E2E failure via peternhale's commit `5d8da1a0` ("test: retry unresolved hover scenarios"), which converts the single-shot hover tests to `hoverOnWordWithRetry` — its E2E run (31048299721) is green with the `apex-hover.spec.ts (Web)` job passing under the timeout.

That commit touches only `apex-hover.spec.ts` (no `HoverPage.ts` locator change), which confirms the root cause was the single-shot-vs-retry timing, not the hover locator. This branch was cut one commit earlier (`af8a875`) and so its retry conversion duplicates `5d8da1a0` while adding an unnecessary locator change. Nothing here is needed on top of the current branch tip.

PR #620, opened 2026-08-05
feat(apex-lsp-parser-ast): scope-aware renameLocal with WorkspaceEdit - W-23631077

kylewalke:
## What does this PR do?

Implements **renameLocal** — item 3.1 (Group 3) of the `textDocument/rename` epic: renaming a local variable or method parameter to a `WorkspaceEdit`.

A local rename must edit the declaration and every scope-bound usage **without** touching a same-named local in a sibling or enclosing block. The existing `findOccurrencesInFile` matches on name + reference-context only, so it cannot distinguish two `x` locals in different scopes.

- **New `findLocalOccurrences` op** (`apex-parser-ast`): binds each name+context candidate back to the cursor's declaring symbol via `SymbolTable.resolveVariableAtPosition` (shadowing-safe, innermost-scope wins) and drops candidates that re-resolve to a different declaration. The declaration's own identifier range is always the first edit.
- **`DispatchRename` worker handler**: replaces the Phase-0 no-op with `resolveLocalRename`, which parses the cursor file **standalone** (a local is single-file, so no workspace-wide scan that would surface same-named locals in other files) and builds a `changes`-map `WorkspaceEdit`. A non-local cursor (field/method/type) still returns `null` for later groups.

Wired into the live request-pool path (mirroring `DispatchReferences`), not `RenameProcessingService.processRename`, since rename routes to `requestPool`.

## Tests

- 7 unit tests for `findLocalOccurrences`: sibling-block and inner-block shadowing, usage/declaration cursors, parameter binding, field-cursor → null, empty → null.
- Integration test through the live worker topology: a local rename returns a `WorkspaceEdit` renaming the declaration + all usages, and a sibling-scope same-named local is provably left untouched.

@W-23631077@

## Notes for reviewers

- Stacked on #612 (item 2.1, case-insensitive prefilter) — its true code dependency. Review/merge after #612 and #615 (Group 2). Once #612 merges, this retargets to `main`.
- `RenameProcessingService.processRename` stays the in-process fallback stub; the live rename executes in the worker handler.


peternhale (changes requested):
### Code review

Found 4 issues:

1. Cursor-name resolution can pick an unrelated same-named symbol instead of a qualified reference's target, silently renaming the wrong declaration (e.g. cursor on `this.total` with an in-scope local `total` resolves to the local, not the field).

`refs.find((r) => !r.name.includes('.'))` prefers any non-dotted ref at the position over the dotted one, with no check that it actually belongs to the same expression:

https://github.com/forcedotcom/apex-language-support/blob/44de06424f98892305190d158cfc2f722915cbd5/packages/apex-parser-ast/src/symbols/ops/findLocalOccurrences.ts#L151-L155

2. `resolveLocalRename`'s try/catch only wraps the compile step — the subsequent `findLocalOccurrences` call and edit-building run unguarded, so any throw there becomes an unhandled Effect defect instead of the graceful `null` the rest of the function returns for other failure modes.

https://github.com/forcedotcom/apex-language-support/blob/44de06424f98892305190d158cfc2f722915cbd5/packages/apex-ls/src/worker.platform.shared.ts#L3496-L3498

3. `local.identifierRanges.length === 0` is dead code: `findLocalOccurrences` always pushes the declaration's own range first, so a non-null `local` never has an empty `identifierRanges`. This check reads as if it handles a degenerate case, but it doesn't guard against the real one — a candidate the parser left unresolved (no `resolvedSymbolId`) is silently dropped upstream, so a rename can apply to only some usages while still reporting success.

https://github.com/forcedotcom/apex-language-support/blob/44de06424f98892305190d158cfc2f722915cbd5/packages/apex-ls/src/worker.platform.shared.ts#L3500-L3502

4. No test exercises the name-collision scenario from issue 1 (a local/field name collision at the cursor). The existing "cursor on a field" test only covers a field with no shadowing local in scope, so the regression in issue 1 could ship undetected.

https://github.com/forcedotcom/apex-language-support/blob/44de06424f98892305190d158cfc2f722915cbd5/packages/apex-parser-ast/test/references/findLocalOccurrences.test.ts#L178-L197

Non-blocking notes:

- `SELF_LOADING_REQUEST_TYPES` gates the entire `'rename'` request type, not just local rename — already flagged by the PR's own comment as needing revisit once `renameField`/`renameMethod` land and start depending on the dataOwner graph.

  https://github.com/forcedotcom/apex-language-support/blob/44de06424f98892305190d158cfc2f722915cbd5/packages/lsp-compliant-services/src/queue/LSPQueueManager.ts#L162-L174

- The declaration-token containment scan duplicates position-containment logic that already exists in `positionUtils.ts` (`isPositionInIdentifierRange`/`isPositionWithinLocation`), used elsewhere by `ApexSymbolManager`.

  https://github.com/forcedotcom/apex-language-support/blob/44de06424f98892305190d158cfc2f722915cbd5/packages/apex-parser-ast/src/symbols/ops/findLocalOccurrences.ts#L158-L173

- `resolveLocalRename` re-implements the standalone parse (`SymbolTable` + `FullSymbolCollectorListener` + `CompilerService().compile`) that `scanCandidatesForOccurrences` already does a few hundred lines up; could share a helper.

  https://github.com/forcedotcom/apex-language-support/blob/44de06424f98892305190d158cfc2f722915cbd5/packages/apex-ls/src/worker.platform.shared.ts#L3478-L3488

- `WorkspaceEditResult` hand-rolls a shape that duplicates `WorkspaceEdit`/`TextEdit` already provided by `vscode-languageserver-protocol`, used elsewhere in this codebase.

  https://github.com/forcedotcom/apex-language-support/blob/44de06424f98892305190d158cfc2f722915cbd5/packages/apex-ls/src/worker.platform.shared.ts#L3434-L3447

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

peternhale (dismissed):
## ✅ Approved

Verified the fix on `d135cb46` addresses all four review points:

1. Member-access mis-pick — `FIELD_ACCESS`/`PROPERTY_REFERENCE` refs at the cursor now fall through to `null` instead of being caught by the old `!name.includes('.')` heuristic; local usage is now selected explicitly via a resolved local-kind `VARIABLE_USAGE`. New regression tests (`acc.total`, `this.total` with a shadowing local, plus the companion cursor-on-local case) cover exactly the collision scenario I flagged, and the commit notes the `acc.total` test fails against the old heuristic.
2. `resolveLocalRename` now wraps parse + `findLocalOccurrences` + edit-building in one try/catch, so any throw degrades to `null` instead of escaping as an unhandled Effect defect.
3. The dead `identifierRanges.length === 0` check is gone, replaced by a real guard: `findLocalOccurrences` aborts with `null` if any same-name/kind candidate is left unresolved by the parser, preventing a partial rename from reporting success.
4. New tests cover the name-collision cases.

CI green on the latest commit. Nice work closing out the review loop.

peternhale (approved):

PR #630, opened 2026-08-07
feat(apex-lsp-extension): renameLocal validation + prepareRename - W-23631080

kylewalke:
## What does this PR do?

Implements **item 3.2** (Group 3) of the `textDocument/rename` epic: the validation + prepareRename layer on top of renameLocal (3.1).

- **`validateRenameName` helper** (`apex-parser-ast`) wraps `IdentifierValidator` to check a rename's `newName` against Apex identifier rules; exported from the package barrel so the worker's dynamic import can reach it.
- **Error delivery**: `resolveLocalRename` validates `newName` after resolving the local and, on failure, returns an `{ error: { code, message } }` shape. `LCSAdapter.onRenameRequest` inspects that shape *after* `handleLspRequest` and throws a `ResponseError` — so the client sees the validation failure instead of a silent `null` (the shape survives the worker→coordinator wire; `handleLspRequest`'s catch-all is deliberately not relied on).
- **`canBeRenamed` guard** — always `true` for locals (a local from a standalone parse is never stdlib), structured for later stdlib-aware groups to extend.
- **`onPrepareRename`** handler + a new `DispatchPrepareRename` worker request type wired end-to-end (wire schema, coordinator routing, queue, `GenericRequestHandler`). Returns the identifier range **containing the cursor** (declaration or usage) + placeholder — VS Code requires the returned range to contain the cursor, so returning the declaration range unconditionally would break rename from a usage site. `renameProvider` flips to `{ prepareProvider: true }` in `DEVELOPMENT_CAPABILITIES` now that the handler exists.

## Tests

- `validateRenameName` unit (11), rename capability-gate (5, incl. the ResponseError-throw path + `prepareProvider` gating), worker-topology integration (6, incl. invalid-name error result + prepareRename cursor-on-usage / cursor-on-declaration / non-local → null), capability snapshot alignment (25).
- Also fixes the shared `vscode-languageserver/browser` test mock to export `ResponseError` as a real class (production throws `new ResponseError(...)`; the mock previously omitted it, which surfaced as a "not a constructor" TypeError only in the full suite, not in isolation).

@W-23631080@

## Notes for reviewers

- **Stacked on #620** (item 3.1). Review/merge after #620. Once #620 merges this retargets to `main`.
- The prepareRename cursor-containment behavior was a review catch (an earlier cut returned the declaration range unconditionally); the integration tests now pin cursor-on-usage vs cursor-on-declaration.


peternhale (changes requested):
## Review: Request Changes

Reviewed alongside #631 (stacked). Overall this is solid, well-tested work — the `{ error }` vs `null` distinction in `resolveLocalRename`, the post-`handleLspRequest` ResponseError conversion, and the cursor-containment fix for `prepareRename` are all correctly reasoned and backed by targeted tests (including the off-by-one boundary case). Two things I'd like addressed before merge:

### 1. Dead map entry in `GenericRequestHandler` (Low)
`packages/lsp-compliant-services/src/registry/GenericRequestHandler.ts:74` adds `prepareRename: 'processPrepareRename'` to `getMethodName`'s map, but:
- No `ServiceConfiguration` entry registers a handler for `prepareRename` (only `rename` → `RenameProcessingService.processRename` exists).
- No service anywhere implements `processPrepareRename`.
- In the worker-pool topology, `prepareRename` is routed directly via `WorkerCoordinator`/`DispatchPrepareRename`, bypassing this registry entirely — so today this is unreachable.
- If the local (non-worker) fallback path is ever exercised for `prepareRename`, it actually fails one line earlier with `No handler registered for request type: prepareRename` (`LSPQueueManager.ts:415`), never reaching this method lookup.

Either wire a real `processPrepareRename` on `RenameProcessingService` for topology parity, or drop the map entry so it doesn't imply support that doesn't exist.

### 2. `canBeRenamed(_declaration: any)` typing (Nit, would like it addressed)
`packages/apex-ls/src/worker.platform.shared.ts:208` uses `any` instead of `ApexSymbol` (already available via `findLocalOccurrences`'s return type in the same module). Since this is explicitly a stub extension point for future stdlib-aware groups, typing it now would make the contract self-documenting and catch misuse at compile time when it's extended.

### Non-blocking observation
`resolvePrepareRenameForLocal`'s cursor-containment check (`endColumn > cursorChar`, half-open) intentionally differs from the pre-existing `isPositionInIdentifierRange` utility in `positionUtils.ts` (`character <= identifierRange.endColumn`, closed-inclusive), which is used elsewhere in the same rename pipeline (e.g. `findLocalOccurrences`'s declaration-token scan). The new code's semantics look correct given how `endColumn` is constructed (`start + text.length`), but the shared utility having different boundary semantics is a latent inconsistency worth a follow-up ticket — not blocking here.

peternhale (changes requested):
## Re-review: changes requested

The two requested fixes are addressed:
- `prepareRename` was removed from the dead local-handler map.
- `canBeRenamed` now takes `ApexSymbol`.

One new blocker remains:

### `prepareRename` still takes the cold-read fallback path

`DispatchPrepareRename` recompiles the active document from the request’s live `content`, exactly like `DispatchRename` (`resolvePrepareRenameForLocal` does a standalone parse). However, `SELF_LOADING_REQUEST_TYPES` includes `rename` but not `prepareRename` in `packages/lsp-compliant-services/src/queue/LSPQueueManager.ts:156-174`.

On a cold open or edit where the data-owner readiness latch is not ready, `prepareRename` therefore skips worker dispatch and falls through to the coordinator-local registry. That registry deliberately has no `prepareRename` handler after the earlier fix, so the client receives `No handler registered for request type: prepareRename` instead of a valid result or `null`. Because `prepareProvider` is now advertised, VS Code will use this request before the rename UI opens, making F2 fail precisely in the cold-start scenario that `rename` was fixed to support.

Please add `prepareRename` to `SELF_LOADING_REQUEST_TYPES` and add the analogous queue-level regression test: when readiness reports not-ready, `submitPrepareRenameRequest` must not await the latch and must dispatch to the request-pool worker.

The existing worker-topology tests exercise direct dispatch and do not cover this queue gate/fallback segment.

kylewalke:
Addressed the re-review blocker in 42cf5360.

**`prepareRename` was taking the cold-read fallback path** — fixed by adding `prepareRename` to `SELF_LOADING_REQUEST_TYPES` alongside `rename`. `DispatchPrepareRename` recompiles the cursor file standalone from `req.content` (via `resolvePrepareRenameForLocal`), so it never reads the dataOwner graph and must skip the gate — otherwise a not-ready verdict on cold open drops it to the coordinator-local registry, which has no `prepareRename` handler, surfacing `No handler registered for request type: prepareRename` right when VS Code fires it before opening the rename UI.

Added the analogous queue-level regression test (`dispatches prepareRename to the worker when symbols not ready (self-loading, W-23631080)`): asserts `awaitSymbolDataReady` is **not** awaited and the pool worker **is** dispatched on a not-ready verdict. Verified it fails without the one-line source change. Full suite green (5088 passed, 0 failed).

peternhale (approved):
Re-reviewed the requested changes at 42cf5360. The dead local handler entry was removed, canBeRenamed is typed with ApexSymbol, and prepareRename now follows the self-loading worker path with a cold-read queue regression test. Approved.

PR #631, opened 2026-08-07
test(apex-lsp-extension): renameLocal e2e spec + rename helper - W-23631082

kylewalke:
## What does this PR do?

Implements **item 3.3** (Group 3) of the `textDocument/rename` epic: end-to-end e2e coverage for renameLocal, plus a reusable rename helper.

- **`ApexEditorPage.rename(newName)`** — a Playwright page-object helper driving VS Code's Monaco rename UI via **F2** (not the command palette, which would collapse the editor selection). It focuses the rename input and `fill()`s the new name (rather than `type()`-ing over the prefill selection), which is robust against a focus/selection race that otherwise leaks keystrokes into the editor. Reused by later groups (4.3 / 5.4 / 6.4).
- **`RenameLocalSample.cls`** fixture — a local variable used across a method, a sibling method with a same-named local (scope-safety proof), and a parameter. Kept under ~30 lines so the whole file fits the Monaco viewport on CI (`getContent()` is viewport-limited).
- **`apex-rename-symbol.spec.ts`** — two cases: renaming a local variable (all occurrences renamed, sibling-scope same-named local untouched) and renaming a parameter. Both **pass end-to-end against a real VS Code Web host**.
- Adds the spec to the Web CI matrix in `e2e-tests.yml` (a hardcoded allowlist — new specs don't run otherwise).

## Tests

Both e2e cases verified passing end-to-end locally against VS Code Web (2/2). This exercise also confirmed the 3.1 self-loading dispatch fix (W-23631077) applies the WorkspaceEdit in a real client, not just in the worker-topology integration test.

@W-23631082@

## Notes for reviewers

- **Stacked on #620** (item 3.1). Review/merge after #620. Once #620 merges this retargets to `main`.
- The rename helper's focus+`fill()` approach and the sub-viewport fixture size both come from debugging a real e2e failure where the edit appeared not to apply — worth preserving as-is.


peternhale (changes requested):
## Review: Request Changes

Good e2e coverage — the `rename()` Playwright helper's design notes (why `fill()` over `type()`, why the retry loop is scoped only around F2+visibility) reflect real debugging of observed flakiness rather than speculative hardening, and the fixture's viewport-size constraint is called out explicitly. One thing I'd like addressed before merge, plus a minor observation:

### 1. CI test-summary tracking not updated (Low-Medium)
`.github/workflows/e2e-tests.yml:276-287` — the `EXPECTED_SPECS` array the Test Summary job uses to flag "spec produced no results" (crash/build failure before tests ran) is explicitly commented `# keep in sync with the matrix above`, but wasn't updated to include `apex-rename-symbol.spec.ts`, unlike the actual test matrix at line 44 (which this PR did update correctly). Aggregate pass/fail counts still work fine (they scan all `junit.xml` files unconditionally), but if this new spec's job crashes before producing results, it will silently NOT show up under "Specs With No Results" in the summary — reducing visibility into that failure mode specifically for the new spec. Please add it to the `EXPECTED_SPECS` array too.

### Non-blocking observation
Both tests in `apex-rename-symbol.spec.ts` call `openFile('RenameLocalSample.cls')` independently, and since `ApexEditorPage.openFile` re-activates an already-open tab rather than reopening from disk, test 2 inherits test 1's unsaved edit (`total` → `renamed`) since the rename is never saved. It doesn't collide with test 2's `value`/`inputParam` rename today, but it's a latent coupling between the two tests worth keeping in mind if the fixture or these tests change later.

kylewalke:
Addressed the review in dfd426b2.

**Test Summary tracking not updated** — added `apex-rename-symbol.spec.ts` to the `EXPECTED_SPECS` array in `.github/workflows/e2e-tests.yml` (line 290), so a crash-before-results for the new spec will now surface under "Specs With No Results" instead of silently vanishing.

On the non-blocking observation (test 2 inheriting test 1's unsaved `total`→`renamed` edit via `openFile` re-activating the open tab): confirmed it doesn't collide today since test 2 renames `value`/`inputParam`. Noted as latent coupling to watch if the fixture changes; left as-is to keep this PR scoped.

peternhale (approved):
Re-reviewed the requested change at dfd426b2. apex-rename-symbol.spec.ts is now included in both the E2E matrix and Test Summary EXPECTED_SPECS tracking. Approved.

PR #633, opened 2026-08-10
fix: de-flake apex-goto-definition Web E2E - W-23799620

kylewalke:
## What does this PR do?

De-flakes the `apex-goto-definition` Web E2E suite, which was intermittently timing out and blocking unrelated PRs (surfacing as `E2E_MISSING` after the 20-min job timeout cancelled the job). Three coordinated changes:

**A — Retry amplification (`createWebConfig.ts`):** CI `retries` 2 → 1. Serial-mode specs (`mode: 'serial'`) re-run the *entire file* on each retry, so `retries=2` meant up to 3× the whole-suite wall-clock — enough for the largest suite to blow past the 20-min job timeout and get cancelled. The workflow already runs a dedicated sequential `--last-failed` retry pass after the parallel run, so one in-job retry is sufficient.

**B — Suite split:** The 29-test `apex-goto-definition.spec.ts` is split into a 20-test basic suite plus a new 9-test `apex-goto-definition-advanced.spec.ts` (advanced + cross-file scenarios — the slowest, most race-prone tests). A flake in a cross-file test now re-runs 9 tests, not 29. The new spec is added to the workflow matrix and the `EXPECTED_SPECS` summary list.

**C — Cross-file race:** Every advanced/cross-file test now gates on `waitForWorkspaceReady()` (full workspace ingestion) after `waitForLanguageServerReady()`. Cross-file resolution needs full ingestion, not just LSP init; a definition request issued mid-ingest returns `null` and VS Code does not auto-retry it. Added `ApexEditorPage.waitForWorkspaceReady()`.

@W-23799620@

## Functionality Before

`apex-goto-definition` Web E2E flakes on cross-file resolution; serial re-runs amplify runtime past the 20-min job timeout; the cancelled job reports `E2E_MISSING` and blocks the PR.

## Functionality After

Deterministic gating on workspace ingestion for cross-file tests, a smaller blast radius per flake (9 vs 29 tests), and bounded in-job retry runtime.


e2e-tests/pages/ApexEditorPage.ts
```diff
@@ -84,6 +87,23 @@ export class ApexEditorPage extends BasePage {
...
+   * `waitForLanguageServerReady()` only gates on LSP init + editor readiness,
+   * which is enough for same-file navigation but races cross-file resolution:
+   * a definition request issued mid-ingest returns null and VS Code does not
+   * auto-retry it. Cross-file goto tests should gate on THIS before the first
+   * cross-file F12.
+   *
+   * @param timeout - Max wait in ms (defaults to 45s desktop / 30s web)
+   */
+  async waitForWorkspaceReady(timeout?: number): Promise<void> {
+    await waitForWorkspaceIngestion(this.page, timeout);
```
peternhale: This wrapper cannot make cross-file tests deterministic while `waitForWorkspaceIngestion()` catches and suppresses its timeout. When ingestion has not reached the ready state, that helper logs a warning and resolves, so callers immediately issue F12 against an incomplete workspace -- the race this PR is meant to eliminate. Please make the wait throw after capturing the status-bar text, or add a strict variant and use it from the cross-file tests, so these tests fail/retry instead of continuing before ingestion completes.

peternhale (changes requested):
The suite split and retry adjustment are well-scoped, but the new workspace-readiness calls do not enforce readiness yet.

Non-blocking follow-up: `e2e-tests/README.md` still says CI has two retries. Please update it to describe the new one-retry Web configuration (and distinguish Desktop if its two-retry configuration is intentional).

peternhale (approved):
Re-reviewed the follow-up commit. The strict workspace-ingestion gate now fails and retries instead of proceeding on timeout, and the retry documentation correctly distinguishes Web from Desktop.

PR #634, opened 2026-08-11
feat(apex-lsp-compliant-services): data-owner CheckMemberConflicts query for rename - W-23631128

kylewalke:
## What does this PR do?

Adds a data-owner-side **`CheckMemberConflicts`** query — the shared hierarchy-aware member-name conflict check that renameField (4.2) and renameMethod (5.3) consume for rename validation. This is WI 4.0 of the `textDocument/rename` epic, gating all of Group 4 (renameField).

Given a defining type FQN + a proposed new name, it walks the type family and returns a verdict:

```
{ conflict: boolean, conflictingTypeFqn?: string, reason?: 'same-type' | 'ancestor' | 'descendant' }
```

Following jorje's `FieldRenameHandler` rules: **same-type** (no modifier filter) → **ancestor** (excludes private/default-visibility members) → **descendant** (all members; reported only when the renamed member is non-private).

## Why the data-owner (not the pool worker)?

The descendant conflict walk needs the **complete** subtype cone. On the request-pool worker, `findSubtypes`/`findSupertypes` read a transient per-request subset graph whose drain-on-read never fetches unloaded files (`ApexSymbolRefManager.ts:2467-2478`) — so a pool-worker walk silently misses descendants in files the request didn't load, greenlighting a destructive rename. The complete cone only exists on the data-owner after `endWorkspaceLoadSession` resolves inheritance edges. Pool-worker rename kinds call this query via coordinator assistance, mirroring the proven `FindOccurrenceCandidates` pattern.

## Notable correctness details

- **Completeness guard:** fails fast if a workspace load session is still active (graph not yet complete), rather than returning a false "no conflict".
- **Apex default visibility:** a member with no access modifier is `Default` visibility, which is effectively private for inheritance — treated as private in the ancestor filter. (Callers computing `isRenamedMemberPrivate` for the descendant gate must do the same; documented in code.)
- **Reusable `walkTypeFamily` helper:** the ancestor/descendant traversal is factored out so renameMethod (5.2) can extend it with interfaces + implementors.

## Scope

Verdict-only. The renameMethod edit-target fan-out (`getInstanceMethodLocations` equivalent) is intentionally **not** here — it's method-specific and belongs to 5.2, layered on this query's family walk.

## Tests

10 integration tests (`CheckMemberConflicts.node.test.ts`): all three verdicts, private + default-visibility gating, case-insensitivity, interface ancestors, and the failure path (non-existent type → typed error).

@W-23631128@


peternhale (changes requested):
Request changes.

1. **Critical** `packages/apex-ls/src/worker.platform.shared.ts:5067`
`CheckMemberConflicts` reads data-owner symbol tables at `public-api` detail. Those omit private, protected, and default-visibility members, but the conflict rules require all same-type members, protected ancestor members, and all descendant members. This produces false negatives and can approve invalid renames. Use a full-declaration member view/index, and add coverage for private same-type, protected ancestor, and private/default descendant conflicts.

2. **Medium** `packages/apex-ls/src/worker.platform.shared.ts:5091`
The same-type lookup cannot exclude the member being renamed. A no-op or case-only rename therefore conflicts with itself. Carry a stable target member identity/signature and exclude it, or make a caller-side case-insensitive no-op short-circuit an explicit tested contract.

3. **Medium** `packages/apex-ls/src/worker.platform.shared.ts:4427`
The hierarchy traversal inherited by this correctness-sensitive query deduplicates types by lowercased short name. Distinct nested types such as `OuterOne.Child` and `OuterTwo.Child` can collapse, skipping a valid family branch and its conflict. Deduplicate by stable symbol identity or normalized FQN, with a repeated-inner-name regression test.

4. **Medium** `packages/apex-ls/test/integration/CheckMemberConflicts.node.test.ts:711`
The interface-ancestor test is masked by a direct `interfaceField` declaration on `Implementation`, so it passes through the same-type branch rather than exercising ancestor detection. The interface-property fixture is invalid Apex. Replace it with a valid interface method and ensure the implementing type has no same-named declaration.

Note: the wire contract exposes `memberKind: method`, but the implementation currently performs name-only matching. Before a method caller consumes this request, extend the contract to include signature data and test overloads, or keep the query field-only.

kylewalke:
Thanks @peternhale — addressed all four findings in `73e3729f`:

**1 (Critical) — public-api detail omitted non-public members.** `hasMemberNamed` now enriches each inspected type's file to `full` detail (`enrichToLevel` using the data-owner's retained document text) before reading its block members, so private/protected/default members are visible. Added coverage: private same-type collision, protected-ancestor collision, and private-descendant collision — each fails without the enrichment.

**2 (Medium) — same-type self-match on no-op/case-only rename.** Added an optional `currentName` to the `CheckMemberConflicts` payload; when present and case-insensitively equal to `newName`, the query short-circuits to `conflict:false` (mirrors the jorje short-circuit). Backward-compatible. Tests for `total→Total` and `total→total`. (Follow-up on the stacked branch: the `resolveFieldRename` consumer will pass `currentName: target.name`.)

**3 (Medium) — hierarchy dedup by lowercased short name.** `collectTransitiveTypes` now dedups the visited set by stable symbol `.id` instead of `.name.toLowerCase()`, so `OuterOne.Child` and `OuterTwo.Child` stay distinct branches. Added a repeated-inner-name regression that fails under the old name-based dedup.

**4 (Medium) — interface-ancestor test was masked.** Rewritten to use a valid interface **method** (`void interfaceMethod();`), with the implementing type declaring no same-named member, querying `memberKind:'method'` and asserting `reason:'ancestor'`. It now genuinely exercises the ancestor branch.

**Note on the method-signature contract:** unchanged — the query stays name-only matching for `memberKind:'method'`; signature-equivalence is deferred to the renameMethod story (5.3) when a method caller first consumes this.

Full pre-commit suite green (5339 tests); CheckMemberConflicts 15/15.

peternhale (changes requested):
Request changes.

1. P1 - Full-detail conflict checking still fails open when retained source is unavailable.

In packages/apex-ls/src/worker.platform.shared.ts around lines 5395-5421, hasMemberNamed attempts enrichToLevel only when storage.getDocument(type.fileUri)?.getText() is truthy. When the source is missing, it proceeds with whatever symbol table is loaded. If that table is still public-api detail, protected/private/default members remain absent and the query can return conflict:false, recreating the original destructive false-negative. Please fail with CheckMemberConflictsError unless a complete member view is known to be available, or encode and verify a storage invariant that guarantees full enrichment for every inspected family type. The query must preserve uncertainty rather than treat incomplete semantic state as no conflict.

2. CI blocker - the new CheckMemberConflicts suite is not reliable under the required matrix.

Four CI jobs are currently failing. Ubuntu lts/-1 and the Windows jobs time out in CheckMemberConflicts.node.test.ts at the 30-second per-test limit under coverage. The focused suite passes locally (15/15), but the required CI workflow remains red. Please make the topology tests fit the coverage runtime reliably, or adjust their structure/time budget with justification, and rerun the full matrix.

The currentName short-circuit, stable-symbol-id hierarchy deduplication, non-public member coverage, and corrected interface-ancestor fixture otherwise address the requested cases.

peternhale:
Correction to my just-submitted review: commit f064eeda landed before the review was submitted and replaces the per-test topology setup with one shared scoped topology. That directly addresses the prior CI timeout pattern, and the new matrix is currently running. Please treat review item 2 as superseded pending the new CI result.

The P1 fail-open semantic-state finding in item 1 remains the requested-change blocker.

kylewalke:
Addressed both in `ec5c115f` (P1) — the CI blocker was already fixed in the prior push.

**P1 — fail open on missing source.** `hasMemberNamed` now **fails closed**: if a family type's file is not already at `full` detail, it must be enrichable from retained source *and* verified to reach `full`, otherwise the query fails with `CheckMemberConflictsError` ("Cannot verify member conflicts … non-public members cannot be inspected"). It no longer reads a possibly-`public-api` table and returns `conflict:false`. Uncertainty is preserved rather than treated as no conflict. The *enrichment-happens* invariant is verified by the existing private/protected/default-member tests — those are false negatives without enrichment — and the new branch guards the source-unavailable case.

**CI blocker — CheckMemberConflicts matrix timeouts.** Fixed in `f064eeda` (pushed before this review landed): the suite spun a fresh worker topology per test (~30s each × 15 → ~450s), timing out on Windows/lts-1. It now shares one topology + one combined workspace across the describe block (Scope opened in `beforeAll`, closed in `afterAll`), running in **~2s**. Every fixture has a unique type name and each query inspects only its own type family, so the combined workspace is behavior-equivalent. The current run is green on ubuntu (all) and Windows current/lts*.

Full pre-commit suite green; CheckMemberConflicts 15/15.

peternhale (changes requested):
Request changes.

1. **P1 - The full-detail guard trusts a stale side-channel and can still fail open.**

In `packages/apex-ls/src/worker.platform.shared.ts` around lines 5102-5126, `hasMemberNamed` decides whether enrichment is needed and verifies its result through `getDetailLevelForFile()`. That value comes from `ApexSymbolManager.fileDetailLevels`, which is set by `enrichToLevel` but is not downgraded or reconciled when a newer symbol table replaces the enriched table.

I reproduced the unsafe sequence against this head: enrich version 1 to `full`, then register a newer version 2 `public-api` table. The canonical `SymbolTable.getDetailLevel()` is then `public-api`, while `getDetailLevelForFile()` still returns `full`. `CheckMemberConflicts` therefore skips enrichment and reads the public-only canonical table, allowing private/protected/default members to disappear from the check and recreating the destructive false-negative this commit intends to close.

Please validate the canonical table's own `getDetailLevel()` before and after enrichment, and ensure enrichment cannot no-op because its separate tracker is stale (for example, reconcile/reset that tracker on replacement or use a full compile path whose canonical result is verified). Add a regression covering full enrichment followed by a newer public-API replacement.

The retained-source failure path and the shared-topology CI fix otherwise address the previous review.

kylewalke:
Good catch — you're right that the previous guard trusted the stale side-channel. Fixed in `3c10867e`.

**Root cause:** `getDetailLevelForFile()` reads `ApexSymbolManager.fileDetailLevels`, which `enrichToLevel` sets but which is never reconciled when a newer lower-detail table replaces the enriched one — so it can read `full` while the canonical `SymbolTable.getDetailLevel()` a reader observes is back to `public-api`.

**Fix — key everything off the canonical table's own `getDetailLevel()`:**
- `enrichToLevel` now decides its no-op from the **canonical table's** `getDetailLevel()`, not the side-channel, so it can't skip when the observable table is below target (the stale-tracker no-op is gone).
- `hasMemberNamed` now validates the **canonical** `typeTable.getDetailLevel()` **before and after** enrichment and fails closed with `CheckMemberConflictsError` if it can't reach `full`.

**Regression** (`ApexSymbolManager.replacementSemantics.test.ts`): enrich a file to `full`, then register a newer `public-api` table (leaving the tracker stale at `full`) — the test asserts the divergence (canonical `public-api`, tracker `full`), then that `enrichToLevel` **re-enriches** the canonical back to `full`. It fails without the fix (enrichment no-ops on the stale tracker and the canonical stays `public-api`).

Full pre-commit suite green; CheckMemberConflicts 15/15; replacementSemantics 10/10.

peternhale (changes requested):
Request changes.

1. **P1 - Full-detail conflict checking still treats a syntax-broken recovered parse as complete.**

In `packages/apex-parser-ast/src/symbols/ApexSymbolManager.ts` around lines 7666-7678, `enrichToLevel` ignores `CompilerService.compile(...).errors`: whenever a result table exists, it registers that recovered table and records the target detail level. `CheckMemberConflicts` around `packages/apex-ls/src/worker.platform.shared.ts:5456-5490` then validates only `typeTable.getDetailLevel() === 'full'` before trusting the member set.

I reproduced this against the current stacked head with a class containing a normal field plus `visible = ;` in a malformed method. Compilation returned parser errors, but after `enrichToLevel(..., 'full', ...)` the canonical table still reported `full`. A recovered parse can omit declarations or scopes in the damaged region, so the query can return `conflict:false` from incomplete semantic state despite the new fail-closed comments.

Please make enrichment surface/reject syntax errors (or mark parse completeness honestly) and have `CheckMemberConflicts` fail with `CheckMemberConflictsError` whenever an inspected canonical table has syntax errors or is explicitly incomplete. Add a regression exercising the diagnostic-return path, not only missing source or missing table.

The previous stale `fileDetailLevels` finding is fixed: the canonical-level guard and replacement regression work as intended, and CI is green.

kylewalke:
Addressed the 2026-08-21 P1 (syntax-broken recovered parse treated as complete) in `c1eb2f02`.

**Root cause:** `enrichToLevel` ignored `CompilerService.compile(...).errors`. A malformed source still RECOVERS a SymbolTable that reports `getDetailLevel() === 'full'`, so the detail-level guard in `hasMemberNamed` passed and the (possibly declaration-dropping) member set was trusted. Reproduced exactly your case: `public class brokenType { public String existingField; public void broken() { visible = ; } }` → 1 syntax error, yet the recovered table reports `full` with the field present.

**Fix — mark parse completeness honestly, then fail closed:**
1. `ApexSymbolManager.enrichToLevel` (packages/apex-parser-ast/src/symbols/ApexSymbolManager.ts) now counts **syntax** errors from the compile result (`ErrorType.Syntax` only — benign **semantic** cross-file unresolved-reference errors are excluded so they don't poison completeness) and stamps the table's metadata: `parseCompleteness: 'incomplete'` when syntax errors are present, `'complete'` when clean. Because `registerSymbolTable` may keep an existing instance as canonical (merge fallback), it re-stamps whatever `getSymbolTableForFile` returns, so the table the reader observes carries the signal truthfully.
2. `CheckMemberConflicts.hasMemberNamed` (packages/apex-ls/src/worker.platform.shared.ts) now, after the detail-level guard, fails with `CheckMemberConflictsError` when the canonical table's `parseCompleteness === 'incomplete'`. It keys strictly off `parseCompleteness` (the syntax-derived signal), **not** the broader `hasErrors` flag — the compilation worker sets `hasErrors` from all errors including benign semantic ones, so keying off it would over-decline valid renames.

**Regressions (both the requested diagnostic-return path):**
- Unit (replacementSemantics): a syntax-broken enrich → `parseCompleteness: 'incomplete'`; a clean enrich that references an unresolved cross-file type (benign semantic error) → NOT incomplete.
- Integration (CheckMemberConflicts.node): a `brokenType` fixture whose FQN resolves (class recovered) → query rejects via the incomplete-parse guard, proving it's not a type-not-found path.

CheckMemberConflicts 16/16; replacementSemantics 11/11; full pre-commit matrix green (0 failed). The other 15 conflict tests (valid files carrying benign semantic errors) still classify correctly, confirming no semantic-error over-decline.

kylewalke:
Follow-up: a fresh from-scratch review surfaced a further **P1** — fixed in `6f001071`.

**The hole:** the incomplete-parse guard I added to `hasMemberNamed` only fires when the canonical table's `parseCompleteness === 'incomplete'`, and only `enrichToLevel` stamped that. But a full-detail table can reach the data owner via a **different producer**: the request-pool cursor recompile (`recompileCursorFileAtFullDetail`) does a raw `compile` + `addSymbolTable` and never stamps completeness, so it lands at `parseCompleteness='unknown'`. `writeBackEnrichedSymbols` serializes that metadata verbatim and `UpdateSymbolSubset` rehydrates it, so a syntax-broken file edited in the buffer (hover/definition/references trigger the full-detail recompile + write-back) becomes `getDetailLevel()==='full'` + `parseCompleteness='unknown'` on the data owner. `hasMemberNamed` then skips enrichment (already full) and passes the guard (`'unknown' !== 'incomplete'`) → reads a truncated member set → can approve a colliding rename. I confirmed the parser drops declarations in the damaged region at full detail (e.g. a member after an unclosed brace) with completeness left `unknown`.

**Fix (root cause, at the producer):** `recompileCursorFileAtFullDetail` now stamps `parseCompleteness` honestly from **syntax** errors (Syntax only — benign semantic cross-file errors don't count), mirroring `enrichToLevel`. It is the pool's only full-detail producer, so the honest signal now flows through write-back → `UpdateSymbolSubset` → the canonical data-owner table, where the existing `=== 'incomplete'` guard catches it. I keyed the fix off completeness (not a blanket 'require complete') so clean files reaching full via write-back are NOT over-declined.

**Regressions added** (recompileCursorFileAtFullDetail.node): a syntax-broken source → the table handed to `addSymbolTable` is `parseCompleteness:'incomplete'`; a clean source with a benign unresolved cross-file ref → NOT incomplete. Combined with the existing `brokenType` decline test, both producer and consumer halves are covered. Full matrix green (0 failed).

Two lower-severity notes from the same review, for the record (not fixed here):
- **isRenamedMemberPrivate trusted from the wire:** the descendant gate relies on a caller-computed boolean; the consumer lives on the 4.2/086 branch. Worth having the query self-verify the renamed member's visibility (fail closed on mismatch) there — I'll carry it as a follow-up on that branch rather than guess at a caller that doesn't exist here yet.
- **Lexical prefilter** can miss references in an unsaved, non-cursor buffer whose stored text doesn't mention the field — inherent to the content prefilter; noted for the epic.

peternhale (changes requested):
Request changes.

1. **P1 - Full-detail conflict checking still trusts unknown parse completeness.**

In `packages/apex-ls/src/worker.platform.shared.ts` around lines 5131-5203, `hasMemberNamed` skips enrichment whenever the canonical table reports `getDetailLevel() === 'full'`, then rejects only `parseCompleteness === 'incomplete'`. A full table with `parseCompleteness === 'unknown'` therefore passes the guard and its member set is trusted.

The latest commits stamp the two known full-detail producers, but the correctness-sensitive consumer still fails open at the protocol boundary: `SymbolTable` defaults completeness to `unknown`, and `UpdateSymbolSubset` deserializes and installs a full-detail table without requiring a complete signal. An unstamped or newly introduced producer can therefore recreate the exact destructive false-negative this series is intended to close.

Please require `parseCompleteness === 'complete'` before reading members. If a full/unknown table has retained source, re-enrich/recompile it to establish completeness; otherwise fail with `CheckMemberConflictsError`. Add an integration regression that installs a full/unknown table through `UpdateSymbolSubset` and verifies the query declines rather than returning `conflict:false`.

2. **Required CI is red.**

The current E2E run has a failed `Test Summary`: `apex-goto-definition.spec.ts` failed the third navigation in the sequential go-to-definition test on retry (`Go-to-definition did not navigate to line 183`), after which the job was cancelled. Please obtain a green required matrix or address the failure before approval.

The syntax-error stamping in both `enrichToLevel` and `recompileCursorFileAtFullDetail` otherwise addresses the specific known producer paths from the prior review.

kylewalke:
Addressed the 2026-08-24 P1 (consumer still trusts `parseCompleteness === 'unknown'`) in `e2b2515c`.

**Fix — require `'complete'`, re-enrich to establish it, fail closed only if unattainable:**
1. `CheckMemberConflicts.hasMemberNamed` (worker.platform.shared.ts) now requires the canonical table to be **both full-detail AND `parseCompleteness === 'complete'`** before reading members. `SymbolTable` defaults completeness to `'unknown'`, so trusting anything other than an explicit `'complete'` fails open — exactly the protocol-boundary gap you flagged (a raw cursor recompile installed via `UpdateSymbolSubset` lands full + `'unknown'`).
2. When the table is not full+complete, it (re-)enriches from retained source to **establish** completeness. `ApexSymbolManager.enrichToLevel` now no longer short-circuits a table that is already at/above the target level when its completeness is still `'unknown'` — it re-parses and stamps `'complete'`/`'incomplete'` from SYNTAX errors. This is what avoids over-declining: a clean file that reached full via an unstamped producer re-enriches to `'complete'` and answers normally; a syntax-broken one re-enriches to `'incomplete'` and declines. If retained source is unavailable, it fails with `CheckMemberConflictsError`.

**Regressions added:**
- integration (new `CheckMemberConflictsWriteBack.node.test.ts`): installs a full/`unknown` table **through `UpdateSymbolSubset`** (accept path exercised, `upd.accepted === true`) and asserts (a) a broken-source file → the query DECLINES with `CheckMemberConflictsError`, and (b) a clean-source file → the query proceeds (`conflict:false`, no over-decline).
- unit (replacementSemantics): a full table with `parseCompleteness: 'unknown'` is re-parsed by `enrichToLevel` to establish `'complete'` rather than short-circuiting.

CheckMemberConflicts 16/16; replacementSemantics 12/12; new write-back suite 2/2; full pre-commit matrix green (0 failed).

Re **required CI red**: the failure is `apex-goto-definition.spec.ts` failing the third navigation on retry (`did not navigate to line 183`) — a known-flaky E2E navigation timing issue, not touched by this change (parser-ast + data-owner query only). Re-running the required matrix now to get a green run.

peternhale (approved):
Approved. The latest changes require an explicit complete full-detail parse, safely re-establish completeness for full/unknown tables, and cover the UpdateSymbolSubset protocol path. The prior blocker is resolved.

PR #635, opened 2026-08-12
refactor(apex-lsp-extension): adopt client SDK - W-23163206

peternhale:
### What does this PR do?

- Rewire the VS Code extension onto `ApexClientCore` and `LanguageClientConnection`.
- Preserve raw `vscode-languageclient` lifecycle, tracing, telemetry, and web/desktop hover supersession middleware.
- Migrate custom `apex/*` operations to typed SDK methods and scope workspace-load resources to the active client lifecycle.
- Add regression coverage for startup, restart, teardown, graph navigation, and typed protocol behavior.

@W-23163206@

### Functionality Before

The extension maintained duplicate web and desktop `ClientInterface` wrappers with hand-rolled custom protocol wiring.

### Functionality After

The extension uses the standalone client SDK while retaining VS Code-host-specific behavior at the raw language-client layer.


kylewalke (commented):
## Code review — adopt client SDK (W-23163206)

Reviewed the full diff (30 files, +1654/−1651) plus the surrounding source at head `24d0f69`. This is a clean, well-executed refactor that rewires the VS Code extension onto `ApexClientCore` + `LanguageClientConnection` while preserving the VS Code-host-specific behavior (tracing, telemetry, hover supersession, workspace-load middleware) at the raw language-client layer.

### Overview

- Collapses the two hand-rolled web/desktop `ClientInterface` wrappers and the custom `apex/*` protocol wiring into typed SDK calls (`client.graphData(...)`, `profilingStart/Stop/Status`, `queueState`, `onQueueStateChanged`, `sendWorkspaceBatch`, `workspaceLoadComplete/Failed`, `onRequestWorkspaceLoad`, etc.). I verified every migrated method exists on `ApexClientCore`.
- `language-server.ts` shrinks ~1381→438 lines and is now organized into small, individually-testable functions (`createClientState`, `completeClientStart`, `cleanupClientState`, staged teardown).
- Hover supersession middleware is extracted to a shared `createHoverMiddleware()` used by both hosts (removes duplication that previously lived inline in `server-config.ts`).
- Graph data now flows through the typed `GraphDataShape` contract with explicit `convertProtocolGraphData` mapping instead of `response.data || response` + `?? []` fallbacks.

### Strengths

- **Robust lifecycle handling.** `cleanupClientState` runs every teardown stage even when one throws, captures the *first* error, logs the rest, and re-throws — and `cleanupAfterStartupFailure` preserves the original startup error over cleanup errors. The lifecycle queue (`enqueueLifecycleOperation`) serializes start/stop/restart so overlapping commands can't race. Good instinct.
- **Scope correctness.** `makeWorkspaceLoadScope = Scope.make()` is a lazy `Effect`, so each `Effect.runPromise` yields a fresh `CloseableScope` per client — the scope is tied to the client lifetime and closed during teardown. The `Scope.addFinalizer(resetWorkspaceState)` wiring means the module-level workspace-state Refs are reset when a client is torn down, so a restart re-loads correctly.
- **New cancellation support.** `loadWorkspaceForServer` now threads an `AbortSignal` with `throwIfAborted()` checkpoints between discovery/compression/send stages, so an in-flight workspace load is actually interrupted on teardown (previously it ran to completion).
- **Graph mapping is contract-accurate.** The hardcoded `protocolReferenceType` wire values (1,4,5,6,7,8,9,10) all match the real `ReferenceType` enum in `ApexSymbolRefManager.ts`, and `startLine` is 1-based → navigation correctly does `line - 1` for the 0-based `vscode.Position`. `graphNodeNavigationUri` correctly distinguishes URI schemes from Windows drive paths.
- **Strong test coverage on the hard parts.** `language-server.test.ts` covers the staged-cleanup matrix (multiple stages failing, error precedence, ApexLib partial-resource cleanup), and `workspace-load-handler.test.ts` covers client-scoped interruption, reload-after-teardown, and late-settling completion after disposal. Hover supersession has a race test.

### Observations (all minor)

1. **Desktop hover goes through two middleware hooks; web through one.** In `server-config.ts` the desktop `middleware` spreads `createHoverMiddleware()` (the `provideHover` supersession hook) *and* adds a `sendRequest` timing logger that also special-cases `textDocument/hover`. These compose correctly (feature-layer vs transport-layer), so it's not a bug — but web (`createWebClientOptions`) only gets `provideHover`, so the per-request timing logs exist on desktop only. If the timing log is meant to be a permanent diagnostic, consider sharing it across both hosts; if it was temporary debugging, consider dropping it. Either way, worth a conscious call.

2. **`_restartHandler` is now unused** in `createAndStartClient` / `restart*`. I confirmed the pre-PR baseline never invoked it either (crash-restart is driven by the LSP `errorHandler` → `CloseAction.Restart` in `server-config.ts`, which is preserved), so this is not a regression — but the parameter is now dead weight threaded through several signatures. Consider dropping it in a follow-up to simplify the API, or add a short comment noting restart is handled by the client's `errorHandler`.

3. **`handleLoadWorkspace` is marked deprecated** ("kept for backward compatibility during migration") but still exported and tested. Fine for this PR; just flagging it as a cleanup candidate once the migration settles so it doesn't linger.

4. **`documentSelector?: any[]`** persists in the workspace-loader signatures. Not introduced by this PR's core intent, but since the file was already touched, tightening to the shared selector type would remove one of the few remaining `any`s in otherwise well-typed code.

### Risk assessment

Low. The behavioral surface (start/stop/restart, workspace load lifecycle, hover supersession, crash-restart via `errorHandler`, graph navigation) is preserved and the tricky parts are backed by regression tests. The main residual risk is the usual "typed SDK contract vs. what the server actually returns" seam — but `GraphDataResult.data` is a required field in the shared contract and the graph mapper reads only guaranteed fields (`node.location.symbolRange.startLine`, etc.), so a shape mismatch would surface as a type error at build time rather than a runtime `undefined`.

**LGTM** pending your call on observation #1 (desktop-only hover timing logs). Nothing blocking.

kylewalke (dismissed):

kylewalke (changes requested):
## Review — adopt client SDK (`ApexClientCore` / `LanguageClientConnection`)

Nice refactor overall — collapsing the two hand-rolled `ClientInterface` wrappers into typed SDK calls removes a lot of duplicated request/notify plumbing, and the new lifecycle/cleanup is genuinely more disciplined: `cleanupClientState` attempts every stage and rethrows the first error without masking, `cleanupAfterStartupFailure` preserves the primary startup error, the Effect `Scope` gives real workspace-load cancellation via `AbortSignal`, and `VSCodeEditorContextAdapter` disposables are now torn down on every client teardown instead of leaking through `context.subscriptions` on restart. Test coverage for the new lifecycle paths is excellent. Findings were cross-checked by three passes and verified against the real SDK, the pre-PR code, and runtime constants.

Requesting changes on one behavioral regression (#1); the rest are recommend-fix.

---

### HIGH — behavioral regression

**1. ApexLib initialization failure is now fatal to the entire language server.**
`language-server.ts` `completeClientStart` awaits `operations.initializeApexLib?.()` inside the outer `try`, and `initializeApexLib` rethrows after disposing its own resources. That rethrow hits the `catch` → `cleanupAfterStartupFailure` (disposes core → stops the raw client, closes the scope) → rethrows, so `createAndStartClient` sets `clientState = undefined`, flips the status bar to Error, and the server never starts.

The old `createWebLanguageClient`/`createDesktopLanguageClient` both wrapped ApexLib init in a `try/catch` that only logged *"⚠️ Failed to initialize ApexLib … Standard library navigation may not work."* and continued. So a throw from `registerTextDocumentContentProvider`, `createFileSystemWatcher`, or the dynamic `import('@salesforce/apex-lsp-compliant-services')` used to degrade stdlib navigation only — now it takes down the whole LS. For a refactor meant to preserve behavior, this is a real regression. Recommend restoring graceful-degrade: treat ApexLib init failure as non-fatal (log + continue), disposing only its partial resources.

---

### MEDIUM

**2. `resetServerStartRetries()` is dropped on the default startup path.**
In `completeClientStart`, `markReady()` (which calls `resetServerStartRetries()` + `updateApexServerStatusReady()`) runs only in the `else` branch — i.e. when `shouldLoadWorkspace === false`. But `apex.loadWorkspace.enabled` **defaults to `true`** (package.json), so on the default path `markReady` is skipped and `resetServerStartRetries()` never runs (readiness instead arrives via `onWorkspaceIngestionComplete`, which only calls `updateApexServerStatusReady()`). The old code called `resetServerStartRetries()` **unconditionally** right after client creation, before workspace load.

Consequence: `handleAutoRestart` increments `serverStartRetries` on each failed start and gives up at `MAX_RETRIES`; without the post-success reset, transient failures accumulate across a session and can wedge auto-restart until a manual restart (which resets separately). Fix: call `resetServerStartRetries()` unconditionally on successful start, independent of the workspace-load branch.

**3. Mode-aware client capabilities may no longer be sent; `createInitializeParams` is now dead code.**
`createInitializeParams` (~60 lines, still `export`ed) has zero callers in `src/` or `test/` after this PR — the raw `LanguageClient` now owns the initialize handshake from `initializationOptions` + `documentSelector`. That means `getClientCapabilitiesForMode(serverMode)`, which the old path fed into `InitializeParams.capabilities`, is no longer wired in. Please confirm the server doesn't rely on those mode-specific capabilities (if it does, this is a silent functional regression), and delete `createInitializeParams` if it's genuinely unused.

**4. Dead E2E export `closeWelcomeTabs` (and pre-existing `assertWelcomeTabExists`).**
The welcome-suppression rewrite (settings-based `workbench.startupEditor:'none'`) removes every call site of `closeWelcomeTabs` (`commands.ts`, `settings.ts`, `vscode-interaction.ts`), but the function remains defined in `shared/utils/helpers.ts` and re-exported via the barrel — now dead. Remove both it and the barrel entries.

**5. GitHub sign-in / toast dismissal coverage reduced.**
The removed `dismissStartupPrompts` explicitly clicked "Skip"/"Continue without Signing In" on the VS Code account sign-in webview and dismissed notification toasts. The two replacement settings suppress the Welcome editor and walkthrough auto-open, but neither governs the sign-in page or toasts. If those reappear in the pinned CI image, the desktop path now has no fallback dismissal. Worth confirming against a real CI run before merge.

---

### LOW

**6. `convertProtocolGraphData` drops the defensive fallback.** `showGraph.ts` now does `const responseData = response.data; convertProtocolGraphData(responseData)` then `responseData.nodes.map(...)`, vs the old `response.data || response` + `nodes ?? []`. The server always populates `.data` with `nodes`/`edges`, so it's safe in practice, but a malformed/empty response now throws a `TypeError` instead of degrading. Note the `mapReferenceType`/`mapSymbolKind` numeric mappings and the new `graphNodeNavigationUri` (correctly handling `file://`/`apexlib://` URIs and Windows drive paths the old `Uri.file` mishandled) are verified correct — that part is an improvement.

**7. `resetWorkspaceState` finalizers accumulate on the per-client scope.** Each accepted `startWorkspaceLoad` appends a `Scope.addFinalizer(scope, ...)` to the single long-lived `workspaceLoadScope`; they only run at teardown, so they pile up over a long session (bounded, cleared on teardown). Compounding: `sharedWorkspaceLoadLayer` is re-provided per `runWorkspaceLoad` call, so `WorkspaceState` Refs are rebuilt fresh each time and the `hasLoaded`/`isLoading` guards never observe state across separate loads. The per-call `Effect.provide` is **pre-existing**, not introduced here — flagging because the new finalizer-per-load rides on top of it.

---

### Investigated and cleared (not issues)
- **"Stuck on Starting forever" — refuted.** The `loadWorkspaceForServer` catch path calls `updateApexServerStatusReady()` before rethrowing, `startWorkspaceLoad` forks the loader (so `completeClientStart` doesn't hang on it), and `startIngestionTimeout()` self-heals to ready after 5 min if the ingestion-complete notification never arrives. Readiness is not permanently lost.
- **`makeWorkspaceLoadScope`** is a reusable factory Effect, not a shared mutable scope — each client gets its own `CloseableScope`. AbortSignal propagation on scope close, cleanup ordering (scope before `core.dispose()`), and no-double-stop (`core.dispose()` → connection → `client.stop()`, never a separate `rawClient.stop()`) are all correct.
- **hoverMiddleware** extraction preserves the old inline behavior exactly (the dropped `(error) => { throw error }` was a redundant re-throw; rejection still propagates). Supersession Map lifecycle unchanged.
- **`registerConfigurationChangeListener`** now returns a Disposable that IS retained as `state.configurationListener` and disposed in `cleanupClientState`. Not leaked.
- **Dependency wiring** (`@salesforce/apex-lsp-client: "*"` + wireit `../apex-lsp-client:compile` + tsconfig project ref) is consistent; the one-line lockfile change is sufficient (the link node pre-existed via `apex-lsp-testbed`); Playwright container tag `v1.62.1` matches `@playwright/test@1.62.1`; the `.cls` deletion switch to `withFileTypes`+`isFile()` is strictly safer; `getModifierShortcut('Shift+P')` correctly resolves to Cmd/Ctrl+Shift+P.

Net: strong, well-tested refactor with better teardown discipline. #1 is the one I'd block on (graceful-degrade → fatal is a user-facing behavior change); #2 and #3 are cheap and worth folding in since they hit the default path.

kylewalke (approved):
## Re-review — all findings resolved ✅

Re-reviewed the `address SDK review findings` commit plus the two E2E follow-ups, verified against the branch code.

**Blocking finding fixed:**
- **ApexLib init no longer fatal** — `completeClientStart` now wraps `initializeApexLib()` in a try/catch that logs *"Standard library navigation may not work"* and continues, restoring the old graceful-degrade behavior. The added test asserting the core is NOT disposed on ApexLib failure is the right guard.

**MED/LOW all addressed:**
- `resetServerStartRetries` split into its own `resetStartRetries` operation called unconditionally before the load branch (no longer dropped on the default `loadWorkspace.enabled=true` path); tests assert the ordering.
- `createInitializeParams` + the unused `getClientCapabilitiesForMode`/`InitializeParams` imports removed.
- Welcome/sign-in suppression: `startVSCodeWeb` now combines settings suppression + `dismissStartupPrompts` (Skip/Continue-without-signing-in) + `closeWelcomeTabs` — the restored helper is genuinely used again, closing the gap I raised.
- `convertProtocolGraphData` fallback (`?? []`) restored; finalizer accumulation guarded via the `scopesWithWorkspaceReset` WeakSet.

Minor: `assertWelcomeTabExists` is restored but still uncalled — it's a debug-only helper that was already orphaned on `main`, so not a regression; drop it whenever convenient.

Solid refactor with good teardown discipline and strong lifecycle test coverage. LGTM.

PR #647, opened 2026-08-13
feat(apex-lsp-parser-ast): cross-file renameField occurrence collection - W-23631084

kylewalke:
## What does this PR do?

Extends `textDocument/rename` to **fields and properties** (renameField occurrence collection + WorkspaceEdit). WI 4.1 of the rename epic.

A field cursor falls through renameLocal's `null` path to a new `resolveFieldRename` on the request-pool worker, which runs the workspace-wide two-phase scan (coordinator lexical prefilter → per-file standalone parse) and **disambiguates each candidate occurrence by its receiver's declared type**, so `acct.total` is renamed while an unrelated `Other.total` is not.

## How it works

- **New `findFieldOccurrences` op** (apex-parser-ast): resolves each candidate's receiver via the co-located `VARIABLE_USAGE` reference — a form-independent signal (assignment LHS emits no chained ref) — and its `resolvedSymbolId` → declared type name. This mirrors jorje's declaring-type-qualified key, done at scan time (we have no persistent index). Receivers that only resolve cross-file are conservatively **skipped, never mis-renamed**.
- Adds the field declaration explicitly (it is not a reference token).
- Reuses renameLocal's validation (`validateRenameName`) and the WorkspaceEdit shape.

## Correctness guards (from adversarial review)

Three corruption/under-rename cases surfaced in review and fixed with regression tests:
- **Local variable shadowing** the field name → skipped (would otherwise corrupt unrelated locals).
- **Explicit `this.`/`super.`** qualifiers → receiver maps to the enclosing type.
- **Inner-class same-named fields** → `getEnclosingType` picks the innermost type.

## Scope / known limitations (deferred, documented in the plan)

Verdict-only occurrence collection. NOT here: hierarchy conflict detection (4.2, wires 4.0's `CheckMemberConflicts`) or the renameMethod edit-target fan-out (5.2). Deferred, safe (skip not corrupt): simple-name vs namespace-qualified type match, static `ClassName.FIELD` access, and the qualify-on-conflict use-site rewrite.

## Stacking

Stacked on **4.0** (`CheckMemberConflicts`, PR #634) — this PR's base is the 4.0 branch, so the diff shows only 4.1's changes. Retarget to `main` once #634 merges.

## Tests

- `findFieldOccurrences.test.ts` (10 cases: cross-file disambiguation, `Other.total` not renamed, implicit-this, the 3 review regressions).
- `RenameThroughWorkerTopology.node.test.ts` extended (11 pass) — cross-file field rename through the live worker pool.

@W-23631084@


peternhale (changes requested):
## Blocking findings\n\n1. **P1: Renaming an inherited field leaves  references broken.**\n\n[](https://github.com/forcedotcom/apex-language-support/blob/feature/W-23631084-renamefield-occurrence-collection/packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts#L216-L229) maps both  and  receivers to the enclosing type. For a field declared in  and  in , renaming  compares  to , skips the occurrence, then still returns an edit that renames the declaration. Applying the edit leaves  unresolved. Resolve  to the actual superclass, or conservatively decline the rename when that relation cannot be proved. Please add an end-to-end regression.\n\n2. **P1: Unqualified inherited field references in subclasses are skipped, producing uncompilable edits.**\n\n[](https://github.com/forcedotcom/apex-language-support/blob/feature/W-23631084-renamefield-occurrence-collection/packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts#L131-L145) says implicit receivers are accepted inside the declaring type or a subclass, but the implementation only accepts exact type-name equality. A bare  in  legitimately resolves to ; it is omitted while the declaration is renamed, leaving the child broken. Use parser-owned inheritance information to establish the relationship, or decline the rename rather than emitting a partial edit. Please add a regression for bare inherited field reads/writes.

peternhale (changes requested):
## Correction: Blocking findings

The previous review was submitted with shell-stripped Markdown due to command quoting. This review supersedes its body; its **Changes requested** state remains intentional.

1. **P1: Renaming an inherited field leaves `super.field` references broken.**

[`findFieldOccurrences.ts:216-229`](https://github.com/forcedotcom/apex-language-support/blob/feature/W-23631084-renamefield-occurrence-collection/packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts#L216-L229) maps both `this` and `super` receivers to the enclosing type. For a field declared in `Account` and `super.total` in `Child extends Account`, renaming `Account.total` compares `Child` to `Account`, skips the occurrence, then still returns an edit that renames the declaration. Applying the edit leaves `super.total` unresolved. Resolve `super` to the actual superclass, or conservatively decline the rename when that relation cannot be proved. Please add an end-to-end regression.

2. **P1: Unqualified inherited field references in subclasses are skipped, producing uncompilable edits.**

[`findFieldOccurrences.ts:131-145`](https://github.com/forcedotcom/apex-language-support/blob/feature/W-23631084-renamefield-occurrence-collection/packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts#L131-L145) says implicit receivers are accepted inside the declaring type or a subclass, but the implementation only accepts exact type-name equality. A bare `total` in `Child extends Account` legitimately resolves to `Account.total`; it is omitted while the declaration is renamed, leaving the child broken. Use parser-owned inheritance information to establish the relationship, or decline the rename rather than emitting a partial edit. Please add a regression for bare inherited field reads/writes.

kylewalke:
Addressed both P1 findings in the latest commit.

**Root of both:** `findFieldOccurrences` silently *skipped* a same-named field access it couldn't attribute (a `super.field`, or a bare inherited field in a subclass), while `resolveFieldRename` still renamed the declaration — a broken partial edit. Fixed by making the op distinguish *proven-unrelated* from *unprovable*, and having the caller **decline** rather than emit a partial edit for the unprovable case.

**1. `super.field` (finding #1):** `super` no longer maps to the enclosing type. `super` refers to an ancestor that a single-file standalone parse can't resolve, so it now yields `unsafe` → the whole rename is declined (`RenameErrorResult`, "Cannot safely rename"). Added an end-to-end topology regression: renaming a base field referenced by a subclass via `super.total` declines instead of emitting a partial edit.

**2. Bare inherited field in a subclass (finding #2):** a bare `total` in a type that is not the declaring type is now classified by whether that type *could* inherit the field — i.e. whether it declares a superclass (`enclosingTypeCouldInherit`, reading the parser-owned `TypeSymbol.superClass`). If it extends something → `unsafe` (possible inherited reference) → decline. If it extends nothing → genuinely unrelated → safely skipped (this keeps cross-class disambiguation working: `Other.total` with no superclass stays distinct from `Account.total`). Added unit regressions for bare inherited read/write, plus the unrelated-vs-subclass distinction.

Establishing/refuting the actual inheritance chain needs the cross-file graph the pool path lacks, so declining is the safe alternative per your guidance. Verified: unit 13/13, renameField topology 12/12, worker build + lint clean.

peternhale (changes requested):
Request changes. This review covers the incremental diff above stacked parent #634; the #634 findings remain applicable separately.

1. **Critical** `packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts:116`
An explicit receiver of a subtype is treated as a proven mismatch. For `Child c; c.total`, where `Child extends Base` and the target is `Base.total`, the code skips the reference because `Child !== Base`, but still renames `Base.total`. This emits a broken partial rename. Only skip a different receiver type after parser-owned hierarchy resolution proves it is outside the target subtype cone; otherwise classify it as unsafe and decline. Add direct and transitive subclass receiver tests.

2. **Critical** `packages/apex-ls/src/worker.platform.shared.ts:3885` and `packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts:116`
The declaring type is carried as `containingType.name`, despite being described as an FQN. Simple names are not unique for nested or namespaced types: renaming `OuterOne.Inner.total` can match `OuterTwo.Inner.total`. Carry a parser-owned stable type identity or canonical FQN through cursor resolution and occurrence classification; add same-named nested-type coverage.

3. **Critical** `packages/apex-ls/src/worker.platform.shared.ts:4015`
Phase-two candidate parse/scan errors are logged and ignored, then a partial `WorkspaceEdit` may still be returned. A failed or incomplete candidate could contain a target reference, leaving it dangling after the declaration changes. Treat each candidate failure/incomplete semantic result as unsafe and decline the entire rename.

4. **High** `packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts:221`
Static accesses such as `Account.total` have a type receiver (`CLASS_REFERENCE`), but lookup accepts only `VARIABLE_USAGE`. They are skipped and left unchanged while the declaration renames. Support parser-owned type-qualified receivers and add a cross-file static-field rename test.

5. **High** `packages/apex-ls/src/worker.platform.shared.ts:4001`
The live buffer replaces only an existing data-owner candidate; a cursor document newly added or modified in an unsaved buffer may not be scanned at all. Ensure the cursor URI is always scanned from `req.content`; consider versioned edits or decline when other candidate snapshots are stale.

6. **Medium** `packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts:92`
Per candidate, every occurrence scans all references and may scan all symbols, creating quadratic work. Build position-indexed reference and symbol-ID maps once per parsed file, and provide cooperative cancellation between candidate files.

kylewalke:
Thanks @peternhale — addressed all six findings in `cc3b9818`. Two of them are correctness-limited by an architectural constraint I want to be explicit about (details below).

**#1 (Critical) — subtype receiver treated as proven mismatch.** A qualified receiver whose resolved type is a **locally-visible subtype** of the declaring type (declares a `superClass` we can't refute standalone) is now classified `unsafe` → the whole rename declines, instead of `skipped` (which emitted the broken partial edit). `Child c; c.total` with `Child extends Base` → unsafe; transitive subclass → unsafe; `Other` with no superclass → still a safe skip.
  *Honest limitation:* the request-pool worker parses each candidate **standalone** and cannot resolve inheritance across files (its graph is a transient per-request subset — confirmed by recon). So a subtype is only *detectable* when its declaration is visible in the candidate file (inner class / same file). A cross-file subclass can't be caught by this op alone — that's the same cross-file-graph limitation already accepted for this WI. A precise fix needs a new data-owner hierarchy query; I can file a follow-up if you'd prefer that over the conservative decline.

**#2 (Critical) — declaring type carried as short name.** Now carried as a canonical FQN end-to-end: `getDeclaringTypeForCursor` returns `containingType.fqn` (fallback `constructFQN`), and `findFieldOccurrences` compares **FQN-to-FQN**. One wrinkle I had to reconcile: the graph `.fqn` carries a block-scope name duplication for nested types (`outerone.outerone.inner`) that the standalone-computed FQN (`outerone.inner`) lacks; `normalizeFqn` collapses consecutive duplicate segments so they compare equal while still distinguishing `outerone.inner` from `outertwo.inner`. Tests cover both.

**#3 (Critical) — phase-2 scan failure logged and ignored.** A candidate parse/scan failure now pushes to the `unsafe` set, firing the existing decline-whole-rename path instead of returning a partial edit.
  *Note:* no integration test — `CompilerService.compile` collects diagnostics rather than throwing on malformed Apex, so the catch block isn't reproducible through the live topology without fault injection. The fix reuses the already-tested decline path.

**#4 (High) — static `Account.total` skipped.** A probe showed the parser actually emits a co-located `VARIABLE_USAGE` named `Account` at the field token (not only a differently-positioned `CLASS_REFERENCE`). `resolveReceiverType` now treats a receiver that resolves to no variable-like symbol as a static type qualifier and disambiguates it by type. Tests: static match (2 occurrences) + static mismatch (skip).

**#5 (High) — cursor URI may be absent from candidates.** The cursor file is now always scanned with the live buffer even when the prefilter omits it (Stage 6's `seen` set prevents double-scan). Integration test with an unopened cursor buffer.

**#6 (Medium) — quadratic scan.** Per-file position-indexed receiver-ref map and type-symbol list are built once at the top of `findFieldOccurrences` instead of rescanning per candidate. Cooperative cancellation: there's no `CancellationToken` in this request path (documented inline) — adding one would be upstream infra, not invented here.

Full pre-commit suite green (5360 tests); findFieldOccurrences 20/20; RenameThroughWorkerTopology 14/14.

peternhale (changes requested):
Request changes. Two critical partial-rename paths remain at the current head.

1. P1 - A qualified receiver whose subtype relationship is cross-file is still treated as proven unrelated.

In packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts around lines 204-234, a mismatched receiver is unsafe only when findLocalType finds its declaration in the standalone candidate file and that local TypeSymbol declares a superclass. A normal caller file containing Child c; c.total but not the Child declaration makes findLocalType return null, so the occurrence is added to skipped as receiver-type-mismatch:Child.

I reproduced this directly against the PR head while renaming Base.total: occurrences=[], skipped=[receiver-type-mismatch:Child], unsafe=[]. If Child extends Base in Child.cls, resolveFieldRename can therefore rename Base.total while leaving c.total dangling. The new tests declare Child locally and do not cover the cross-file topology described by the original finding. This cannot be deferred while the operation still returns a WorkspaceEdit. Use parser-owned data-owner hierarchy information to prove the receiver is outside the subtype cone, or classify every unprovable non-local receiver as unsafe and decline.

2. P1 - CompilerService diagnostics do not trigger the candidate-scan failure path.

In packages/apex-ls/src/worker.platform.shared.ts around lines 4055-4095, only a thrown exception is caught and added to unsafeOccurrences. CompilerService.compile normally returns a recovered SymbolTable plus an errors array for malformed or incomplete Apex; the code does not inspect that array. I reproduced a candidate with four parser errors that lexically contains total but produced zero occurrences and zero unsafe notes. The declaration would still be appended and a partial edit could be returned.

Please treat any candidate compilation with errors or otherwise incomplete semantic output as unsafe and decline the whole rename, with a regression exercising the normal diagnostic-return path rather than only the exception catch.

The FQN anchoring, static access handling, live cursor-buffer inclusion, position indexes, and thrown-exception decline otherwise look directionally correct.

kylewalke:
Both P1s addressed in `4ecf19c2`.

**P1 #1 — cross-file subtype receiver still a proven skip.** You're right; my earlier fix only caught locally-declared subtypes. Now the rule is strict: a mismatched qualified receiver is a proven `skipped` **only** when its type is declared in the same candidate file **and** declares no superclass (so it provably can't be in the target's subtype cone). Every other case — a non-local receiver (unknown hierarchy, e.g. your `Child c; c.total` with `Child` in Child.cls) or a local type with a superclass — is now `unsafe` → the whole rename declines. This trades some cross-hierarchy renames (which now report "Cannot safely rename") for never emitting a dangling edit; restoring precision would need a data-owner hierarchy query, which I've left out of scope here.
  Regressions: unit tests for the cross-file subclass receiver and non-local static access (both → `unsafe`), plus an **integration** test proving `resolveFieldRename` *declines* when a caller file holds an unprovable non-local receiver (i.e. it returns an error, not a WorkspaceEdit).

**P1 #2 — diagnostic-return path not caught.** Correct — `CompilerService.compile` recovers and returns an `errors` array rather than throwing, so the exception-only catch missed it. Stage 6 now inspects the result and declines when a candidate has **syntax** errors. I deliberately count *syntax* errors only, not the whole `errors` array: a standalone single-file parse can't resolve cross-file references, so `errors` routinely contains benign *semantic* (unresolved-reference) entries for otherwise-valid files — declining on those would reject nearly every real multi-file rename (it did, in the nested-FQN test, until I narrowed it). Syntax errors mean the parse tree itself is broken and occurrences may be missing.
  Regression: an integration test with a candidate that has 4 real parser errors → the rename declines via the diagnostic path (not the exception catch).

findFieldOccurrences 21/21; RenameThroughWorkerTopology 15/15; full pre-commit suite green.

peternhale (changes requested):
Request changes. Two P1 partial-rename paths remain at the current head.

1. **P1 - Chained member receivers are classified using the root receiver, so a real target reference can be safely skipped.**

In `packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts` around lines 171-238, `findColocatedReceiverRef` returns the first co-located `VARIABLE_USAGE`. For `a.inner.total`, that is `a`, not the immediate receiver `inner`. I reproduced this with a local `Container` containing `Account inner`, then `Container a; a.inner.total = 5;` while renaming `Account.total`. The result was `occurrences=[]`, `unsafe=[]`, and `skipped=[receiver-type-mismatch:Container]`. Because `Container` is local and has no superclass, the code treats the candidate as proven unrelated; `resolveFieldRename` can then append and rename the declaration while leaving `a.inner.total` dangling.

Please resolve the parser-owned immediate member-access receiver and its declared type. If that cannot be established, chained accesses must be `unsafe` so the rename declines. Add a regression where the leaf receiver is the declaring type.

2. **P1 - `findLocalType` proves hierarchy mismatch by ambiguous leaf name.**

Around lines 447-455, `findLocalType` strips a qualified type to its leaf and returns the first same-named local type. I reproduced a file with `OuterOne.Child` (no superclass) declared before `OuterTwo.Child extends Base`, followed by `OuterTwo.Child c; c.total = 5;` while renaming `Base.total`. The lookup selected `OuterOne.Child`, classified `c.total` as `receiver-type-mismatch:OuterTwo.Child`, and returned no unsafe occurrence. This again permits a broken partial edit.

Use the parser-owned canonical FQN for the local-type lookup, or treat ambiguous/unprovable matches as unsafe. Please add repeated-inner-name coverage for receiver hierarchy classification.

The non-local receiver and syntax-diagnostic fixes from the previous review otherwise work, and the current CI matrix is green.

kylewalke:
Both remaining P1s fixed in `c51bc16f`. I probed the parser to confirm the receiver structure before fixing.

**P1 #1 — chained receiver classified by the root.** Confirmed: for `a.inner.total` the parser emits co-located receiver `VARIABLE_USAGE`s for **both** the root `a` and the immediate receiver `inner` at the field token, and `findColocatedReceiverRef` returned the first (`a`). Probe results: simple `acct.total`, static `Account.total`, and `this.total` each yield **exactly one** co-located receiver; a chain yields **two+**. So `findColocatedReceiverRefs` now returns all distinct co-located receivers, and **≥2 ⇒ chained ⇒ `unsafe` → decline** (we can't reliably resolve the immediate member-access receiver's type standalone). Regressions: `Container a; a.inner.total` where `inner` is the declaring type → unsafe; and the leaf-receiver-is-declaring-type variant → unsafe.

**P1 #2 — `findLocalType` ambiguous leaf.** Fixed: it now matches by **canonical FQN** (`computeTypeFqn`), and falls back to a leaf match **only when it is unique** in the file; an ambiguous or unmatched receiver → `null` → `unsafe`. Regression: `OuterOne.Child` (no superclass) declared before `OuterTwo.Child extends Base`, with a `OuterTwo.Child c; c.total` receiver renaming `Base.total` → now `unsafe` (previously matched `OuterOne.Child` and was wrongly skipped).

findFieldOccurrences 24/24; RenameThroughWorkerTopology 15/15; full pre-commit suite green. CI matrix was green on the prior head and these are parser-ast-only changes.

peternhale (changes requested):
Request changes.

1. **P1 - Name-based receiver deduplication still lets chained target references be safely skipped.**

In `packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts` around lines 352-373, `findColocatedReceiverRefs` deduplicates co-located receivers by lowercased name and removes every receiver whose name equals the target field. The caller identifies a chain only when at least two entries survive. Both transformations can collapse a genuine chain to one apparent receiver, re-entering the unsafe root-receiver classification that this commit intends to close.

I reproduced two cases against head `5a6452f7` while renaming `Account.total`:

- `Container` has `Account a`, and `Container a; a.a.total = 5;`. Both receiver nodes are named `a`, so name dedup leaves one. The result was `occurrences=[]`, `unsafe=[]`, `skipped=[receiver-type-mismatch:Container]`.
- `Container` has `Account total`, and `Container a; a.total.total = 5;`. The immediate receiver is named like the target and is filtered out. The result contained two `receiver-type-mismatch:Container` skips and no unsafe occurrence.

`resolveFieldRename` can therefore append and rename the `Account.total` declaration while leaving these real references dangling. Please detect chains using parser-owned chain/member-access identity rather than distinct identifier text. If the immediate receiver cannot be established, preserve the ambiguity as `unsafe` and decline. Add both repeated-name and target-named-immediate-receiver regressions.

The previous chained-receiver and repeated-inner-FQN reproductions are otherwise fixed; the existing focused suite passes 24/24 and CI is green.

kylewalke:
Addressed the 2026-08-21 P1 (name-based receiver dedup collapsing chains) in `30b365a5`.

**Root cause:** chain detection was text-based — `findColocatedReceiverRefs` deduped co-located receivers by lowercased name and dropped any entry named like the target field. Both transforms collapse a genuine chain to one apparent receiver, re-entering the root-receiver classification: `a.a.total` (root+immediate both `a`) deduped to one `a`; `a.total.total` (immediate `Container.total` named like the field) had its immediate receiver dropped.

**Fix (parser-owned structural identity, not identifier text):**
- Chain detection now counts the **raw** co-located `VARIABLE_USAGE` entries the parser emits — one per chain element — with no name-dedup and no field-name exclusion. I verified across every access form that a simple/static/`this`/`super` access emits exactly one and a chain emits ≥2 (a 4-deep chain emits 4).
- The receiver lookup is gated on **whether a `FIELD_ACCESS` ref exists at the candidate's position**, not on the candidate's post-dedup context. A bare implicit-this / local `total` is a lone `VARIABLE_USAGE` (no `FIELD_ACCESS`), so its own entry no longer leaks into the receiver bucket — which is what the removed field-name exclusion had been (incorrectly) guarding. Crucially, at a chained final field (`a.total.total`) a same-position `VARIABLE_USAGE` and `FIELD_ACCESS` co-exist and position-dedup can surface either as the candidate; the `FIELD_ACCESS`-at-position signal is robust to that, the context gate was not.

Net: `a.a.total` and `a.total.total` now classify `unsafe` (`chained-receiver-unprovable`) → decline, instead of a silent `skipped`.

**Regressions added:** repeated-name chain (`a.a.total`) and target-named immediate receiver (`a.total.total`), both asserting `unsafe` (not `skipped`). findFieldOccurrences 26/26; RenameThroughWorkerTopology 15/15; full pre-commit matrix green (0 failed).

kylewalke:
Follow-up: a fresh from-scratch review surfaced a further **P1** in the chain-detection I added — fixed in `8b99ca7c`.

**The hole:** my raw co-located-receiver count only reveals a chain for an **assignment-LHS write** (`a.inner.total = 5` stamps both `a` and `inner` at the field token). I verified against the real parser that every **non-write** form of `a.inner.total` stamps only the ROOT as a co-located receiver:
- read `x = a.inner.total`, argument `f(a.inner.total)`, `return a.inner.total`, `if (a.inner.total > 0)` → `coVU=[a]` (count 1)
- `this.inner.total` → `coVU=[]` (count 0)

So the `≥2` guard was dead for reads, and the leaf was mis-classified by its root: dropped as a mismatch (dangling partial edit) or — when the root type happens to equal the declaring type — renamed as an unrelated token (corruption).

**Fix:** detect chains via the parser's form-independent structural identity — the whole-expression `chainNodes` array. A multi-hop access has `chainNodes.length ≥ 3` (root + ≥1 intermediate + leaf); a simple qualified access is length 2; bare is none. I index the leaf position of every length-≥3 chain and decline (`unsafe`) any candidate at such a position, OR'd with the existing `coVU ≥ 2` write-form signal (the write form emits no whole-expression chain ref, so both signals are needed). Verified this catches read/arg/return/if/ternary/method-in-chain/deep/`this.`-rooted forms.

**Regressions added:** read RHS, method argument, binary expr, ternary, `this.inner.total`, and a root-is-declaring-type corruption case — all asserting `unsafe` (never a silent skip or occurrence). findFieldOccurrences 32/32; RenameThroughWorkerTopology 15/15; full matrix green (0 failed).

peternhale (changes requested):
Request changes.

1. **P1 - An unqualified method-result receiver is still treated as implicit `this`, allowing both dropped and corrupt edits.**

In `packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts` around lines 182-240, the new structural guard recognizes a multi-hop chain only when `chainNodes.length >= 3` (or the write-form co-located receiver count is at least two). An access such as `getAccount().total` is still qualified, but its parser structure is a two-node method-result-plus-field access and it has no co-located `VARIABLE_USAGE` receiver. It therefore falls through to the bare/implicit-`this` branch.

I reproduced both destructive directions against head `fac496c1` with focused parser tests:

- In an unrelated `Caller`, `getAccount().total` while renaming `Account.total` produced `occurrences=[]`, `unsafe=[]`, and an `implicit-this-unrelated-type:Caller` skip. The declaration can be renamed while this real reference is left dangling.
- Inside `Account`, `getOther().total` was returned as an occurrence of `Account.total`, so the rename can rewrite an unrelated `Other.total` token.

Please use the parser-owned immediate receiver for qualified field accesses, including method/constructor results. If its type cannot be established, classify the candidate as `unsafe` and decline the rename. Add both regressions above and an end-to-end test proving `resolveFieldRename` never returns a partial or corrupt `WorkspaceEdit` for this receiver shape.

The repeated-name, target-named intermediate, read-form, and `this`-rooted multi-hop cases from the previous review are otherwise fixed, and the current CI matrix is green.

kylewalke:
Addressed the 2026-08-24 P1 (method-result receiver treated as implicit `this`) in `ebafe8e0`.

**Root cause (verified against the parser):** a method/constructor-result or cast receiver is a *qualified* access, but in the READ form the parser emits ONLY a whole-expression `chainNodes` ref (length 2) — no top-level `FIELD_ACCESS` at the leaf and no co-located `VARIABLE_USAGE` receiver. So `getAccount().total` fell through to the bare/implicit-`this` branch: dropped as `implicit-this-unrelated-type` in an unrelated class (dangling), or returned as an occurrence inside `Account` (corrupts an unrelated `Other.total`).

**Fix:** I index the leaf position of every chain (`chainNodes.length >= 2`), and distinguish it from a top-level `FIELD_ACCESS`. A candidate that is a chain leaf but has **no top-level `FIELD_ACCESS` and no co-located variable receiver** is a qualified access whose immediate receiver is a non-variable expression (method/constructor result, cast) — its declared type can't be established from a standalone parse, so it's classified `unsafe` (`non-variable-receiver-unprovable`) → decline. This is precise: a simple `a.total` read DOES emit a top-level `FIELD_ACCESS` (→ Case 1, resolve receiver), `this.total` read stays on its existing path, and genuine bare implicit-`this` has no chain leaf (→ Case 2). I confirmed the exact emission for method-result/ctor/cast reads vs simple/static/`this`/bare via focused parser probes before touching the logic.

**Regressions added:**
- unit: `getAccount().total` in an unrelated `Caller` → `unsafe` (not `skipped`); `getOther().total` inside `Account` → `unsafe` (not an occurrence); `((Account)o).total` cast → `unsafe`.
- end-to-end (RenameThroughWorkerTopology): renaming `AccountMr.total` while `CallerMr` has `getAccount().total` → `resolveFieldRename` returns an error (`Cannot safely rename … non-variable-receiver-unprovable`), never a partial/corrupt `WorkspaceEdit`.

findFieldOccurrences 35/35; RenameThroughWorkerTopology 16/16; full pre-commit matrix green (0 failed).

peternhale (changes requested):
Request changes.

1. **P1 - Constructor-result field receivers are still absent from semantic occurrence collection, allowing a declaration-only partial rename.**

In `packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts` around lines 178-205, the new `chainLeafPos` guard correctly catches method-result and cast receivers when the parser emits a whole-expression `chainNodes` reference. However, the code comments also claim constructor results such as `new Account().total` are covered, while the full parser currently emits only the constructor/type references for that expression — it emits no `total` reference and no chain leaf.

I reproduced this against head `a26232c1` with syntactically clean sources, both with `Account` cross-file and with `Account` declared as a local nested type. For `Integer x = new Account().total;`, `findFieldOccurrences` returned:

```
occurrences=[]
skipped=[]
unsafe=[]
```

Because the candidate scan reports no unsafe occurrence, `resolveFieldRename` proceeds past the decline guard, appends the field declaration, and can return a declaration-only `WorkspaceEdit` while leaving `new Account().total` dangling.

Please fix the parser-owned reference collection so the constructor-result field leaf is represented and classified, or otherwise conservatively decline when this semantic shape cannot be established. Do not infer the missing reference from raw source text. Add a focused `findFieldOccurrences` regression and an end-to-end topology regression proving the rename returns an error rather than a partial edit.

The method-result read/write, cast receiver, repeated-name, target-named intermediate, and multi-hop access fixes otherwise work as intended.

kylewalke:
Addressed the 2026-08-24 P1 (constructor-result field receiver absent from occurrence collection) in `d04e5e8c`.

**Root cause (verified against the parser):** for `new Account().total` the reference collector emitted only the `Account` CONSTRUCTOR_CALL — **no `total` reference, no chain leaf, no memberAccess context**. In `ApexReferenceCollectorListener`, `enterAnyId` pushes the `.total` leaf into the active chain scope, but `extractBaseExpressionFromParser` returns `''` for a `NewExpression` base, so `finalizeChainScope` bails on the empty base and **discards the leaf**. With no candidate, `findFieldOccurrences` returned `occurrences=[] skipped=[] unsafe=[]` and `resolveFieldRename` proceeded to a declaration-only edit — dangling `new Account().total`.

**Fix (parser-owned, per your preferred option — no raw-text inference):** `extractBaseExpressionFromParser` now resolves a `NewExpression` left-expression to its constructed type's leaf name (stripping generics/qualifier). The chain then finalizes and the parser emits the field leaf as a multi-hop `chainNodes` (length ≥ 3) reference. My existing multi-hop guard classifies that leaf `unsafe` (`chained-receiver-unprovable`), so the whole rename declines. No occurrence-collection change was needed beyond the parser now representing the leaf.

Verified emission before/after and confirmed no over-reach: simple `a.total`, static, `this.`, and implicit-this all still classify correctly. **Full parser-ast suite: 3252 passed, 0 failed** (the base-extraction change touches a shared listener, so I ran the whole suite).

**Regressions added** (both requested shapes + more):
- unit (findFieldOccurrences): `new Account().total` read → `unsafe`; `new Account().total = 5` write → `unsafe`; `new Other().total` inside `Account` → `unsafe` (no over-rename); constructor-result on a **local nested** declaring type → `unsafe`.
- end-to-end (RenameThroughWorkerTopology): renaming `AccountCr.total` while `CallerCr` has `new AccountCr().total` → `resolveFieldRename` returns an error (`Cannot safely rename … chained-receiver-unprovable:multi-hop-member-access`), never a partial/corrupt `WorkspaceEdit`.

findFieldOccurrences 39/39; RenameThroughWorkerTopology 17/17; full pre-commit matrix green (0 failed).

peternhale (changes requested):
Request changes.

1. **P1 - Constructor-result handling fabricates invalid semantic references.**

In `packages/apex-parser-ast/src/parser/listeners/ApexReferenceCollectorListener.ts` around lines 2629-2644, returning the constructed type name as `baseExpression` makes the normal identifier-chain path treat that name as if it began at the `DotExpression` start.

I reproduced this against head `d04e5e8c` with `Integer x = new Account().total;`. The collected semantic references include:

- a malformed `Account.Account.total` chain;
- a synthetic `Account` `CHAIN_STEP` and standalone `CLASS_REFERENCE` at columns 20-27, covering `new Acc`;
- the real `Account` constructor reference at columns 24-31.

The previous declaration-only rename path is closed because `total` now appears as an unsafe chain leaf, but the shared semantic snapshot is corrupted: position-based consumers can see `Account` while the cursor is on `new`, and hierarchical resolution receives a duplicated constructor hop. The implementation also derives a semantic type name by applying `/\<.*\>$/s` to `createdName().getText()`, contrary to the repository's parser-owned-semantics rule prohibiting regex/text heuristics for semantic facts.

Please represent the `NewExpression` receiver structurally using the existing constructor node and its real parser-owned identifier range, rather than routing its type name through the ordinary lexical identifier-base path. Add a collector/reference regression that asserts the complete chain and ranges (including that no identifier reference covers `new`), not only a `findFieldOccurrences` classification test.

The rename-specific focused suite otherwise passes 39/39, and the prior constructor-result partial-rename finding is fixed.

kylewalke:
Addressed the 2026-08-24 P1 (constructor-result handling fabricated invalid semantic references) in `a68d843c`.

**Reverted** the rejected approach entirely — no more returning a type-name string through the lexical identifier-base path, and no `/<.*>$/` regex.

**Structural fix** (in `ApexReferenceCollectorListener`): the `NewExpression` receiver is now represented from the constructor's **real parser-owned identifier range**, mirroring the method-result chain shape:
- `createNewChainScope` detects a constructor receiver (`unwrapToNewExpression`, which also unwraps `SubExpression` parentheses) and builds the receiver hop via `buildConstructorReceiverNode` — using the grammar-owned `creator → createdName → idCreatedNamePair.anyId` + its `SymbolLocation` (no text/regex heuristics; returns null for collection creators).
- `enterAnyId` now skips an `anyId` whose parent is `IdCreatedNamePairContext` (the constructor **type** identifier), which was the source of the doubled `Account.Account` hop.
- `finalizeChainScope` emits the chain via `createChainedExpressionReference` (location computed purely from node ranges), only when the structural receiver exists — it no longer routes through `createChainRootReference`.

**Verified emission** for `new Account().total` (read + write + parenthesized `(new Account()).total`):
- chain `Account.total` = `[Account CONSTRUCTOR_CALL @ the `Account` token, total FIELD_ACCESS @ its true range]` — a clean 2-node chain;
- **no reference covers `new`** (asserted); **no doubled hop**; **no standalone top-level `total` FIELD_ACCESS**; the pre-existing single `CONSTRUCTOR_CALL` for `new Account()` is preserved.

Because the emission is now a clean 2-node non-variable-receiver chain (identical shape to method-result/cast), `findFieldOccurrences` classifies it `non-variable-receiver-unprovable` — the **same** decline path as those siblings (I updated my 4 constructor unit-test assertions to that reason for consistency; no special-case code in the op).

**New collector/reference regression** (`test/references/constructorResultFieldAccess.test.ts`, as requested): asserts the complete chain + node ranges/contexts, that **no ref covers `new`**, no doubled hop, and no standalone top-level `total` — for cross-file `Account`, local-nested `Account`, and the parenthesized form.

**Same-variety sweep** (per request): probed other expression-result receivers — `arr[0].total`, `getList()[0].total`, `getAccount().total`, `((Account)o).total` are already represented and decline correctly (unchanged). I additionally fixed the **parenthesized constructor** `(new Account()).total`, which had the same dropped-leaf gap.

**One residual, disclosed:** a collection-creator literal that is immediately indexed and field-accessed — `new List<Account>{ ... }[0].total` — still drops the leaf (the receiver is an indexed collection literal, a distinct shape from a bare/paren constructor result; `buildConstructorReceiverNode` deliberately returns null for `List/Set/Map` creators). It's an extremely rare construct with no test here; I left it rather than expand this shared-listener change into indexed-creator-literal territory. Happy to file a follow-up if you'd like it closed.

findFieldOccurrences 39/39; new collector suite 3/3; constructor topology decline test green; **full apex-parser-ast suite 3255 passed / 0 failed** (shared-listener change, so I ran the whole suite); full pre-commit matrix green (0 failed).

peternhale (changes requested):
Request changes on the latest head (`f8e8ca7`).

1. **P1: same-line nested types are attributed to the outer type, corrupting unrelated field references.**

`packages/apex-parser-ast/src/symbols/ops/findFieldOccurrences.ts:654-675` chooses the “innermost” containing type using only `endLine - startLine`. When an outer and nested class are formatted on one line, both spans are `0`, so the first symbol (the outer type) remains selected.

I reproduced this with:

```apex
public class Outer { public Integer total; public class Inner { public Integer total; public void f() { total = 5; } } }
```

Calling `findFieldOccurrences(..., declaringTypeFqn: 'Outer')` incorrectly returns the bare `total` inside `Inner.f()` as an `Outer.total` occurrence (expected 0, received 1). Renaming `Outer.total` therefore edits an unrelated `Inner.total` use. Compare complete source ranges, including columns, or use parser-owned nesting/parentage to select the deepest containing type. Please add this single-line regression.

2. **Blocking design violation: renameField makes semantic completeness depend on a raw-source regex prefilter.**

`packages/apex-ls/src/worker.platform.shared.ts:4030-4036` obtains the only phase-two candidate set from `dataOwner:FindOccurrenceCandidates`; that handler filters stored documents through `textMentionsSymbol`, implemented as a regular expression over raw Apex text. Any false negative removes a file before the parser can inspect it, so the returned `WorkspaceEdit` can silently omit a real reference.

This conflicts with the repository agreement that semantic language features must rely on lexer/parser output, symbol tables, resolved references/xref data, or stable semantic snapshots—not raw-source regex or text heuristics. Please drive candidate discovery from parser-owned semantic state (or conservatively parse the relevant document set) and preserve uncertainty when that state is unavailable.

Validation performed against this head:

- Changed parser suites passed: 42/42.
- Focused same-line nested-type regression failed as described above.
- Current GitHub checks are green, but do not cover this case.

peternhale (changes requested):
Request changes on the latest head (`8374698`). The prior same-line nested-type and raw-regex findings are fixed, but one correctness blocker remains.

1. **P1: renameField can emit a partial WorkspaceEdit while the workspace document set is still loading.**

`packages/apex-ls/src/worker.platform.shared.ts:5389-5422` returns every document currently present in storage for `skipTextFilter`, but it does not check `svc.symbolManager.isWorkspaceLoadSessionActive()`. During an active workspace-load session, “all stored documents” is only a partial workspace. `resolveFieldRename` nevertheless treats that result as complete and can rename the declaration before files containing references have reached the store.

I reproduced this through the worker topology:

- Begin a workspace-load session and leave it active.
- Keep a caller containing `LoadingBase b; b.total = 2;` not yet loaded/stored.
- Invoke rename on `LoadingBase.total` using the live cursor buffer.
- The conflict query correctly rejects because the hierarchy graph is incomplete, but the local fallback proceeds for the plain final class.
- `FindOccurrenceCandidates` returns the currently empty stored set; the live cursor file is inserted and the operation returns a `WorkspaceEdit` changing only `LoadingBase`.
- When the caller subsequently loads, `b.total` remains stale.

The focused regression expected a `RenameErrorResult`, but received:

```text
{"changes":{"file:///test/LoadingBase.cls":[...declaration and local usage edits...]}}
```

This is the same partial-rename safety boundary as candidate discovery: when workspace completeness is unavailable, preserve uncertainty rather than treating the current store contents as the full reference set. For `skipTextFilter`, reject `FindOccurrenceCandidates` while a workspace-load session is active (as `CheckMemberConflicts` already does), or return an explicit completeness state and make `resolveFieldRename` decline. Please add the active-load topology regression.

Validation on this head:

- Original same-line nested-type probe now passes.
- Parser/lexer suites passed: 51/51.
- Existing conflict/rename topology suites passed: 42/42.
- Repository compile and GitHub checks are green.
- Focused active-load completeness regression fails as described above.

kylewalke:
Thanks — both addressed on the latest head, plus a further P1 I found and fixed via a from-scratch review pass before replying.

### 1. Active-load partial-rename (your P1) — fixed
`FindOccurrenceCandidates` now fails closed for `skipTextFilter` when `svc.symbolManager.isWorkspaceLoadSessionActive()` (same boundary `CheckMemberConflicts` already enforces). During an active load session the store is only a partial workspace, so returning it would let renameField edit a declaration while not-yet-loaded references dangle. `resolveFieldRename` already declines on a query error, so the rename now returns a `RenameErrorResult` instead of a partial edit. find-references (no `skipTextFilter`) is best-effort discovery and is intentionally unaffected. Added the active-load topology regression (plain final class → declines while a session is active).

### 2. Static-qualified field access `Type.field` — separate P1, now fixed
A clean re-review of the whole PR surfaced a corruption/partial-edit bug the prior rounds missed. The receiver index only collected `VARIABLE_USAGE`; when a static class qualifier **resolves within the standalone-parsed file** the parser emits it as a `CLASS_REFERENCE` co-located at the field-access position, which was never indexed — so the access fell through to the bare implicit-`this` case and was classified by the *enclosing* type. Two verified manifestations:

- **Silent drop → dangling edit:** an inner class referencing the outer class's static field (`Account.total` inside `Account.Helper`) was `skipped`, so renaming `Account.total` edited the declaration and left that reference un-renamed.
- **Wrong-token rename → corruption:** an unrelated sibling static access (`Root.Account.total`) was renamed while renaming `Root.Foo.total`.

Cross-file/unknown qualifiers are emitted as `VARIABLE_USAGE` and were unaffected, which is what masked this — it bit the idiomatic inner-class/outer-static pattern.

Fix: index co-located `CLASS_REFERENCE` static qualifiers (excluding `this`/`super`) and add classification **Case 1c** (reached only with no variable receiver, after chain detection). It resolves the qualifier to its type FQN via the same `computeTypeFqn`/`normalizeFqn` machinery the receiver path uses, then: qualifier == declaring type → occurrence; resolves in-file to a *different* type → skip (a static access binds the named type exactly, so it provably targets that other type — safe to skip, no decline); unresolvable → unsafe (decline). Chained static (`Outer.Inner.total`) still hits the chain-unsafe path first. Added `findFieldOccurrences` unit tests (inner→outer, unrelated-sibling, same-type self, chained-static guard) and a worker-topology test proving the inner-class outer-static pattern renames end-to-end with no partial edit.

### 3. Context on the earlier round
Candidate discovery no longer uses the raw-text prefilter (`skipTextFilter` returns the full stored set, parsed per-candidate). The one consequence — an unparseable file forcing a decline — is narrowed by a **lexer** check (`lexMentionsIdentifier`, not a regex): a broken file whose token stream lacks the field identifier provably can't reference it and is skipped rather than blocking the rename.

### Open question (unverified — flagging for your input)
The review also raised a lower-confidence concern I could not reproduce from a standalone probe: `getDeclaringTypeForCursor` returns the graph's `containingType.fqn`, while `findFieldOccurrences` compares against a standalone-parse FQN normalized only for the block-artifact case. If the graph `.fqn` carries a namespace prefix the standalone parse doesn't (e.g. `ns.account` vs `account`), receiver compares could mismatch. Do you know whether `.fqn` is namespace-qualified in a namespaced project? If so I'll add namespace normalization to the compare; if not, it's a non-issue.

Validation: full apex-parser-ast suite green (3267); RenameThroughWorkerTopology 20/20; find-references topology unaffected; full pre-commit matrix green.

peternhale (approved):
Approved — the active workspace-load partial-rename finding is addressed. FindOccurrenceCandidates now fails closed for rename while the stored document set is incomplete, and resolveFieldRename declines instead of returning a partial WorkspaceEdit.

Nit (non-blocking): please strengthen the active-load topology regression to assert that the returned error message identifies the active/incomplete workspace-load state. The current assertion accepts any error, so an unrelated earlier failure could make the test pass without exercising the intended completeness guard.

PR #655, opened 2026-08-18
feat(apex-lsp-parser-ast): replace stub generator - W-23491682

peternhale:
### What does this PR do?

Replaces the documentation/XML-based Apex stub generator with the Salesforce Symbol Table API generator in `packages/apex-parser-ast`.

- Fetches API type definitions in one streaming request and generates the configured `StandardApexLibrary` namespaces.
- Preserves hand-authored builtins and validates captures before removing stale generated stubs.
- Adds `npm run update:stubs` and focused generator coverage.
- Removes the retired generator package and its GitHub Actions workflow.

@W-23491682@

### Functionality Before

Stub generation scraped documentation and optionally augmented it with a local, uncommitted `bytecode.xml` file.

### Functionality After

Stub generation fetches the Symbol Table API and regenerates stubs through `npm run update:stubs`.

### Verification

- `npm run typecheck`
- `npm test -- --runInBand apexStubGenerator api-stub-workflow apiStubScripts`
- Commit hooks: full lint and test suite, 5,371 passing tests


kylewalke (changes requested):
## Review — Replace doc-scraper stub generator with Symbol Table API generator

Thanks for this — the deletion side is clean, the new scripts are well-structured, and the injection/path-traversal hardening (`APEX_IDENTIFIER`, `SAFE_PATH_COMPONENT`, annotation escaping) plus the destructive-op ordering (validate + full load *before* any delete) are genuinely well done. Findings below were cross-checked by three independent passes and, where noted, verified against the real Apex parser and the existing committed stubs.

Requesting changes on two HIGH correctness bugs that are pervasive across the 53 regenerated namespaces, plus a documentation/architecture mismatch.

---

### HIGH — must fix

**1. Enum constants are never emitted.**
`apexStubGenerator.js` `generateTypeBody` selects `keyword = 'enum'` but the body only emits fields/properties/methods/innerTypes — there is no enum-constant handling (the old package had `generateEnumContent`/`ENUM_EXTRA_VALUES`). Every regenerated enum comes out empty, e.g. `global enum LoggingLevel { }`. Impacted target-namespace enums include `System/LoggingLevel`, `Schema/DisplayType`, `System/Quiddity`, `DataSource/DataType`, and dozens more — all of `LoggingLevel.DEBUG`-style resolution breaks. Existing committed stubs (e.g. `FormulaEval/FormulaReturnType.cls`) carry their values, confirming the format regression. If the API instead models constants as static fields, `generateField` emits `global static ... DEBUG;` inside an `enum` body, which is a hard syntax error (verified against the parser: `mismatched input 'global'`).

**2. Interface methods are emitted with visibility modifiers → invalid Apex.**
In `generateMethod`, the interface branch emits `${modStr}${returnType}${methodName}(...);` with `modStr` still containing `public`/`global`. Verified against the parser: `global interface X { public void m(); }` produces the semantic error *"Modifiers are not allowed on interface methods"*, while the modifier-free form compiles clean. Every existing hand-authored interface stub (`Database/Batchable.cls`, `Database/BatchableContext.cls`, `CommerceTax/TaxEngineAdapter.cls`) deliberately omits method modifiers. Since a symbol-table API almost certainly reports interface methods as `public`/`global`, this fires for essentially every interface method in every regenerated namespace.

> Note: `apexStubGenerator.test.ts` and `api-stub-workflow.test.ts` currently **assert** `'public void interfaceMethod();'` as correct and never actually compile it — so the test suite codifies bug #2. Please add a real `CompilerService.compile(...).errors` assertion for a representative enum and interface so this can't regress silently.

---

### MEDIUM

**3. The "two-tier" TypeRegistry design is documented but not implemented.**
`fetch-api-stubs.mjs` (design-decision comment) and `SKILL.md` claim non-`TARGET_NAMESPACES` types "remain available to the type registry" via the fetched JSON. The unmodified `generate-stdlib-cache.mjs` reads only `src/resources/StandardApexLibrary` and `src/resources/builtins` — it has zero references to `build/api-stubs`, `*.json`, or `typeStubs` (grep-confirmed). The per-namespace JSON written for the other ~83 namespaces into gitignored `build/api-stubs/` is consumed by nothing. Effective coverage is the same 53 namespaces that produce `.cls` files. Please drop the claim or file a follow-up to actually wire it in.

**4. Coverage claim overstates what ships.** `API_STUBS_README.md` cites "~8,345 types / 136 namespaces (4x increase)," but only the 53 `TARGET_NAMESPACES` produce `.cls`/registry entries (per #3). The 136 is total *fetched*, most discarded. Please reframe to the 53 actually bundled.

**5. `category` vocabulary is self-contradictory.** `fetch-api-stubs.mjs` uses `BUILTIN | DATABASE | DYNAMIC` (default `BUILTIN`, which is what the code sends); `API_STUBS_README.md` documents `CLASS | INTERFACE | ENUM | TRIGGER`; `example-api-call.sh` defaults `CLASS` and hardcodes `API_VERSION="v67.0"`. A user copying the README/example sends a category the script never uses. Make the example call the real script and align the docs to the shipping default.

**6. Two divergent `BUILTIN_CLASSES` skip-lists, one manually kept in sync.** `generate-api-stubs.mjs` skips 14 System classes (incl. List/Map/Set/System/Url/String/Exception); the unmodified `generate-stdlib-cache.mjs` has its own list of 7. `SKILL.md` calls only the former "authoritative" and the README's builtin list omits the 7 extras. Please note the second list and the lock-step requirement (or dedupe to a shared source).

**7. No CI replacement for the deleted `generate-stubs.yml`.** Refresh is now fully manual and org-gated (264+ org + authenticated `sf` CLI), which can't run in CI as written. Deliberate, but worth calling out explicitly — "Automated CI refresh" is currently only a "Future Enhancement."

---

### LOW / robustness (data-dependent — reachable only if API shape shifts)

- **`shouldSkipFile` casing mismatch:** System builtins compared case-insensitively, namespaced builtins (`DMLOptions.cls`, `DescribeSObjectResult.cls`) via exact `Map.get(filename)`. If the API emits different casing for a namespaced builtin, the hand-crafted override is bypassed. Use the same comparison for both.
- **Trigger filename trap:** `loadNamespaceStubs` strips only `.cls` before the `SAFE_PATH_COMPONENT` check, so a generated `Name.trigger` throws `Invalid generated filename` and aborts the namespace. Gated today by `category=BUILTIN` (no triggers), but latent since `generateApexStubs` fully supports triggers.
- **Non-void generic return with missing `returnType`:** default-return uses `method.returnType.name` not the demangled generic type, so a mangled name with no `returnType` field can produce `List<String> getList() { }` (non-void, no `return`) — uncompilable.
- **`formatType` on a type object with no `.name`** emits the literal `undefined`; **`demangleMethodName`** can truncate a real method name containing `_r<Capital>`; **`<clinit>`** static initializers pass through as invalid identifiers. All low-likelihood given current API data — worth a guard or a documented assumption.
- **`.gitignore` adds `/packages/apex-stubs-generator/`** for a directory this PR deletes — confirm intent.
- **Work-item numbers:** README lists `W-23631682` (Work Item) and `W-23491682` (Upstream Fix), but the branch/PR and all code comments cite `W-23491682`. Confirm the driving WI.

---

### Verified clean
`tsconfig.base.json` / `knip.json` / `package-lock.json` deletions are complete; the root `packages/*` workspace glob auto-drops the removed package; wireit correctly re-runs `generate:stdlib-cache` when `StandardApexLibrary/**/*.cls` change, so `update:stubs → compile` regenerates the protobuf cache; `test-compile-stubs.js` referenced in the README does exist; `calculateCaptureChecksum` is order-independent and content-sensitive; `cleanNamespaceDirectory` preserves builtins and non-target namespaces.

Net: strong scaffolding and safety, but #1 and #2 produce data loss / non-compiling Apex across the regenerated stdlib and the tests currently lock in #2 — those plus the #3 doc mismatch should land before merge.

kylewalke (approved):
## Re-review — all findings resolved ✅

Thanks for the thorough response. Re-reviewed the `address API stub review findings` commit and verified each item against the branch code (I ran the updated generator through the parser to confirm output, not just the commit message).

**Both HIGH findings fixed and verified:**
- **Enum constants** — `generateTypeBody` now emits them as declarations from `values || enumValues || enumConstants || fields` and skips field/method emission for enums. Confirmed output: `global enum FieldEnum { X, Y }` / `global enum ValEnum { A, B }` (valid enum syntax, not static fields).
- **Interface method modifiers** — visibility stripped for `INTERFACE`. Confirmed: `global interface Iface { void run(); }`. The new compile-assertion test in `api-stub-workflow.test.ts` is exactly what was needed.

**MED/LOW all addressed:** two-tier claim + coverage + category vocab corrected in docs, `example-api-call.sh` now calls the real script, builtin-list alignment documented, case-insensitive builtin + namespace matching, `.(cls|trigger)$` filename strip, `genericReturnType || formatType(...)` for default returns, `!typeRef.name` guard, `_r` demangle now requires `$$l`, `<clinit>` filtered, single work-item ref. Each has a focused test.

Nice hardening beyond the review too (annotation-`parameters`-as-array guard, tightened `TARGET_NAMESPACES`). LGTM.

PR #665, opened 2026-08-24
feat(apex-lsp-compliant-services): renameField hierarchy conflict detection - W-23631086

kylewalke:
### What does this PR do?

Wires the data-owner `CheckMemberConflicts` query (WI 4.0) into `renameField` as **Stage 4.5** of `resolveFieldRename` — after newName validation, before the occurrence scan. For a non-no-op rename it asks the complete workspace graph whether the new name collides in the same type, a non-private ancestor member, or a descendant member (the latter only when the renamed field is not private), replicating jorje's `FieldRenameHandler` rules. A genuine hierarchy **declaration** collision returns a `RenameErrorResult` (code `-32600` → LSP `ResponseError`) naming the colliding type; use-site lexical shadowing is 4.1's rewrite, not a conflict.

The renamed field's declaring type and effective visibility are resolved in one pass by `resolveFieldContextForCursor` (a refactor of `getDeclaringTypeForCursor`): it returns the declaring-type **FQN** (not the short name, so nested `Outer.Inner` types disambiguate) plus whether the field is effectively private (Apex `Default` treated as private).

**Review hardening folded in before opening:**

- **`currentName` wiring** — Stage 4.5 sends the member's current name so the data-owner can authoritatively short-circuit a no-op/case-only rename and locate the member being renamed.
- **Visibility self-verify** — the 4.0 query no longer trusts the wire `isRenamedMemberPrivate`. It reads the renamed member's actual visibility from its own full+complete table (via an extracted `getTypeMembers` helper shared with `hasMemberNamed`), treats `Default` as private, and **fails closed** on any disagreement, so the descendant-conflict gate runs off the verified bit. Absent `currentName` it falls back to the wire value (backward-compatible).
- **Best-effort pre-check** — conflict detection never blocks the core rename on an inability to check. The query fails closed against false negatives, so a query *error* (e.g. the declaring type lives in an unopened/unsaved buffer absent from the data-owner store) is logged and the rename proceeds — the occurrence scan still guards correctness. Only a definitive `conflict:true` blocks.

All jorje rule fidelity (private filters, descendant gating, case-insensitivity) lives in the already-unit-tested 4.0 query.

### Stacked PR

Targets **#647** (WI 4.1, cross-file occurrence collection), not `main`. Review/merge #647 first; this branch will retarget `main` once #647 lands. Builds on the merged `CheckMemberConflicts` query from #634 (WI 4.0).

### Tests

- 2 worker-topology tests: rejects a same-type collision, allows a non-colliding rename.
- 5 direct-query tests for the visibility self-verify: descendant check applies when the verified member is non-private, skips when verified private, fails closed when the caller under- or over-reports visibility, and fails closed when the renamed member is not found.
- Full local pre-commit matrix green (all packages, 0 failed).

@W-23631086@


peternhale (changes requested):
Request changes on the latest head (`38fa977`). This review covers #665's incremental diff above stacked parent #647; #647's blockers remain applicable separately.

1. **P1: a failed conflict query is converted into a fail-open rename that can create duplicate declarations.**

`packages/apex-ls/src/worker.platform.shared.ts:4058-4105` documents `CheckMemberConflicts` as fail-closed, but catches any rejection, sets `conflict = null`, and proceeds. The later occurrence scan only classifies references; it does not detect same-type, ancestor, or descendant declaration-name collisions.

For example, if a live/unsaved declaring type contains both `total` and `amount` but is unavailable or incomplete in the data-owner store, renaming `total` to `amount` makes `CheckMemberConflicts` reject. This catch then proceeds, the occurrence scan can succeed on the live buffer, and the returned edit creates duplicate `amount` declarations. The same failure mode bypasses ancestor/descendant collision protection.

Please preserve the query's uncertainty by declining/retrying the rename, using a prior stable semantic snapshot, or running an equivalent parser-owned check against the live complete state. Do not proceed merely because the occurrence scan succeeds. Add a topology regression that forces `CheckMemberConflicts` to reject while the live source contains a same-type collision, and assert that no `WorkspaceEdit` is returned.

Validation performed against this head:

- Repository compile passed.
- `CheckMemberConflicts.node.test.ts`: 21/21 passed.
- `RenameThroughWorkerTopology.node.test.ts`: 19/19 passed.
- Current GitHub checks are green, but there is no query-error/fail-open regression.

peternhale (dismissed):
Approved on the latest head (`ab1c4a4`) for #665's incremental diff above stacked parent #647.

The previously reported P1 is resolved: a failed `CheckMemberConflicts` query no longer proceeds blindly. The new parser-owned local fallback checks a full, known-complete declaring-type table, rejects same-type collisions, and declines unverifiable hierarchy state. The unstored same-type collision regression covers the original corruption path.

Validation:

- Repository compile passed.
- `CheckMemberConflicts.node.test.ts` and `RenameThroughWorkerTopology.node.test.ts`: 42/42 passed together.
- Current GitHub checks are green.

#647 still has a separate active-workspace-load candidate-completeness blocker; this approval covers #665's own incremental changes and does not override the stacked parent's review state.

peternhale (dismissed):
Approved on the current rebased head. The incremental contents across all four touched files are identical to the previously approved version, and the prior fail-open conflict-query P1 remains resolved. PR #647 is now the correct stacked base, and the current CI/E2E checks are green.

PR #667, opened 2026-08-27
feat(apex-lsp-extension): renameField prepareRename + e2e coverage - W-23631087

kylewalke:
### What does this PR do?

Delivers **4.3 renameField e2e coverage**, plus the two fixes that coverage surfaced.

**1. prepareRename now recognizes fields/properties (not just locals).** The server advertises `renameProvider: { prepareProvider: true }`, so VS Code fires `textDocument/prepareRename` FIRST on F2 and won't open the rename box unless it returns a range containing the cursor. `DispatchPrepareRename` only resolved locals, so F2 on a field showed **"The element can't be renamed"** — renameField was unreachable through the editor even though `resolveFieldRename` fully supports it. The new field branch resolves via the **same `svc` cursor primitives `resolveFieldRename` uses**, so prepareRename accepts exactly the field cursors the rename can service (no accept-then-decline drift), returning the identifier range containing the cursor — declaration or usage token.

**2. Declaration edit no longer corrupts the type token.** The e2e caught a bug the unit/worker-topology tests missed (they assert edit *count* + *newText*, never *ranges*): under full-workspace ingestion the graph-stored field symbol's `identifierRange` can span the field's **type** token, so `declarationLocationForCursor` returned that and the rename edited `Integer` instead of the field name. Usage edits were correct because they come from `findFieldOccurrences`' standalone parse. Fix: recompute the declaration token range from a **standalone parse** of the declaring file (`fieldDeclarationRangeFromParse`) — the same source as the usage ranges — so declaration and usages stay consistent.

**3. e2e coverage** (`apex-rename-symbol.spec.ts`, reusing the 3.3 `ApexEditorPage.rename()` helper), both passing on web:
- Single-file field rename — declaration + `this`-qualified + implicit-`this` usages.
- Cross-file field rename — declaration in `RenameFieldModel` + field-access occurrences in `RenameFieldClient`, proving the multi-file `WorkspaceEdit`. Gated on full workspace ingestion.
- New fixtures: `RenameFieldSample`, `RenameFieldModel`, `RenameFieldClient` (+ `-meta.xml`).
- Scoped `BasePage`'s editor locator to exclude the Monaco rename widget (`.monaco-editor.rename-box`), which lingers in the DOM after a rename and otherwise breaks `openFile` with a strict-mode violation when switching files post-rename (renameMethod/renameType e2e will need this too).

**4. Integration coverage:** field prepareRename worker-topology tests for a declaration cursor and a usage cursor.

**5. Review nit (W-23631084 final review):** the active-load rename regression asserted only that *some* error was returned; it now asserts the decline carries the active/incomplete workspace-load signal (`-32600` + a message matching `workspace load session still active`/`incomplete`).

Validation: full pre-commit test matrix green (5505 passed, 0 failed); `RenameThroughWorkerTopology` 25/25 (incl. the 2 field prepareRename tests + the strengthened active-load regression); both renameField e2e tests pass on web. The live e2e is exercised in CI.

@W-23631087@

### Functionality Before

F2 on an Apex **field** → "The element can't be renamed" (prepareRename returned null for non-locals). Field rename was only reachable programmatically, never through the editor.

### Functionality After

F2 on a field opens the rename box; confirming renames the declaration and all field occurrences — single-file and cross-file — as a `WorkspaceEdit`, with the declaration token renamed correctly.

### DIFF
--- a/e2e-tests/pages/BasePage.ts
+++ b/e2e-tests/pages/BasePage.ts
@@ -30,10 +30,9 @@ export class BasePage {
   constructor(page: Page) {
     this.page = page;
     this.workbench = page.locator(SELECTORS.WORKBENCH);
-    // Exclude the Monaco rename widget (`.monaco-editor.rename-box`), which stays
-    // in the DOM after the first rename. Without this, `.monaco-editor` matches
-    // two elements post-rename and `editor.waitFor()` (openFile) hits a
-    // strict-mode violation when switching files. It's never the editor we want.
+    // Exclude the Monaco rename widget (`.monaco-editor.rename-box`), which lingers
+    // in the DOM after a rename — otherwise `.monaco-editor` matches two elements
+    // and editor.waitFor() (openFile) throws a strict-mode violation on file switch.
     this.editor = page.locator(`${SELECTORS.MONACO_EDITOR}:not(.rename-box)`);
     this.explorer = page.locator(SELECTORS.EXPLORER);
     this.sidebar = page.locator(SELECTORS.SIDEBAR);
--- a/e2e-tests/pages/BasePage.ts
+++ b/e2e-tests/pages/BasePage.ts
@@ -30,12 +30,10 @@ export class BasePage {
   constructor(page: Page) {
     this.page = page;
     this.workbench = page.locator(SELECTORS.WORKBENCH);
-    // Exclude the Monaco rename widget: it carries the `.monaco-editor` class
-    // (`.monaco-editor.rename-box`) and Monaco keeps it in the DOM after the
-    // first rename, so an unscoped `.monaco-editor` locator would resolve to TWO
-    // elements once any rename has run — breaking `editor.waitFor()` (e.g. in
-    // openFile) with a strict-mode violation when switching files post-rename.
-    // The rename-box is never the editor we want, so excluding it is always safe.
+    // Exclude the Monaco rename widget (`.monaco-editor.rename-box`), which stays
+    // in the DOM after the first rename. Without this, `.monaco-editor` matches
+    // two elements post-rename and `editor.waitFor()` (openFile) hits a
+    // strict-mode violation when switching files. It's never the editor we want.
     this.editor = page.locator(`${SELECTORS.MONACO_EDITOR}:not(.rename-box)`);
     this.explorer = page.locator(SELECTORS.EXPLORER);
     this.sidebar = page.locator(SELECTORS.SIDEBAR);
--- a/e2e-tests/pages/BasePage.ts
+++ b/e2e-tests/pages/BasePage.ts
@@ -30,7 +30,13 @@ export class BasePage {
   constructor(page: Page) {
     this.page = page;
     this.workbench = page.locator(SELECTORS.WORKBENCH);
-    this.editor = page.locator(SELECTORS.MONACO_EDITOR);
+    // Exclude the Monaco rename widget: it carries the `.monaco-editor` class
+    // (`.monaco-editor.rename-box`) and Monaco keeps it in the DOM after the
+    // first rename, so an unscoped `.monaco-editor` locator would resolve to TWO
+    // elements once any rename has run — breaking `editor.waitFor()` (e.g. in
+    // openFile) with a strict-mode violation when switching files post-rename.
+    // The rename-box is never the editor we want, so excluding it is always safe.
+    this.editor = page.locator(`${SELECTORS.MONACO_EDITOR}:not(.rename-box)`);
     this.explorer = page.locator(SELECTORS.EXPLORER);
     this.sidebar = page.locator(SELECTORS.SIDEBAR);
     this.statusbar = page.locator(SELECTORS.STATUSBAR);
--- a/e2e-tests/test-data/apex-samples/force-app/main/default/classes/RenameFieldClient.cls
+++ b/e2e-tests/test-data/apex-samples/force-app/main/default/classes/RenameFieldClient.cls
@@ -1,6 +1,5 @@
-// Consumer for cross-file renameField e2e (W-23631087 / 4.3). Holds a
-// locally-typed RenameFieldModel instance and reads/writes its `quantity`
-// field, so a rename of RenameFieldModel.quantity must update these usages.
+// Consumer for the cross-file renameField e2e (W-23631087): reads/writes
+// RenameFieldModel.quantity through a local instance.
 public with sharing class RenameFieldClient {
     public Integer useModel() {
         RenameFieldModel model = new RenameFieldModel();
--- a/e2e-tests/test-data/apex-samples/force-app/main/default/classes/RenameFieldClient.cls
+++ b/e2e-tests/test-data/apex-samples/force-app/main/default/classes/RenameFieldClient.cls
@@ -1,0 +1,10 @@
+// Consumer for cross-file renameField e2e (W-23631087 / 4.3). Holds a
+// locally-typed RenameFieldModel instance and reads/writes its `quantity`
+// field, so a rename of RenameFieldModel.quantity must update these usages.
+public with sharing class RenameFieldClient {
+    public Integer useModel() {
+        RenameFieldModel model = new RenameFieldModel();
+        model.quantity = 5;
+        return model.quantity;
+    }
+}
--- a/e2e-tests/test-data/apex-samples/force-app/main/default/classes/RenameFieldClient.cls-meta.xml
+++ b/e2e-tests/test-data/apex-samples/force-app/main/default/classes/RenameFieldClient.cls-meta.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>62.0</apiVersion>
+    <status>Active</status>
+</ApexClass>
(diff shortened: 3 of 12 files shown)


peternhale (changes requested):
Reviewed head `8cde8fef6d08bda765585c1c0e3013feacde01af`.

Requesting changes for these correctness issues:

1. **P1 — Known-corrupt declaration ranges remain a fallback.**

   In `packages/apex-ls/src/worker.platform.shared.ts:4645-4673`, `declRange` starts with the graph range that this PR identifies as potentially pointing at the type token. If declaration content is unavailable, parsing throws, or the standalone lookup returns no match, the code silently applies that suspect range anyway. This can reproduce the `Integer` → new-name corruption.

   Please fail closed unless the standalone parser supplies a verified declaration range.

2. **P1 — prepareRename exposes non-user-owned fields and properties.**

   At `packages/apex-ls/src/worker.platform.shared.ts:3981-3984`, the new branch checks only the symbol kind. Standard-library or generated SObject members can therefore be advertised as renamable through F2, while the field rename path has no source-provenance guard and may produce edits involving virtual resource URIs or decline only after opening the UI.

   Please reject declarations outside user-owned Apex sources in both prepare and rename.

3. **P2 — Field prepareRename reintroduces the identifier-end off-by-one.**

   At `packages/apex-ls/src/worker.platform.shared.ts:3991-4008`, the field path uses `exactCursorReference` / `positionInRange`, whose end comparison is inclusive. Parser identifier ranges are half-open, and the existing local path explicitly rejects a cursor at `endColumn`. The field path can therefore accept a cursor immediately after the identifier.

   Please use half-open containment and add the equivalent field boundary regression test.

The targeted rename E2E passes. The PR's overall checks are currently red due to a Windows worker timeout and a hover E2E failure that appear unrelated, but should still be rerun before merge.

kylewalke:
Thanks for the review — all three addressed on the latest head (`60172346`).

**1. P1 — fail closed on an unverified declaration range.** `resolveFieldRename` no longer falls back to the graph `identifierRange` (the range that can span the type token). It recomputes the declaration range from a standalone parse and, if that can't be produced (content unavailable, parse error, or no unambiguous match), the rename now **declines** rather than emitting the suspect range — so the `Integer`→new-name corruption can't reproduce through the fallback.

**2. P1 — reject non-user-owned declarations.** Added a source-provenance guard (a positive allowlist of the workspace document schemes) applied in **both** paths: `prepareRename` returns null (F2 shows "can't be renamed") and `resolveFieldRename` declines, so standard-library and generated-SObject members are never offered for rename or edited via synthetic URIs. The guard keys off the *declaration's* source, so a stdlib/SObject field is rejected even when the cursor is on a usage in an editable file.

**3. P2 — half-open identifier containment.** The field `prepareRename` path now uses strict half-open containment (`endColumn > cursor`), matching the local path, so a cursor immediately after the identifier is rejected. Added the field boundary regression test.

Also found and fixed a few adjacent issues while I was in there (hardening the provenance guard and the declaration-range disambiguation to fail closed).¹

Validation: rename worker-topology suite 26/26, including the new boundary regression. Agreed the red checks look like an unrelated Windows worker timeout + hover E2E flake — I'll rerun them before merge.

¹ Details in the commit history (`47005a02`, `60172346`) if useful.

peternhale (changes requested):
Reviewed head `60172346ec485d3be024f15e5110ee4a47c98afc`.

Requesting changes for these remaining correctness issues:

1. **P1 — The stdlib provenance guard still passes `apexlib://` declarations.**

   `USER_OWNED_URI_SCHEMES` is initialized from `getAllImmutableSchemes()`, but that function explicitly includes `apexlib`. Consequently, `isUserOwnedApexUri` returns true for standard-library declarations, and both prepareRename and resolveFieldRename still admit the source class this guard is intended to reject.

   Please use a source-of-truth that excludes synthetic schemes and add a negative prepare/rename regression test for a standard-library member.

2. **P1 — A missing declaration still permits a usage-only WorkspaceEdit.**

   The new fail-closed handling runs only inside `if (declaration)`. When `declarationLocationForCursor` returns null, the code skips that entire block and can still return the occurrences gathered during Stage 6. Since those occurrences intentionally exclude declarations, this produces a partial rename that changes usages while leaving the declaration untouched.

   Please decline when the declaration itself cannot be resolved, as well as when its standalone range cannot be verified, and cover that path with a regression test.

3. **P2 — Declaring-type disambiguation cannot match nested fields.**

   `fieldDeclarationRangeFromParse` compares `getSymbolById(field.parentId).name` with the declaring type leaf. Fields and properties are parented to a generated class block (`class_N`), not directly to the class symbol, so when multiple nested types declare the same field name this comparison never matches and every rename is declined.

   Please walk through the class block to the owning type (preferably comparing the owning FQN) and add a nested-type same-name regression test.

The half-open cursor-boundary correction and removal of the suspect graph-range fallback are otherwise sound. Current CI and E2E checks are green, but they do not exercise the cases above.

kylewalke:
Thanks — all three addressed on the new head (`22b51859`).

**1. P1 — stdlib provenance leak.** You're right: `getAllImmutableSchemes()` includes `apexlib`, so the positive allowlist still admitted standard-library declarations. Added a dedicated `MUTABLE_DOCUMENT_SCHEMES` source-of-truth in `DocumentSelectorUtils` (`file` / `vscode-test-web` / `memfs` / `reefs` — the editable filesystems, synthetic `apexlib`/`apex-sobject` excluded by construction) and keyed the guard off that instead. Regression test: an `apexlib://` field now declines both `prepareRename` (null) and `rename` (no changes).

**2. P1 — usage-only partial edit.** Fixed. The fail-closed handling ran only inside `if (declaration)`, so a null declaration with usages fell through and returned a usage-only edit. Extracted the decision into a small pure function (`proceed` / `nothing-to-rename` / `decline-partial`): usages found without a locatable declaration now **decline** rather than emit a partial edit. Unit-tested all three branches.

**3. P2 — nested same-leaf disambiguation.** `fieldDeclarationRangeFromParse` now walks each candidate field's parent chain to its **owning-type FQN** and matches by suffix (namespace-tolerant), replacing the immediate-leaf comparison; genuine ambiguity fails closed to null. Added a unit test (same-leaf types in different branches resolve distinctly; a bare-leaf request declines).

One clarification on #3 while verifying: I couldn't reproduce the "comparison never matches → every rename declined" behavior. In the standalone parse the class-body block symbol's `name` is the type leaf, so the old comparison *did* match and produced an edit. What I did find (via a reversed-order test) is that when two nested types share **both** a leaf name and a field name in one file, the edit targets the *first* such type regardless of cursor position — and that mistarget originates **upstream**, in the shared cursor→containing-type resolver that hands `fieldDeclarationRangeFromParse` an already-wrong declaring type. The helper is now hardened + unit-tested, but fixing the end-to-end same-leaf case needs a change to that core resolver; I've scoped it as a separate follow-up rather than fold a riskier resolver change into this PR. Happy to file it under the epic if you'd like.

Validation: apex-ls suite green (407); apex-lsp-shared green (504); new `fieldDeclarationRangeFromParse` unit suite 9/9; rename worker-topology suite 27/27 (incl. the new stdlib-guard regression).

peternhale (changes requested):
Reviewed current head `bb29b71911665498f0943444b8472addfbaa85ae`.

Requesting a fix for one remaining issue:

**P2 — Configured editable document schemes are rejected by the new provenance guard.**

`USER_OWNED_URI_SCHEMES` is built only from the four default entries in `MUTABLE_DOCUMENT_SCHEMES`. However, `apex.environment.additionalDocumentSchemes` explicitly supports custom schemes and says they apply to all capabilities by default. An editable Apex document using a configured scheme such as `orgtest://` is accepted by the language client, but field `prepareRename` returns `null` at `worker.platform.shared.ts:4105-4107`, and direct field rename declines at `worker.platform.shared.ts:4207-4209`. Local rename does not have this restriction, so rename behavior becomes inconsistent by symbol kind.

Please make the provenance decision honor configured editable schemes (or explicitly model scheme writability and thread that source of truth to the worker), and add a regression test covering field prepareRename and rename on a configured custom editable scheme.

The three issues from the prior review are otherwise addressed in the production paths, and the current CI/web E2E checks are green.

peternhale (approved):
Looks good overall — the prepareRename field branch and the declaration-range fix (recomputing from a standalone parse instead of trusting the graph's `identifierRange`) are well-reasoned, with solid fail-closed handling for ambiguity, read-only sources, and workspace-load races.

A few minor gaps worth a follow-up (non-blocking):

1. **`fieldDeclarationRangeFromParse` single-match fast path** ([worker.platform.shared.ts:3459](https://github.com/forcedotcom/apex-language-support/blob/main/packages/apex-ls/src/worker.platform.shared.ts#L3459)) — returns the lone same-named field's range without verifying it belongs to `declaringTypeFqn`, unlike the 2+-match branch which disambiguates by owning type and fails closed. Narrow edge case (only bites if the true target field gets dropped from the `fields` filter and a same-named field elsewhere is the sole survivor), but it breaks the function's own documented fail-closed invariant.

2. **Redundant recompile of the declaring file** ([worker.platform.shared.ts:4818](https://github.com/forcedotcom/apex-language-support/blob/main/packages/apex-ls/src/worker.platform.shared.ts#L4818)) — for the common single-file rename case, the declaring file gets fully parsed a second time here even though Stage 6's candidate loop just above already compiled the same content. The Stage 6 table is discarded rather than reused.

3. **Duplicate reference lookup in `resolvePrepareRenameForField`** ([worker.platform.shared.ts:4116](https://github.com/forcedotcom/apex-language-support/blob/main/packages/apex-ls/src/worker.platform.shared.ts#L4116)) — repeats the `getReferencesAtPosition` + `exactCursorReference` lookup that `resolveCursorSymbol` (called just above) already performs, doubling an async lookup on every F2 field request.

None of these are blockers — approving as-is.

PR #687, opened 2026-09-04
feat(apex-lsp-compliant-services): renameMethod conflict detection + validation - W-23631133

kylewalke:
### What does this PR do?

Implements **5.3 renameMethod: conflict detection + validation wiring** (Phase 3 of the rename epic). A method rename that would collide with an existing member now returns a `ResponseError` instead of emitting a corrupting `WorkspaceEdit`.

**Key semantic:** methods overload, so a conflict requires a same-name **and same-signature** member — `foo(Integer)` → `bar` collides with an existing `bar(Integer)` but NOT with `bar(String)` (a legal overload). Field/property checks stay name-only.

Built in slices:
- **Signature-aware conflict matching** (`CheckMemberConflicts` data-owner query). The handler already had a `memberKind: 'method'` branch but matched by name only; it now also requires `doesSignatureMatch` for methods when a signature is supplied (the step-0 visibility self-verify matches the exact overload too). Fixed a latent bug: `queryDataOwner` hand-copies each request's fields and was silently dropping `CheckMemberConflicts.signature`, so it never reached the data owner.
- **Stage 4.5 wired into `resolveMethodRename`**, mirroring the field path: skip a no-op/case-only rename → query `CheckMemberConflicts` (`memberKind: 'method'` + signature + visibility) → conflict is a `ResponseError`; on query unavailability, `verifyMethodRenameLocally` (a parser-owned, fail-closed method analogue of `verifyFieldRenameLocally`) declines on any uncertainty. `resolveMethodContextForCursor` now also returns the method's `isPrivate` (Private OR Default), which gates the descendant conflict check.
- **Validation confirmation**: `validateRenameName(newName, SymbolKind.Method)` (Stage 4) already rejects an invalid identifier with a -32602 error, and the `isUserOwnedApexUri` provenance guard (`canBeRenamed` is-user-sourced) already rejects stdlib/generated methods (both from 5.2). Added an invalid-name regression test.

Out of scope (later WIs): method `prepareRename` (Phase 4 / 7.1) and 5.4 e2e. Qualify-on-conflict is deferred, matching the renameField precedent.

**Stacked on #673 (5.2 renameMethod WorkspaceEdit construction).** The base is the 5.2 branch, so this diff shows only the conflict/validation work; it should merge after #673 (which merges after #667). Kept as a **draft** until then.

Validation: full pre-commit matrix green (5549 passed, 0 failed); CheckMemberConflicts 24/24; renameMethod worker-topology 11/11.

@W-23631133@

### Functionality Before

A method rename applied its `WorkspaceEdit` without checking for member-name collisions — renaming `foo` → `bar` in a type that already declared a same-signature `bar` would mint a duplicate declaration.

### Functionality After

A method rename that collides with a same-signature member in the same type, a non-private ancestor, or (when the renamed method is non-private) a descendant is declined with a `ResponseError`; a legal overload at a different signature is not treated as a conflict, and an invalid new identifier is rejected up front.

### DIFF
--- a/packages/apex-ls/test/integration/CheckMemberConflicts.node.test.ts
+++ b/packages/apex-ls/test/integration/CheckMemberConflicts.node.test.ts
@@ -155,6 +155,17 @@ const BROKEN_SRC = `public class brokenType {
     }
 }`;
 
+// Overloaded methods for signature-aware conflict matching (WI 5.3). Renaming
+// `beta(Integer)` collides with `alpha(Integer)` (same signature) but NOT with
+// `gamma(String)` (a legal overload at a different signature).
+const MOVERLOAD_URI = 'file:///test/mOverload.cls';
+const MOVERLOAD_SRC = `public class mOverload {
+    public void alpha(Integer a) { }
+    public void alpha(String a) { }
+    public void beta(Integer a) { }
+    public void gamma(String a) { }
+}`;
+
 // One combined workspace containing every fixture. Ingested once.
 const WORKSPACE_ENTRIES = [
   { uri: BASE_URI, content: BASE_SRC, languageId: 'apex', version: 1 },
@@ -240,6 +251,12 @@ const WORKSPACE_ENTRIES = [
     version: 1,
   },
   { uri: BROKEN_URI, content: BROKEN_SRC, languageId: 'apex', version: 1 },
+  {
+    uri: MOVERLOAD_URI,
+    content: MOVERLOAD_SRC,
+    languageId: 'apex',
+    version: 1,
+  },
 ];
 
 type ConflictResult = {
@@ -254,6 +271,7 @@ type ConflictQuery = {
   memberKind: 'field' | 'method';
   isRenamedMemberPrivate: boolean;
   currentName?: string;
+  signature?: string[];
 };
 
 describe('CheckMemberConflicts data-owner query', () => {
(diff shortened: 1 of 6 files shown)


packages/apex-ls/test/integration/CheckMemberConflicts.node.test.ts
```diff
@@ -155,6 +155,31 @@ const BROKEN_SRC = `public class brokenType {
     }
 }`;
 
+// Overloaded methods for signature-aware conflict matching (WI 5.3). Renaming
+// `beta(Integer)` collides with `alpha(Integer)` (same signature) but NOT with
+// `gamma(String)` (a legal overload at a different signature).
+const MOVERLOAD_URI = 'file:///test/mOverload.cls';
+const MOVERLOAD_SRC = `public class mOverload {
+    public void alpha(Integer a) { }
+    public void alpha(String a) { }
```
daphne-sfdc: This method is unused in the test.  Line 527 only uses the Integer one.

packages/apex-ls/test/integration/CheckMemberConflicts.node.test.ts
```diff
@@ -477,6 +522,89 @@ describe('CheckMemberConflicts data-owner query', () => {
...
+    expect(result.conflict).toBe(true);
+    expect(result.reason).toBe('descendant');
+  });
+
+  it('falls back to name-only method matching when no signature is supplied', async () => {
+    // Without a signature (legacy/absent), a same-named method is a conflict
+    // regardless of overload — preserves backward-compatible behavior.
+    const result = await runConflictQuery({
+      definingTypeFqn: 'mOverload',
+      newName: 'alpha',
```
daphne-sfdc: Let's put `gamma` here instead of `alpha`.  That way we can prove that without the signatures, the two methods become inconclusive about overloading instead of the valid safe renaming in Line 543.

daphne-sfdc (approved):
## Conflict detection and overloading (PR #687)

Stage 4.5 of `resolveMethodRename` runs a hierarchy-aware, **signature-aware**
conflict check (`dataOwner:CheckMemberConflicts`) *before* building the edit.
A genuine collision → **ResponseError (-32600)**, a toast, **no partial edit**.
Three scopes (same-type, ancestor, descendant), each with its own visibility
rule, plus a static short-circuit. Dedicated fixtures isolate each rule:

| # | Start point | Action | Expect |
|---|---|---|---|
| K1 | `ConflictSameType.measure()` | rename `measure` → `zone` | **Decline** (same-type): `zone()` already exists with the SAME signature |
| K2 | `ConflictSameType.measure()` | rename `measure` → `label` | **Allowed** (different arity): `label(String)` takes 1 arg vs `measure()`'s 0 → legal overload |
| K3 | `ConflictSameType.gauge(Integer)` | rename `gauge` → `label` | **Allowed** (same arity, different type): `label(String)` is also 1 arg but a different param type → legal overload, not a conflict (the core of #687) |
| A1 | `ConflictAncestorChild.compute()` | rename `compute` → `existing` | **Decline** (ancestor): `ConflictAncestorBase.existing()` is PUBLIC → visible to the subclass |
| A2 | `ConflictAncestorChild.compute()` | rename `compute` → `hidden` | **Allowed**: `ConflictAncestorBase.hidden()` is PRIVATE → excluded from the ancestor walk |
| D1 | `ConflictDescendantBase.seed()` (public) | rename `seed` → `existing` | **Decline** (descendant): non-private → walk runs, collides with `ConflictDescendantChild.existing()` |
| D2 | `ConflictDescendantBase.secret()` (private) | rename `secret` → `existing` | **Allowed**: private member has no descendant visibility → walk skipped, subclass `existing()` ignored |
| ST1 | `ConflictStaticBase.tally()` (static) | rename `tally` → `count` | **Allowed**: statics skip the hierarchy walk; `ConflictStaticChild.count()` is legal hiding, not a conflict |
| ST2 | `ConflictStaticBase.tally()` (static) | rename `tally` → `sameType` | **Decline** (same-type): the same-type check still applies to statics; `sameType()` collides |
| ST3 | `ConflictStaticChild.count()` (static) | rename `count` → `tally` | **Allowed** (opposite of ST1): statics skip the hierarchy walk on the ancestor side too; ancestor `ConflictStaticBase.tally()` is legal hiding, not a conflict |

> **Visibility self-verify (step 0):** the descendant gate hinges on whether the
> renamed member is private. The handler re-reads the member's ACTUAL visibility
> from the complete graph and fails closed if it disagrees with the caller — so
> D2's "skip the walk" can't be spoofed into hiding a real D1-style collision.

[Conflict detection and overloading test files.zip](https://github.com/user-attachments/files/32526932/Conflict.detection.and.overloading.test.files.zip)

PR #691, opened 2026-09-09
feat(apex-lsp-compliant-services): generalize onPrepareRename across rename kinds - W-23631152

kylewalke:
### What does this PR do?

Implements **5.5: Generalize `onPrepareRename` across rename kinds.** Adds method `prepareRename` and unifies the prepareRename dispatch so F2 opens the rename box for methods (previously only locals and fields were wired).

- **`resolvePrepareRenameForMethod`** — mirrors `resolvePrepareRenameForField`: recompiles + loads referenced types, resolves the cursor symbol, gates on `kind === 'method'`, applies the `isUserOwnedApexUri` source-provenance guard (rejects synthetic read-only schemes like `apexlib://`), computes the cursor-containing range (usage token via `exactCursorReference`, else declaration), and enforces half-open containment (`endColumn > cursor`). Returns `{ range, placeholder }` or null.
- **`DispatchPrepareRename`** now dispatches `local ?? field ?? method`, mirroring the rename RESOLVE chain, with a documented extension seam for TYPE prepareRename (deferred to Group 6 / renameType).

Scope note: an un-renamable cursor surfaces as a plain null fall-through (parity with the local/field paths — no distinct error shape introduced). Full four-kind coverage awaits renameType (Group 6); this delivers the method kind and the generalization framework.

**Stacked on #687 (5.3).** Base is the 5.3 branch, so the diff shows only the prepareRename work; merges after the 5.x stack. Draft until then.

Validation: full pre-commit matrix green (5557 passed, 0 failed); prepareRename worker-topology tests 46/46 (method declaration cursor → range; usage cursor → range; identifier-end boundary → null; `apexlib://` method → null; non-renamable cursor → null).

@W-23631152@

### Functionality Before

F2 on a method returned "The element can't be renamed" — `prepareRename` was wired only for locals and fields, so the rename box never opened for methods.

### Functionality After

F2 on a method declaration or usage opens the rename box (returns the identifier range), gated to user-owned Apex sources and rejecting a cursor past the identifier end — consistent with the local and field paths.

### DIFF
--- a/packages/apex-ls/test/integration/RenameMethodThroughWorkerTopology.node.test.ts
+++ b/packages/apex-ls/test/integration/RenameMethodThroughWorkerTopology.node.test.ts
@@ -325,6 +325,63 @@ describe('renameMethod through the worker topology (W-23631132, slice 4)', () =>
       content: SOURCES[uri],
     }) as Promise<RenameResult>;
 
+  // Dispatch a single pool prepareRename with the cursor file's live buffer.
+  const prepareRename = (
+    uri: string,
+    position: { line: number; character: number },
+  ): Promise<{
+    range: {
+      start: { line: number; character: number };
+      end: { line: number; character: number };
+    };
+    placeholder: string;
+  } | null> =>
+    dispatcher.dispatch('prepareRename', {
+      textDocument: { uri },
+      position,
+      content: SOURCES[uri],
+    }) as Promise<{
+      range: {
+        start: { line: number; character: number };
+        end: { line: number; character: number };
+      };
+      placeholder: string;
+    } | null>;
+
+  // W-23631152: DispatchPrepareRename now dispatches local → field → method, so
+  // F2 opens the rename box on a method cursor in a fully ingested workspace
+  // (the same cone these rename tests exercise). prepareRename returns only the
+  // identifier range + placeholder — it does not walk the family — so a method
+  // DECLARATION cursor resolves to its own identifier range here.
+  it('returns prepareRename range for a method DECLARATION cursor (W-23631152)', async () => {
+    // Cursor on RmBase.run declaration (LSP line 1, char 25 — inside `run`),
+    // the same position the override rename test uses.
+    const result = await prepareRename(RM_BASE_URI, { line: 1, character: 25 });
+    logger.debug(`[prepare-rename:method-decl] ${JSON.stringify(result)}`);
+
+    expect(result).not.toBeNull();
+    expect(result!.range.start.line).toBe(1);
+    expect(result!.range.end.line).toBe(1);
+    expect(result!.range.start.character).toBeLessThanOrEqual(25);
+    expect(result!.range.end.character).toBeGreaterThan(25);
+    expect(result!.placeholder).toBe('run');
+  }, 120_000);
+
+  it('returns the usage range for a method CALL cursor (W-23631152)', async () => {
+    // Cursor on RmUtil.caller()'s in-body `compute()` CALL (LSP line 2, char
+    // 47). prepareRename must return the CALL's identifier range (via
+    // exactCursorReference), not the declaration on line 1.
+    const result = await prepareRename(RM_UTIL_URI, { line: 2, character: 47 });
+    logger.debug(`[prepare-rename:method-call] ${JSON.stringify(result)}`);
+
+    expect(result).not.toBeNull();
+    expect(result!.range.start.line).toBe(2);
+    expect(result!.range.end.line).toBe(2);
+    expect(result!.range.start.character).toBeLessThanOrEqual(47);
+    expect(result!.range.end.character).toBeGreaterThan(47);
+    expect(result!.placeholder).toBe('compute');
+  }, 120_000);
+
   it('renames a base method AND its cross-file override declaration', async () => {
     // Cursor on RmBase.run declaration (LSP line 1, char 25 — inside `run`).
     const result = await rename(
--- a/packages/apex-ls/test/integration/RenameThroughWorkerTopology.node.test.ts
+++ b/packages/apex-ls/test/integration/RenameThroughWorkerTopology.node.test.ts
@@ -482,12 +482,16 @@ describe('rename through the worker topology (Phase 0 no-op)', () => {
         }),
       );
 
-      // Cursor on the method name `compute` (line 1, char 20), which is NOT a
-      // local — prepareRename should return null (not renamable as local).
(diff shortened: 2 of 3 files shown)


packages/apex-ls/test/integration/RenameMethodThroughWorkerTopology.node.test.ts
```diff
@@ -325,6 +325,63 @@ describe('renameMethod through the worker topology (W-23631132, slice 4)', () =>
...
+    // Cursor on RmBase.run declaration (LSP line 1, char 25 — inside `run`),
+    // the same position the override rename test uses.
+    const result = await prepareRename(RM_BASE_URI, { line: 1, character: 25 });
+    logger.debug(`[prepare-rename:method-decl] ${JSON.stringify(result)}`);
+
+    expect(result).not.toBeNull();
+    expect(result!.range.start.line).toBe(1);
+    expect(result!.range.end.line).toBe(1);
+    expect(result!.range.start.character).toBeLessThanOrEqual(25);
+    expect(result!.range.end.character).toBeGreaterThan(25);
```
daphne-sfdc: We actually know the real numbers for the range and can be more accurate here.

``` suggestion
    expect(result!.range.start.character).toBe(24);
    expect(result!.range.end.character).toBe(27);
```

packages/apex-ls/test/integration/RenameMethodThroughWorkerTopology.node.test.ts
```diff
@@ -325,6 +325,63 @@ describe('renameMethod through the worker topology (W-23631132, slice 4)', () =>
...
+    // 47). prepareRename must return the CALL's identifier range (via
+    // exactCursorReference), not the declaration on line 1.
+    const result = await prepareRename(RM_UTIL_URI, { line: 2, character: 47 });
+    logger.debug(`[prepare-rename:method-call] ${JSON.stringify(result)}`);
+
+    expect(result).not.toBeNull();
+    expect(result!.range.start.line).toBe(2);
+    expect(result!.range.end.line).toBe(2);
+    expect(result!.range.start.character).toBeLessThanOrEqual(47);
+    expect(result!.range.end.character).toBeGreaterThan(47);
```
daphne-sfdc: Same here.

```suggestion
    expect(result!.range.start.character).toBe(44);
    expect(result!.range.end.character).toBe(51);
```

packages/apex-ls/test/integration/RenameThroughWorkerTopology.node.test.ts
```diff
@@ -904,6 +908,263 @@ describe('rename through the worker topology (Phase 0 no-op)', () => {
...
+        start: { line: number; character: number };
+        end: { line: number; character: number };
+      };
+      placeholder: string;
+    };
+    // On LSP line 1, and the range must CONTAIN the cursor (char 22).
+    expect(prepareInfo.range.start.line).toBe(1);
+    expect(prepareInfo.range.end.line).toBe(1);
+    expect(prepareInfo.range.start.character).toBeLessThanOrEqual(22);
+    expect(prepareInfo.range.end.character).toBeGreaterThan(22);
```
daphne-sfdc: Here too.

```suggestion
    expect(prepareInfo.range.start.character).toBe(19);
    expect(prepareInfo.range.end.character).toBe(26);
```

packages/apex-ls/test/integration/RenameThroughWorkerTopology.node.test.ts
```diff
@@ -904,6 +908,263 @@ describe('rename through the worker topology (Phase 0 no-op)', () => {
...
+        start: { line: number; character: number };
+        end: { line: number; character: number };
+      };
+      placeholder: string;
+    };
+    // The call is on LSP line 2, and the returned range must contain the cursor.
+    expect(prepareInfo.range.start.line).toBe(2);
+    expect(prepareInfo.range.end.line).toBe(2);
+    expect(prepareInfo.range.start.character).toBeLessThanOrEqual(17);
+    expect(prepareInfo.range.end.character).toBeGreaterThan(17);
```
daphne-sfdc: And here.

```suggestion
    expect(prepareInfo.range.start.character).toBe(15);
    expect(prepareInfo.range.end.character).toBe(21);
```

daphne-sfdc (approved):
## prepareRename — does the rename box open? (PR #691)

`prepareRename` runs *before* the rename box appears. On **F2** (or right-click →
Rename Symbol) VS Code first asks the server "can this be renamed, and what's the
range/placeholder?" A range → the box opens pre-filled; null → "The element can't
be renamed" and no box. #691 wires the **method** kind (before it, F2 on a method
returned "can't be renamed" — only locals/fields were wired). These are
prepareRename behaviors, distinct from the rename edit itself (#673/#687).

All rows use **`PrepareRenameTarget.cls`**, a dedicated fixture with a single,
uniquely-named method `greet()` (**no overloads**), keeping #691's "does the box
open?" checks focused on range + provenance only.

| # | Start point (where the caret sits) | Expect |
|---|---|---|
| P1 | On the `greet()` **declaration** name | **Box opens**, pre-filled with `greet` |
| P2 | On the `greet()` **usage / call site** (`return greet();` in `callGreet`) | **Box opens** — usage cursor resolves to the same identifier range, not just the declaration |
| P3a | Caret **at the first character** of `greet` | **Box opens** — the start edge is inclusive (`startColumn <= cursor`) |
| P3b | Caret **one before** `greet` (the space just before the first character) | **No box** — a cursor before `startColumn` is outside the identifier → "can't be renamed" |
| P3c | Caret **one past the end** of `greet` (immediately after the last character, before `(`) | **No box** — identifier ranges are half-open `[start, end)`, so a cursor at `endColumn` is outside → "can't be renamed" |
| P4 | On the **stdlib** call `valueOf` in `String.valueOf(42)` (`callStdlib`) | **No box** — source-provenance guard rejects non-user-owned (`apexlib://`) methods even when the call itself is in an editable file |
| P5 | On **whitespace / punctuation / a keyword** (not an identifier) | **No box** — plain null fall-through, same as locals/fields |

P3b/P3c (the identifier boundaries — outside at `start-1` and at `end`, inside
everywhere in `[start, end)`) and P4 (the provenance guard) are the cases least
likely to be hit by casual "does the box show up" testing and most likely to
surface a regression.

[Prepare rename test files.zip](https://github.com/user-attachments/files/32542969/Prepare.rename.test.files.zip)

PR #692, opened 2026-09-09
test(apex-lsp-extension): renameMethod e2e coverage + openFile harness fix - W-23631134

kylewalke:
### What does this PR do?

Implements **5.4: renameMethod e2e coverage.** Adds Playwright web e2e cases for renaming a method across an inheritance hierarchy, plus a harness reliability fix.

- **Three method-rename e2e cases** in `apex-rename-symbol.spec.ts` (+ 7 fixtures under `test-data/apex-samples/.../classes/`, mirroring the field-rename e2e):
  - **ancestor override** — rename a base method; the subclass `override` declaration + a cross-file call site are renamed.
  - **descendant override** — rename from a subclass override; the base declaration + a sibling override are renamed.
  - **interface implementer** — rename an interface method; the implementing class's method declaration + a call site are renamed.
  Each asserts the specific declaration shape and that the old name is gone from the class/interface body — asserting the right ranges/files were edited, not just "an edit happened".
- **Harness fix:** `ApexEditorPage.openFile` now confirms the requested file became the ACTIVE editor tab (with a quick-open recovery), via the new `waitForActiveTab`. Previously it waited only for "an editor" to be visible, so in a shared serial browser context a lingering rename widget / stale focus could leave a different fixture's content under assertion. This is a central improvement benefiting all specs.

**Stacked on #691 (7.1).** These cases require 7.1's method `prepareRename` (F2 won't open the rename box otherwise), so this PR is based on the 7.1 branch and merges after it.

Validation: the full `apex-rename-symbol.spec.ts` web run is **7/7** (local + field + all three method cases) in a single serial pass, zero flaky; full unit/topology matrix green (5557 passed, 0 failed).

@W-23631134@

### Functionality Before

No e2e coverage for method rename; the web spec's `openFile` did not verify the active editor, making serial cross-file rename assertions read the wrong file.

### Functionality After

F2 method rename is exercised end-to-end in the browser across ancestor/descendant/interface hierarchies, and the shared e2e harness reliably switches editors before asserting.

### DIFF
--- a/e2e-tests/pages/ApexEditorPage.ts
+++ b/e2e-tests/pages/ApexEditorPage.ts
@@ -51,14 +51,9 @@ export class ApexEditorPage extends BasePage {
    * @param filename - Name of the file to open (e.g., "ApexClassExample.cls")
    */
   async openFile(filename: string): Promise<void> {
-    const fileLocator = this.page
-      .locator(`[aria-label*="${filename}"]`)
-      .first();
-
-    const fileExists = await fileLocator.count();
-    if (fileExists > 0) {
-      await fileLocator.dblclick();
-    } else {
+    // Open via quick-open (Ctrl+P style). Reused for the initial open and the
+    // fallback below.
+    const openViaQuickInput = async (): Promise<void> => {
       await this.executeCommand('File: Open File');
       const quickInput = this.page.locator('.quick-input-widget');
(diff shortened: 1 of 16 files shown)


daphne-sfdc (approved):
> `apex-rename-symbol.spec.ts/junit.xml`: **100.0%** (7/7 passed)

Automated results for the new E2E tests pass, and the tests themselves look good.

